PR #34, opened 2016-07-21
Fix casing and spacing

duosearch:


cpacia:
Keep in mind any time you change a proto file you need to re-compile the pb files.
```
protoc --go_out=./ profile.proto
```

This creates a conflict with the other `fileDescriptor`s in the same package so I've been renaming them.

duosearch:
@cpacia Would it make sense to generate the pb files at compile time?

AubreyF:
@duosearch I prefer generating them by hand as necessary, as the resulting Structs can then be inspected by autocompleting IDEs, etc.

cpacia:
I fixed this and update the pb file

PR #36, opened 2016-07-22
Add closestpeers api call

cpacia:
for third party crawling


wds4:
On my node, closestpeers doesn't work but closestPeers does.

PR #60, opened 2016-08-04
Refactor repo path calculation and put OSX data in App Support.

### LINKED ISSUE #59: Instead of ~/.OpenBazaar2 please use ~/Library/Application Support for OS X
Instead of ~/.OpenBazaar2 please use ~/Library/Application Support as the destination for the user folder in OS X.

tyler-smith:
Resolves #59

### DIFF
--- a/openbazaard.go
+++ b/openbazaard.go
@@ -15,7 +15,6 @@ import (
 	"runtime"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


hoffmabc:
Ok this looks good. It created the folder properly on OS X.

PR #64, opened 2016-08-06
adds test for GetAPIUsernameAndPw function

davalb:
I want to contribute something to the new OpenBazaar server and thought unit tests might be good place to start. I chose the config.go file because I was looking for some simple functions to start with.
Please let me know what needs to be changed and what could be improved! Thanks.

### DIFF
--- a/repo/config_test.go
+++ b/repo/config_test.go
@@ -1,0 +1,31 @@
+package repo
+
+import (
+	"testing"
+)
+
+func TestGetAPIUsernameAndPw(t *testing.T) {
+	var path = "testdata/config"
+	username, pw, err := GetAPIUsernameAndPw(path)
+	if username != "TestUsername" {
+		t.Error("Expected TestUsername, got ", username)
+	}
+	if pw != "TestPassword" {
+		t.Error("Expected TestPassword, got ", pw)
+	}
+	if err != nil {
+		t.Error("GetAPIUsernameAndPw threw an unexpected error")
+	}
+
+	path = "testdata/nonexistent"
+	username, pw, err = GetAPIUsernameAndPw(path)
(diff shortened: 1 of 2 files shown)


davalb:
Wow, I wonder why the coverage has decreased so strongly? Is it because I added the repo/testdata/config file? The file is a fixture for testing-purposes. Is there a way to exlude it from the coverage-computation?

tyler-smith:
Hey @davalb, thanks for getting involved! I think you're correct about the test coverage thing. 

@hoffmabc I think you've been setting up coveralls, could you check that out please?

I think making the test config file path a constant would be good so we can reuse it when testing the other config funcs.

```go
import (
	"testing"
)

const testConfigPath = "testdata/config"

func TestGetAPIUsernameAndPw(t *testing.T) {
	username, pw, err := GetAPIUsernameAndPw(testConfigPath)
// etc ...
```

It's great you've started to test the config stuff because I want to change the unchecked type assertions into a Config struct that we unmarshal into. This will make that refactoring easier.

hoffmabc:
I believe it decreased because the coverage tool wasn't calculating the directory code in the total until you added a unit test file.

tyler-smith:
That makes sense. Good to keep in mind.

cpacia:
Thanks @davalb !

cpacia:
It looks like you need to run gofmt for travis to pass. If you have an ide, you can usually set it up to run automatically on save.

davalb:
@tyler-smith thanks! Glad it is useful, I want to write tests for the other funtctions in the config.go file as well. I added the constant to the test-file as you suggested.
@cpacia Just ran gofmt and made changes to the code accordingly.

PR #65, opened 2016-08-07
Fix panic, and add test

JustinDrake:
Missing fields in the config file would create a panic. Added a test for that and made a few cleanups.

I'll cleanup the rest of `repo/config.go`, refactor it and test it fully if this pull request goes through.


JustinDrake:
The command "diff -u <(echo -n) <(gofmt -d -s $(find . -type f -name '*.go' -not -path "./vendor/*"))" exits with 0 for me. Don't understand.

cpacia:
It looks like its just gofmt. Make sure you run that before committing. Most IDEs can be set to run it on save.

hoffmabc:
Are you running that command in the top level folder? Have you committed all gofmt changes? Travis certainly thinks there are more issues.

cpacia:
Fwiw I was just going to leave the panic when there is no password because we don't really want people running it on a remote server without a password. 

With this fixing the panic, you'll probably want it to print an error message and exit if they don't enter a pw.

JustinDrake:
@cpacia There's a panic if any field is missing, not just the password.

cpacia:
Yeah. Well if that's going to be changed them there probably needs to be validation of the config file and an error and exit if its not valid. All the values are needed so it can't continue missing data.

PR #72, opened 2016-08-09
adds tests for ExtendConfigFile and InitConfig

davalb:
when run the test will create the file repo.lock and a datastore-folder in the testdata directory. Should I write a cleanup-routine after the test to delete these files again? Or should I add these files to .gitignore?


cpacia:
lgtm thanks!

PR #82, opened 2016-08-16
test for init.go maybeCreateOBDirectories

davalb:
Hey guys, I am trying to write a test for maybeCreateOBDirectories and I encountered a little problem, maybe you can help. When I run this test it fails with this message
```
--- FAIL: TestMaybeCreateOBDirectories (0.00s)
	init_test.go:52: drwxr-xr-x
	init_test.go:53: -rwxrwxrwx
```
But I would expect the filepermissions to be equal. If you look at line 81 of init.go, the file-permissions of the root-dir are set to `os.ModePerm`. What am I missing here?

### DIFF
--- a/repo/init_test.go
+++ b/repo/init_test.go
@@ -2,6 +2,7 @@ package repo
 
 import (
 	"os"
+	"path"
 	"path/filepath"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


cpacia:
I don't see an issue either @tyler-smith do you know anything about file permissions in testing?

tyler-smith:
I think it's a umask issue. I'll check it out.

tyler-smith:
`fm := fi.Mode()` should be `fm := fi.Mode().Perm()`, but additionally umask is causing Go to not set the correct permissions on that directory.

@cpacia What do you think the best way to handle that is?

cpacia:
I'm not sure. Does the permission need to be tested or is that just part of the std library functionality?

tyler-smith:
It's probably not necessary.

hoffmabc:
I believe the ending permission will be:

modePerm & umask of the process

So it will end up with potentially a more restrictive permission than requested in code.

Have we tested that? @tyler-smith @davalb?

tyler-smith:
That is correct. The test as written assumes umask=0. We could force umask=0 in the code so we get the exact permissions requested if we want.

hoffmabc:
Well I think we need to figure out what the goal is. If the goal is that we need that directory writable for ALL then let's just ensure that. I think we should be respectful towards local system privilege restrictions, but we need to throw an error or something if they're lockign down the system to OpenBazaar's detriment.

davalb:
That makes sense. Could somebody write down what the minimum necessary file permissions are on those folders? I could modify the test so that it tests that the file permissions are not more restrictive than that.

tyler-smith:
@davalb I think the minimum would be the user has read and write permissions, so >= 600.

davalb:
thanks @tyler-smith , I changed the test so that it only checks to see if the owner has read and write permissions. Now the test passes.
So, I am submitting this for merge now. Please check, thanks!

tyler-smith:
@davalb lgtm, thanks!

PR #84, opened 2016-08-17
Add a custom JSON field inside metadata

dsmurrell:
After some conversation with @drwasho (http://imgur.com/a/s6Unw) and @SamPatt (http://imgur.com/a/TT1Pm) a while back, I'd like to submit a pull request to add a text field for JSON which allows nodes to add any amount of structured information to their listings. Extensive third party uses of this field might inform future updates to the protocol that weren't on the radar before or up the priority of those that were.

A good example use case is for a listing posted by a handy man in Cambridge that can only service people in this area. He may want to specify some JSON which allows others to know his location like:

```
{
    "location" : {
        "latitude" : "52.2053° N",
        "longitude" : "0.1218° E"
    }
}
```

### DIFF
--- a/pb/protos/contracts.proto
+++ b/pb/protos/contracts.proto
@@ -36,7 +36,7 @@ message Listing {
         ContractType contractType = 2;
         ListingType listingType   = 3;
         uint32 expiry             = 4; // unix timestamp
-        string customJSON         = 5; // JSON for extensions and plugins
+        string nonBindingMetadata = 5; // JSON for extensions and plugins
 
         enum ContractType {
             UNKNOWN       = 0;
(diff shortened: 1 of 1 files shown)


cpacia:
@dsmurrell this might work better as a protobuf `any` type.

cpacia:
Also the issue I have with this is that unless the extra data is displayed to the user in the UI, you end up with a situation where the buyer is putting his signature on a contract that he hasn't fully read and hasn't consented to. Kind of violates the reason for having a contract.

dsmurrell:
@cpacia if the intended format is JSON, would it not make sense to constrain the type to `string` instead of `any`?

I've changed the name to reflect that anything in this field is non binding.

@hoffmabc, @drwasho do you have any alternate suggestions to offer to Chris' concern? Another option would be to show this information to the buyer in the UI.

cpacia:
Closing because the contracts.proto has had many changes after this PR. We can open an issue and discuss the best way to handle this or if we want to do it.

PR #87, opened 2016-08-18
Simpler and more consistent labels in contracts.proto

JustinDrake:
Remove redundant information (e.g. `buyerOrder` -> `order`). Simplify label (e.g. `RicardianContract` -> `Contract`).

### DIFF
--- a/pb/protos/contracts.proto
+++ b/pb/protos/contracts.proto
@@ -2,11 +2,11 @@ syntax = "proto3";
 
 import "countrycodes.proto";
 
-message RicardianContract {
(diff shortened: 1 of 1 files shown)


cpacia:
The reason I did it like it is, is so it visually calls out which part of contract is produced by the buyer and seller.

JustinDrake:
Ok, understood. My understanding of the "Go programming philosophy" is to keep names clear and short. Since sellers or moderators cannot make an order (only buyers can!), I think `order` is better than `buyerOrder`.

cpacia:
We can maybe revisit this later as an issue.

PR #94, opened 2016-08-23
Add GET listings api call

cpacia:


tyler-smith:
@cpacia this looks good to me.

PR #95, opened 2016-08-24
First pass total calulation

cpacia:


api/jsonapi.go
```diff
@@ -533,20 +533,100 @@ func (i *jsonAPIHandler) PUTImage(w http.ResponseWriter, r *http.Request) {
 func (i *jsonAPIHandler) POSTListing(w http.ResponseWriter, r *http.Request) {
```
tyler-smith: What about something like this for error handling? Just to make things a bit cleaner IMO.

```go
func (i *jsonAPIHandler) POSTListing(w http.ResponseWriter, r *http.Request) {
  	w.Header().Add("Content-Type", "application/json")

    var err error
    defer func(){
      if err != nil {
        w.WriteHeader(http.StatusBadRequest)
        fmt.Fprintf(w, `{"success": false, "reason": "%s"}`, err)
      }
    }()

    //...

    err = i.node.SetListingInventory(ld.Listing, ld.Inventory)
    if err != nil {
    	return
    }

    f, err := os.Create(path.Join(listingPath, "listing.json"))
    if err != nil {
      return
    }

    //...
}
```

cpacia:
Yeah might work but we probably still need the WriteHeader line in each error clause because the response status code is different for type of error.

core/listings.go
```diff
@@ -475,33 +574,45 @@ func validate(listing *pb.Listing) (err error) {
 				return errors.New("Shipping option titles must be unique")
 			}
 		}
-		for _, shippingRule := range shippingOption.ShippingRules {
-			if int(shippingRule.RuleType) == 2 && listing.Item.Grams == 0 {
+		if shippingOption.ShippingRules != nil {
+			if len(shippingOption.ShippingRules.Rules) == 0 {
+				return errors.New("At least on rule must be specified if ShippingRules is selected")
+			}
+			if int(shippingOption.ShippingRules.RuleType) == 2 && listing.Item.Grams == 0 {
```
tyler-smith: @cpacia I think we should use enums here instead of integers.

core/order.go
```diff
@@ -252,6 +292,237 @@ func (n *OpenBazaarNode) Purchase(data *PurchaseData) error {
...
+				if err != nil {
+					return 0, err
+				}
+				shippingPrice := uint64(item.Quantity) * shippingSatoshi
+				itemShipping += shippingPrice
+
+				// Apply shipping rules
+				if option.ShippingRules != nil {
+					for _, rule := range option.ShippingRules.Rules {
+						if int(option.ShippingRules.RuleType) == 0 {
```
tyler-smith: @cpacia A switch is a bit more idiomatic but that's purely stylistic. Not a major issue.

core/order.go
```diff
@@ -252,6 +292,237 @@ func (n *OpenBazaarNode) Purchase(data *PurchaseData) error {
...
+					}
+				}
+				shippingTotal += itemShipping
+			}
+		}
+	}
+
+	// Process combined shipping rules
+	if len(combinedOptions) > 0 {
+		lowestPrice := uint64(2100000000000000)
```
tyler-smith: This seems like it should be a const even though it's fairly clear what it represents.

PR #100, opened 2016-08-28
Test create mnemonic

davalb:
This function proved to be difficult to test. The only way to test the failures of the NewEntropy and NewMnemonic function calls was to use dependency injection in createMnemonic. This provided a way to pass in mock versions of these functions that would fail.

### DIFF
--- a/repo/init.go
+++ b/repo/init.go
@@ -41,7 +41,7 @@ func DoInit(repoRoot string, nBitsForKeypair int, testnet bool, password string,
 	}
 
 	if mnemonic == "" {
-		mnemonic, err = createMnemonic()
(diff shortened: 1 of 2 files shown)


tyler-smith:
Looks good but I'm going to think abut it some. I think I should have made bip39.NewMnemonic take a bytesize and return a securely random mnemonic, and have a separate bip39.NewMnemonicFromEntropy. That would move these tests to that repo. I'm not sure; I'll think about it. It's certainly not a major concern either way, just a bit of refactoring.

@davalb @cpacia any thoughts there?

tyler-smith:
Ok. I think that's my preference but not worth the time right now.

@davalb Thanks for another great PR!

PR #104, opened 2016-08-31
Order processing

cpacia:


bitcoin/listeners/transaction_listener.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			if contains && c > 0 {
+				q := int(item.Quantity)
+				if c-q < 0 {
+					q = 0
+					orderId, err := calcOrderId(contract.BuyerOrder)
+					if err != nil {
+						continue
+					}
+					log.Warning("Order %s purchased more inventory for %s than we have on hand", orderId, path)
```
tyler-smith: @cpacia Are we not going to completely stop the order in this case?

bitcoin/wallet.go
```diff
@@ -24,9 +24,22 @@ type BitcoinWallet interface {
 	// Get the confirmed and unconfirmed balances
 	Balance() (confirmed, unconfirmed int64)
 
+	// Check if we have sufficient funds to make a tx
+	CheckSuffientFunds(amount int64, feeLevel spvwallet.FeeLevel) error
```
tyler-smith: Couldn't this be done outside the interface with something like `Balance() - amount - fee`? Keeping the interfaces as small as possible is good for allowing other implementations of the interface to be written easier.

core/order.go
```diff
@@ -567,59 +664,299 @@ func (n *OpenBazaarNode) getPriceInSatoshi(price *pb.Listing_Price) (uint64, err
...
+		return errors.New("Order is missing a timestamp")
+	}
+
+	// Validate that the hash of the items in the contract match claimed hash in the order
+	var itemHashes []string
+	for _, item := range contract.BuyerOrder.Items {
+		exists := false
+		for _, hash := range itemHashes {
+			if hash == item.ListingHash {
+				exists = true
```
tyler-smith: Seems like we could `continue` the outer `for` here. Saves iterations and then we can remove the `if !exists` check.

core/order.go
```diff
@@ -567,59 +664,299 @@ func (n *OpenBazaarNode) getPriceInSatoshi(price *pb.Listing_Price) (uint64, err
...
+	}
+	if total != contract.BuyerOrder.Payment.Amount {
+		return errors.New("Calculated a different payment amount than the buyer")
+	}
+
+	// Validate shipping
+	containsPhysicalGood := false
+	for _, listing := range listingMap {
+		if listing.Metadata.ContractType == pb.Listing_Metadata_PHYSICAL_GOOD {
+			containsPhysicalGood = true
```
tyler-smith: Break from the loop here.

core/order_confirmation.go
```diff
@@ -0,0 +1,163 @@
...
+		return err
+	}
+	var guidSig []byte
+	var bitcoinSig *btcec.Signature
+	var sig *pb.Signatures
+	sigExists := false
+	for _, s := range contract.Signatures {
+		if s.Section == pb.Signatures_ORDER_CONFIRMATION {
+			sig = s
+			sigExists = true
```
tyler-smith: break from loop here.

tyler-smith:
@cpacia I made some inline comments. I think this stuff is a place we should invest in a lot of tests. Other than that it all looks good to me.

bitcoin/listeners/transaction_listener.go
cpacia: We check the inventory level when the order is placed, but don't increment it until the payment comes through. It's possible we could over sell our inventory this way. The only alternative would be to increment when the order comes in but that opens your node up to an attack where the other side sends bogus orders to make your node tell everyone you're out of stock.

bitcoin/wallet.go
cpacia: It would be nice if we could do it like that, but we don't actually know the fee without building the transaction. The number of utxos used to get to the correct amount determines the fee that's paid. So we can't really calculate it out of the wallet. 

But I might be removing this if we end up changing the purchase flow tomorrow.

PR #132, opened 2016-09-11
Cleanup order.go (redundant MessageType check)

JustinDrake:


cpacia:
I'm not following how that is redundant. It's checking that the message is either an OrderConfirmation or Error

cpacia:
Oh nvm I see. The line above.

PR #135, opened 2016-09-13
Refactor to cookie authentication

cpacia:


repo/config.go
```diff
@@ -62,6 +47,21 @@ func GetAPIEnabled(cfgPath string) (bool, error) {
 	return enabled, nil
 }
 
+func GetAPIAuthentication(cfgPath string) (authenticated bool, username, password string, err error) {
+	file, err := ioutil.ReadFile(cfgPath)
+	if err != nil {
+		return false, "", "", err
+	}
+	var cfg interface{}
```
tyler-smith: @cpacia Why not have a concrete AuthConfig type?

cpacia: I suppose I could, but then if I want to pass that into the ipfs gateway I'd have to import from this package which I'd rather not do. 

So it's either create an authconfig type and only use it for our package or not use it at all and be consistent.

cpacia: I'll merge it for now and work on this later.

PR #141, opened 2016-09-27
Ts msg receiver

tyler-smith:
- Fixes config_test.go
 - Uses channel for MessageReceiver.messageQueue
- Use sync.WaitGroup zero value instead of pointer (No behavioral change)
- Lint imports

### DIFF
--- a/net/retriever/retriever.go
+++ b/net/retriever/retriever.go
@@ -31,12 +31,12 @@ type MessageRetriever struct {
 	service      net.NetworkService
 	prefixLen    int
 	sendAck      func(peerId string, pointerID peer.ID) error
-	messageQueue []pb.Envelope
+	messageQueue chan pb.Envelope
 	*sync.WaitGroup
 }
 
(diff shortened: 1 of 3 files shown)


net/retriever/retriever.go
```diff
@@ -92,10 +93,16 @@ func (m *MessageRetriever) fetchPointers() {
...
-	for _, env := range m.messageQueue {
-		m.handleMessage(env, nil)
+
+DRAIN_LOOP:
+	for {
+		select {
+		case env := <-m.messageQueue:
+			m.handleMessage(env, nil)
+		default:
+			break DRAIN_LOOP
```
cpacia: @tyler-smith When is default called. I see this breaks out of the loop but I'm not sure what triggers it.

tyler-smith: It executes if and only if none of the other cases can be fulfilled. In this case that means that the m.messageQueue channel is empty. So the entire loop construction is pull items off the queue and handling them until the queue is empty. Once it's empty it breaks from the loop and continues on.

cpacia: Ok I see. This will probably work except it only gets to this point after the wait group finishes. But with the current code the goroutines don't call wg.Done() until the function exits. But the function wont exit due to `m.messageQueue <- env`. So I think this would deadlock.

PR #168, opened 2016-10-21
Cleanup comments

JustinDrake:
I have tried to be relatively thorough cleaning up comments. Many files have been modified, so this would need to be merged relatively fast to avoid merge conflicts.


cpacia:
Thanks

PR #174, opened 2016-10-24
Minor cleanups to init.go

JustinDrake:


repo/init.go
```diff
@@ -17,18 +17,13 @@ import (
 
 var log = logging.MustGetLogger("repo")
 
-var ErrRepoExists = errors.New(`ipfs configuration file already exists!
-Reinitializing would overwrite your keys.
-(use -f to force overwrite)
-`)
-
```
cpacia: `ErrorRepoExists` is also used in openbazaard.go... which will cause the build to fail to compile if this is merged. You want to add it back in then I'll merge.

JustinDrake:
Readded `ErrRepoExists`

PR #176, opened 2016-10-24
Cleanup retriever.go

JustinDrake:


net/retriever/retriever.go
```diff
@@ -132,56 +135,58 @@ func (m *MessageRetriever) fetchHTTPS(pid peer.ID, url string, addr ma.Multiaddr
...
-		if err != nil {
-			return
-		}
-		valid, err := pubkey.Verify(ser, env.Signature)
-		if err != nil || !valid {
-			return
-		}
+	// Unmarshal plaintext
+	env := pb.Envelope{}
+	err := proto.Unmarshal(plaintext, &env)
```
cpacia: This line must be
```
err := proto.Unmarshal(plaintext, &env)
```
to avoid a compilation error.

PR #181, opened 2016-10-28
Rename Signatures to Signature

JustinDrake:
I haven't recompiled the protobufs. Apparently I'm meant to run `protoc go_out=./ contracts.proto` and expect errors, but I can't even install `protoc`. Is there documentation for MacOS X?


JustinDrake:
Screw that. I now realise there are two signatures.

PR #190, opened 2016-10-31
Fix bug in order.go

JustinDrake:
The contract type of value 1 is Listing_Metadata_DIGITAL_GOOD, but it seems we want Listing_Metadata_PHYSICAL_GOOD instead (with value 0).

### DIFF
--- a/core/order.go
+++ b/core/order.go
@@ -638,7 +638,7 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
 		if l == nil {
 			return 0, fmt.Errorf("Listing not found in contract for item %s", item.ListingHash)
 		}
-		if int(l.Metadata.ContractType) == pb.Listing_Metadata_PHYSICAL_GOOD {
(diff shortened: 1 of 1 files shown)


core/order.go
```diff
@@ -638,7 +638,7 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
 		if l == nil {
 			return 0, fmt.Errorf("Listing not found in contract for item %s", item.ListingHash)
 		}
-		if int(l.Metadata.ContractType) == 1 {
+		if int(l.Metadata.ContractType) == pb.Listing_Metadata_PHYSICAL_GOOD {
```
cpacia: Change to 

```
if l.Metadata.ContractType == pb.Listing_Metadata_PHYSICAL_GOOD
```

cpacia:
I recently switched to a fork of the jsonpb package to get rid of the weirdness where zero value enums are omitted from the json. 

That let's us get rid of having weird zero enums like NA, but there might be things I overlooked, like this, in the refactor. 

If you just fix that bug I'll merge it.

Issue #196, opened 2016-10-31
Abstract away the 5 verifySignaturesOn* functions


JustinDrake:
(Feel free to tag as TODO, and assign me.) The following 5 functions do a lot of the same stuff and there's a significant opportunity to DRY up the code. It took me quite a long time to parse and precisely understand what was going on in those functions, as the code in uncommented and a bit messy. It's important that these crypto functions be crystal clear for review.

- `verifySignaturesOnOrderCompletion`
- `verifySignaturesOnOrderConfirmation`
- `verifySignaturesOnOrderFulfilment`
- `verifySignaturesOnListing`
- `verifySignaturesOnOrder`

JustinDrake:
This refactor is being postponed for #199.

yurizhykin:
I experimented a bit with abstracting those into `verifySignature()` function, that accepts pubkey, message  and function to call from inside to obtain signatures (which is exactly in what they differ). Will submit a PR for review, if that's ok.

cpacia:
@yurizhykin that would be great. Just keep in mind I just merged a change to that code today so make sure you're working with the latest.

yurizhykin:
@cpacia , yep, I specifically waited until that PR was merged:)

PR #202, opened 2016-11-04
Refactor contract signatures

### LINKED ISSUE #199: Change bitcoin signature handling
Currently the Ricardian Contract structure calls for all objects to be signed with both the identity key and bitcoin key. 

The reason for the bitcoin signature is that there is a possible attack that goes like:
```
1. Vendor receives a moderated order.
2. Vendor generates a new OpenBazaar identity key.
3. Vendor creates a new Ricardian Contract identical to the order he just receives but swaps out the buyer's identity for the new fake identity and proceeds to sign the contract with the fake identity key and his own identity key.
4. The vendor files a dispute with the moderator saying he sent the goods and the buyer went AWOL before releasing the funds. The malicious Ricardian Contract is sent along to the moderator.
5. The moderator checks the signatures on the contract and notes everything validates. He attemps to contact the buyer to see what's up, but the message is sent to the fake identity and not to the buyer's real identity, so the moderator either never gets a response or the vendor responds on the fake identity and says FU!
6. Moderator releases the funds to the vendor.
```

A similar attack can be crafted by the buyer as well.

By signing the listing and order with the key each party uses as part of the multisig, this attack is thwarted.

The funds were sent to an address which uses `buyerKey`, `vendorKey`, `moderatorKey`. If the vendor were to try the above attack, he would need the `buyerKey` in order to create a fake Ricardian Contract. 

If he tries to fake it by changing the buyer's public key in the `Order` and signing with this new key, then the moderator will calculate that `buyerKey` + `vendorKey` + `moderatorKey` != the p2sh escrow address the vendor wants the funds released from.

What we need to consider is if we can simplify the contract signing. Right now the signature fields looks like:

```
message Signatures {
    Section section = 1;
    bytes guid      = 2;
    bytes bitcoin   = 3;
```

But as far as I know only the listing and order needs the bitcoin signature. Can anyone think of an attack where the other contract sections would also need a bitcoin signature?

If it's only those two sections that need the bitcoin signature.... and obviously only moderated orders need it, can we refactor the contract to put the bitcoin signature elsewhere so that we don't bloat the contract with unnecessary signatures? What's the cleanest way to do that?

This would obviously supersede #198 

@JustinDrake

cpacia:
Removes the bitcoin signature from the Signatures list and moves it into the
BuyerID and VendorID sections. The bitcoin signature is only needed once and
covering the GUID. This is to prevent an attack where the vendor tries to give
bad data to the moderator to get him to release the funds.  This refactor reduces
the number of signatures that need to be checked and reduces the size of the
contract.

Closes #199


JustinDrake:
I had a quick look and can't see anything obviously wrong. I will do a more thorough review when I start with #196.

Issue #191, opened 2016-10-31
Bad coupon handling


JustinDrake:
The following lines of code seem wrong:

					if vendorCoupon.GetPriceDiscount() > 0 {
						itemTotal -= itemTotal
					} else {
						itemTotal -= uint64((float32(itemTotal) * (vendorCoupon.GetPercentDiscount() / 100)))
					}

First of all, if `vendorCoupon.GetPriceDiscount()` is negative, then `itemTotal` grows. Secondly, `itemTotal -= itemTotal` is equivalent to `itemTotal = 0`, which also seems like the wrong thing to do.

PR #205, opened 2016-11-08
Fix coupon handling in CalculateOrderTotal()

yurizhykin:
Fix for #191 and my first PR

As I understand, original intention was to subtract `GetPriceDiscount()` or `GetPercentDiscount()` as percentage from `itemTotal`, if any of those returns non-zero.


cpacia:
LGTM Thanks!

PR #208, opened 2016-11-10
Move signature verification to dedicated function

yurizhykin:
Added new file `core/signatures.go` that contains two functions for verifying message and bitcoin signatures respectively. Verification is delegated to `verifyMessageSignature()` via closure argument that
captures logic for extracting signatures from contract context. Also, some imports rearrangement due to `goimports` tool.

Seems to be 150+ lines of code diff, so code review would be very much appreciated.

### DIFF
--- a/core/completion.go
+++ b/core/completion.go
@@ -5,6 +5,13 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"gx/ipfs/QmT6n4mspWYEya864BhCUJEgyxiRfmiSY9ruQwTUNpRKaM/protobuf/proto"
+	"gx/ipfs/QmYf7ng2hG5XBtJA3tN34DQ2GUN5HNksEw1rLDkmr6vGku/go-multihash"
+	"io/ioutil"
+	"os"
+	"path"
+	"time"
+
 	"github.com/OpenBazaar/jsonpb"
 	"github.com/OpenBazaar/openbazaar-go/ipfs"
 	"github.com/OpenBazaar/openbazaar-go/pb"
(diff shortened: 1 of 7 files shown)


yurizhykin:
Forgot to mention: this closes issue #196

cpacia:
Thanks I'll review tomorrow.

cpacia:
@yurizhykin I've got test failures when I run the qa test suite. Looks like signature verification is failing. I'd have to dig a little deeper to find out why

yurizhykin:
Are there any instructions on how to run QA tests? It's a shame, but I didn't run them, as there was no mention in README and CONTRIBUTE.

cpacia:
I want to hook the test suite to travis but the "offline" tests all fail on travis for some reason, but not on my local computer. I can't figure out why.

yurizhykin:
I'm currently having a hard time running tests myself, because it all fails with `Permission denied` error, and it is not obvious why exactly that happens.

cpacia:
hmm maybe you need to `chmod +x` the file.

yurizhykin:
No, it turned out, I already had `/tmp/openbazaar-go` dir owned by root. Now there's another problem: 
`unknown flag `disableexchangerates'`.

cpacia:
So make sure before you run the tests you compile the openbazaar-go binary against your latest code. 

I just run `go install` in the openbazaar-go directory and it puts the binary in `$GOPATH/bin/openbazaar-go`

So I run the test suite as 
```
./runtests.sh $GOPATH/bin/openbazaar-go /bin/bitcoind
```

yurizhykin:
I just realized that `openbazaar-go` binary that I had in `$GOPATH/bin` was compiled a month ago, when I started playing around with it, and I've never recompiled it since, so yeah, thatnks, that was it:)

JustinDrake:
@yurizhykin It looks like you could abstract away the sig extractor as well. Aren't they all basically doing the same thing?

yurizhykin:
@JustinDrake I've already started yesterday, will finish today's evening. The problem noted by @cpacia is very strange, as `Verify()` method in both situations accepts same parameters (pubkey, sig and message), but returns different values, so I tried changing `verifyMessageSignature()` to accept signature instead of extractor closure, though behaviour seems to remain the same. Will continue today.

yurizhykin:
Finally finished it. Moved out signature extraction as well.

cpacia:
Cool I'll take a look.

cpacia:
I think this LGTM, Thanks!

PR #216, opened 2016-11-11
Cleanups in Listings.go

JustinDrake:
Includes changing `Desc` to `Description` for consistency. This may require an equivalent change on the frontend.

### DIFF
--- a/core/listings.go
+++ b/core/listings.go
@@ -53,7 +53,7 @@ type listingData struct {
 	Title        string    `json:"title"`
 	Category     []string  `json:"category"`
 	ContractType string    `json:"contractType"`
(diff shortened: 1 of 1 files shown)


cpacia:
Just have to let @rmisio know that `desc` was changed to `description` in the listing index.

rmisio:
> Just have to let @rmisio know that desc was changed to description in the listing index.

Who do you think you are!?

<sup>(just kidding, sounds good :))</sub>

cpacia:
Also @JustinDrake do you want to update api.yaml to reflect the change from `desc` to `description`.

cpacia:
Actually looks like that api call isn't in the yaml and needs to be created.

Issue #217, opened 2016-11-11
Simplify IPFS interface


JustinDrake:
As I understand from talking to @cpacia, the IPFS interface code in the `ipfs` folder was written "in the dark" (no documentation) from the IPFS command line tool and ended up being a dirty hack. For example, inside `add.go` the function `AddFile` calls the functions `PreRun`, `Run` and `PostRun` which are not relevant to adding files. As explained in [this issue](https://github.com/ipfs/go-ipfs/issues/3060), there is a much neater way of doing things (see whyrusleeping's answer):

```
func AddFile(ipfs *core.IpfsNode, file string) (string, error) {
    fi, err := os.Open(file)
    if err != nil {
        return "", err
    }

    return coreunix.Add(ipfs, fi)
}
```

instead of

```
func AddFile(ctx commands.Context, fpath string) (string, error) {
	args := []string{"add", fpath}
	req, cmd, err := NewRequest(ctx, args)
	if err != nil {
		return "", err
	}
	res := commands.NewResponse(req)
	cmd.PreRun(req)
	cmd.Run(req, res)
	var fileHash string
	for r := range res.Output().(<-chan interface{}) {
		fileHash = r.(*coreunix.AddedObject).Hash
	}
	cmd.PostRun(req, res)
	if res.Error() != nil {
		return "", res.Error()
	}
	if fileHash == "" {
		return "", addErr
	}
	return fileHash, nil
}
```

cpacia:
If someone wants to do that it's fine, but I think this is the only command that could be simplified that way. And doing so would make this function inconsistent with the rest of the IPFS package.

JustinDrake:
Why is `Add` is the only command that could be simplified? It seems to me that `PreRun`, `Run` and `PostRun` are just helper functions for the IPFS command line, which we are not using. Looking into coreunix, I see [`add.go`](https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go).

cpacia:
Because as far as I know add is the only one where you have a function like coreunix.add. If you can find others in the IPFS code to do the same thing for the other functions then that works.

JustinDrake:
Well for one I think there's cat.go

https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/cat.go

Issue #214, opened 2016-11-11
DRY up net.go


JustinDrake:
In `core/net.go`, it's almost the same code written over and over ahead. Maybe @yurizhykin would be interested in cleaning that up?

The nine functions starting from `Follow` and ending with `SendOrderCompletion` can easily be DRYed up.

yurizhykin:
@JustinDrake sure, I will take a look at it after I'm done with #196

yurizhykin:
Submitted a PR for this, but please see note there in #221

PR #221, opened 2016-11-14
DRY up net.go

rodentrabies:
In response to #214. Abstacted out process of sending message to peer given by peer ID. Total of -113 lines of code, which is good, but I'm not sure it is really useful, as it makes code in those functions non-symmetrical and, as a result, harder to mentally parse as a whole (similar to @cpacia's note at #217). On the other hand, less code == less space to make something wrong, so either close or merge, depending on what is more suitable for the goals of the project.

### DIFF
--- a/core/net.go
+++ b/core/net.go
@@ -61,21 +61,16 @@ func (n *OpenBazaarNode) SendOfflineMessage(p peer.ID, m *pb.Message) error {
 	return nil
 }
 
-func (n *OpenBazaarNode) SendOfflineAck(peerId string, pointerID peer.ID) error {
+func (n *OpenBazaarNode) peerIDSendMessage(peerId string, message pb.Message) error {
 	p, err := peer.IDB58Decode(peerId)
 	if err != nil {
 		return err
 	}
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-
-	a := &any.Any{Value: []byte(pointerID.Pretty())}
-	m := pb.Message{
-		MessageType: pb.Message_OFFLINE_ACK,
-		Payload:     a}
-	err = n.Service.SendMessage(ctx, p, &m)
-	if err != nil { // Could not connect directly to peer. Likely offline.
-		if err := n.SendOfflineMessage(p, &m); err != nil {
+	err = n.Service.SendMessage(ctx, p, &message)
+	if err != nil {
+		if err := n.SendOfflineMessage(p, &message); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


core/net.go
```diff
@@ -61,27 +61,30 @@ func (n *OpenBazaarNode) SendOfflineMessage(p peer.ID, m *pb.Message) error {
...
 
+func (n *OpenBazaarNode) SendOfflineAck(peerId string, pointerID peer.ID) error {
+	message := pb.Message{
+		MessageType: pb.Message_OFFLINE_ACK,
+		Payload:     &any.Any{Value: []byte(pointerID.Pretty())},
+	}
+	return n.peerIDSendMessage(peerId, message)
+}
+
 func (n *OpenBazaarNode) GetPeerStatus(peerId string) string {
```
duomoney: Should this function be pushed up?

duomoney:
What do you mean by "it makes code in those functions non-symmetrical"? Are you referring to `SendOrder` being exceptional?

core/net.go
```diff
@@ -158,142 +136,51 @@ func (n *OpenBazaarNode) SendOrder(peerId string, contract *pb.RicardianContract
 }
 
 func (n *OpenBazaarNode) SendOrderConfirmation(peerId string, contract *pb.RicardianContract) error {
-	p, err := peer.IDB58Decode(peerId)
-	if err != nil {
-		return err
-	}
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
 	any, err := ptypes.MarshalAny(contract)
```
duomoney: For consistency, rename the `any` variables to `a`, or vice-versa.

yurizhykin: Done

yurizhykin:
@duomoney, yes, exactly that

JustinDrake:
@yurizhykin I find it strange that `SendOrder` uses `SendRequest` but the other functions use `SendMessage` or `SendOfflineMessage`. This is for two reasons:

1) It seems to suggest that `SendOrder` doe not work in the offline case. I though offline orders was part of OB 2.0.
2) It seems to suggest that the response for the actions other than `SendOrder` are not worth looking at (e.g. for errors). Error checking here seems to be missing.

Maybe @cpacia can clarify.

jackkleeman:
@JustinDrake In the offline case, SendOrder returns an error, at which point Purchase() resorts to using SendOfflineMessage. Why it is done this way (outside of net.go) unlike all other messaging I am not sure.

cpacia:
If the vendor is offline it changes the data we need to send so it doesn't work to just send the message as-is as an offline message. We need to modify it before sending the offline message.

cpacia:
I fixed the merge conflicts and and bugs created by other changes to net and merge.

PR #225, opened 2016-11-15
Fix bug in images.go

JustinDrake:
The wrong error is returned (`err` instead of `aerr`). Normalised to just `err`.


JustinDrake:
There are actually more bugs. I'll submit another PR

Issue #228, opened 2016-11-16
The Profile listing count is not updated when a listing is deleted.


rmisio:
The listingCount returned by the GET profile API is not updated when a listing is deleted.

yurizhykin:
Trying to fix this one, and cannot find anything on how to GET profile to use it in a qa test. There is no API call for this, or am I missing something?

PR #232, opened 2016-11-17
Update profile counts when needed

yurizhykin:
Abstracted profile counts updating into the method `updateProfileCounts()` in `core/profile.go`, that is called from `DeleteListing()` and `UpdateFollow()` methods (where counts recomputation is needed).

Fix for #228


cpacia:
LGTM

PR #233, opened 2016-11-19
Local API to fetch own listings

### LINKED ISSUE #204: Provide local API to fetch own listings.
We need a local api (e.g. `ob/listings`) for a user to fetch their own listings. At least one of the reasons for this is that if a user deletes a listing and then goes to / refreshes their store before a republish is complete, the deleted listing would still be retuned by an IPNS fetch and consequently displayed in the UI.

I think there were other reasons for having a local listings API (maybe some additional data needed to be included that's only relevant to our own listings), but I'm not recalling them at this moment.

yurizhykin:
Added /ob/listings API that returns empty list if no listing is available, and list of listings otherwise. 
Also qa test for case:
- GET listings on fresh node and check if empty list
- POST listing
- GET listings and check if one listing is returned

Fixes #204

### DIFF
--- a/api/endpoints.go
+++ b/api/endpoints.go
@@ -133,6 +133,10 @@ func get(i *jsonAPIHandler, path string, w http.ResponseWriter, r *http.Request)
 		i.GETInventory(w, r)
 		return
 	}
+	if strings.Contains(path, "/ob/listings") {
+		i.GETListings(w, r)
+		return
+	}
 	if strings.Contains(path, "/ob/listing") {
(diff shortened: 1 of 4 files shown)


yurizhykin:
@cpacia should the empty list be returned in case `index.json` does not exist or the error as it is?

cpacia:
I think he wants an empty list if the file does not exist

yurizhykin:
@cpacia, done as requested

PR #259, opened 2016-12-10
Change IPFS gateway PUT handling

cpacia:
Currently PUT is kind of hard to work with and requires many back and forth
messages to upload a directory. Additionally, we don't really want nodes who
are pushing to the gateway to have to upload their entire directory each time as
this wastes bandwidth.

This commit changes the put handler to accept a directory hash. The gateway then traverses the graph, downloading only those nodes it doesn't have. 

The downside to this approach is if client has firewall issues, the gateway might not be able to get files from the network. Possibly we might want to force an outgoing connection to the gateway node if one doesn't exist.

### DIFF
--- a/ipfs/blocks.go
+++ b/ipfs/blocks.go
@@ -1,0 +1,15 @@
+package ipfs
+
+import (
+	key "github.com/ipfs/go-ipfs/blocks/key"
+	"github.com/ipfs/go-ipfs/commands"
+)
+
+func BlockstoreHas(ctx commands.Context, hash string) (bool, error) {
+	node, err := ctx.GetNode()
+	if err != nil {
+		return false, err
+	}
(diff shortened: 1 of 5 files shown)


ipfs/blocks.go
```diff
@@ -0,0 +1,15 @@
+package ipfs
+
+import (
+	key "github.com/ipfs/go-ipfs/blocks/key"
+	"github.com/ipfs/go-ipfs/commands"
+)
+
+func BlockstoreHas(ctx commands.Context, hash string) (bool, error) {
+	node, err := ctx.GetNode()
+	if err != nil {
```
hoffmabc: Is it good practice to return false here if there is an error? Someone might not check the error message and assume it's false.

Issue #184, opened 2016-10-29
Refund after fulfillment 


JustinDrake:
`POSTRefund` has the following piece of code:

```
if state != pb.OrderState_FUNDED {
    ErrorResponse(w, http.StatusBadRequest, "order must be funded before refunding")
    return
}
```

What if the vendor has fulfilled the order (e.g. sent the goods) but the goods were damaged or the buyer was not satisfied, and the vendor wants to make a refund?

Should the vendor be able to make a refund if the order state is `FULFILLED`?

cpacia:
The code snippet above only checks that it was funded. Should still work if it's fulfilled.

JustinDrake:
As I understand, if the state is `FULFILLED` then `if state != pb.OrderState_FUNDED` fails, so the function `return`s, and `i.node.RefundOrder` doesn't get called.

	if state != pb.OrderState_FUNDED {
		ErrorResponse(w, http.StatusBadRequest, "order must be funded before refunding")
		return
	}
	err = i.node.RefundOrder(contract, records)

cpacia:
Looks like you're right.

PR #260, opened 2016-12-11
Fix issue #184

bearbin:
Allow vendors to refund buyers after the order has been
fulfilled. Also update the error message to be clearer.


cpacia:
Thanks

PR #265, opened 2016-12-14
Build the snap in travis

elopio:


elopio:
Thanks @hoffmabc ! 
Now a nice next step would be to hook travis to make a release to edge every time that the master branch changes. I can't do that because it requires the store credentials of somebody from the openbazaar project.

I can guide you when you have some time, just ping me in slack, but basically it's just to run `snapcraft enable-ci travis`.

Issue #243, opened 2016-11-28
Directly pass DHT record to node (without polling)


JustinDrake:
One of the limitations of using a DHT as an asynchronous messaging layer is that pointers expire (e.g. after 7 days). Another limitation is that [the DHT retriever](https://github.com/OpenBazaar/openbazaar-go/blob/master/net/retriever/retriever.go) is a polling mechanism (currently polling every hour), which isn't real-time.

Duo Search wants to provide a caching layer for these asynchronous offline messages to prevent pointers from expiring, and to accelerate delivery of DHT records in case of a flaky connection or a temporary disconnection. For such a caching layer to be built there needs to be a facility to pass offline records to a node. I'll bring up this point later today on the technical call.

cpacia:
Pointers expire, but they are republished by the node until it gets a response from the recipient (there is an eventual timeout of like 30 days but that is configurable). 

Also, the polling is really only needed when a node comes online. Once they are online the messages can be sent directly. The DHT really shouldn't be used if both nodes are online, except for maybe when one node has severe firewall issues.

JustinDrake:
> they are republished by the node

What if the node stays offline longer than the expiration period?

> the polling is really only needed when a node comes online

I disagree with that. A node can think it is online (e.g. the `openbazaar-go` daemon is running), but actually it is unreachable (e.g. some sort of network failure). Messages to such a node would be sent to the DHT, polled up to an hour later.

On another note (relevant to Duo Search) it would allow us to implement https://github.com/OpenBazaar/openbazaar-go/issues/98.

JustinDrake:
As part of the technical hangout on Monday November 28, there was agreement that adding a new [`MessageType`](https://github.com/OpenBazaar/openbazaar-go/blob/master/pb/protos/message.proto#L9) to hook into the retriever's [`attemptDecrypt`](https://github.com/OpenBazaar/openbazaar-go/blob/master/net/retriever/retriever.go#L137) function would be a minimal change that allows for nodes to directly communicate offline records. This gives service providers the flexibility to build relays around offline records that can be cached and communicated at will.

PR #267, opened 2016-12-14
Add 'OFFLINE_RELAY' message type

jackkleeman:
See #243 . An offline relay is a message containing within it an encrypted offline message, in the format that it would be pointed to in DHT. When receiving a relay message, the node treats it as if it would treat a newly retreived offline message. I have chosen not to go through the MessageRetriever functions, attemptDecrypt and handleMessage, firstly because these are not currently exported, and secondly because attemptDecrypt would attempt to send an OFFLINE_ACK, which is not necessary here. Instead, the Handler replicates their action.


jackkleeman:
Should I change to compiling protos with version 1?

cpacia:
There was some kind of issue with the compilation. This is using proto 3 but the compiler should have detected it. Unless you have a really old one. I fixed it though.

Issue #272, opened 2016-12-15
Messaging always does a IPFS lookup for the peer's public key, even when you already have it


jackkleeman:
In order-related messaging, the sender already has the public key of the recipient, as it is stored in the listing, but instead of using this information an IPFS lookup is performed (in core.EncryptMessage()). This forces all nodes to publish their public keys to IPFS, which is not useful for Duo or other buyer-only portals.

I propose a second function, core.EncryptMessageWithKey, which allows you to directly specify a public key instead of the peerID, and core.SendOfflineMessageWithKey which uses this. We could then adapt all contract-related functions in net to pull out the public key (of either buyer or seller) and use these new functions. 

This is more efficient, and also opens the door for buyers that are not nodes, and given the new offline message relay functionality, that have no interaction with IPFS at all.

cpacia:
That's probably fine to do. I would maybe just make the net functions take in an optional key and use it if it's passed in and if not, just look it up.

jjeffryes:
Does #274 close this issue?

jackkleeman:
Yup!

PR #274, opened 2016-12-16
Adapt core.EncryptMessage to optionally take a public key

jackkleeman:
See #272. core.EncryptMessage now takes a pointer to a public key as its second argument. If left nil, an IPFS lookup is done for the public key. Most functions in core/net.go now take advantage of this feature, mostly by looking in the contract; SendOrderConfirmation, SendCancel + SendReject (both via datastore lookup for public key), SendRefund, SendOrderFulfillment, SendOrderCompletion.

I have chosen to use libp2p.PubKey throughout, arguably instead we could use []byte, and finally unmarshal in core.EncryptMessage, and this would mean we don't need to import libp2p in net.go. I am not sure what the best option is.

### DIFF
--- a/core/net.go
+++ b/core/net.go
@@ -448,7 +448,7 @@ func (n *OpenBazaarNode) SendChat(peerId string, chatMessage *pb.Chat) error {
 	}
 	err = n.Service.SendMessage(ctx, p, &m)
 	if err != nil {
-		if err := n.SendOfflineMessage(p, &m); err != nil {
+		if err := n.SendOfflineMessage(p, nil, &m); err != nil {
 			return err
 		}
 	}
--- a/core/net.go
+++ b/core/net.go
@@ -424,31 +424,9 @@ func (n *OpenBazaarNode) SendDisputeClose(peerId string, resolutionMessage *pb.R
 	}
 	err = n.Service.SendMessage(ctx, p, &m)
 	if err != nil {
-		if err := n.SendOfflineMessage(p, nil, &m); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-func (n *OpenBazaarNode) SendChat(peerId string, chatMessage *pb.Chat) error {
-	p, err := peer.IDB58Decode(peerId)
-	if err != nil {
-		return err
-	}
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	a, err := ptypes.MarshalAny(chatMessage)
-	if err != nil {
-		return err
-	}
-	m := pb.Message{
-		MessageType: pb.Message_CHAT,
-		Payload:     a,
-	}
-	err = n.Service.SendMessage(ctx, p, &m)
-	if err != nil {
+==== BASE ====
 		if err := n.SendOfflineMessage(p, &m); err != nil {
+==== BASE ====
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


core/net.go
```diff
@@ -188,8 +190,10 @@ func (n *OpenBazaarNode) SendOrderConfirmation(peerId string, contract *pb.Ricar
 	}
 	err = n.Service.SendMessage(ctx, p, &m)
 	if err != nil {
-		if err := n.SendOfflineMessage(p, &m); err != nil {
-			return err
+		if k, err := libp2p.UnmarshalPublicKey(contract.GetBuyerOrder().GetBuyerID().GetPubkeys().Guid); err != nil {
```
JustinDrake: The error handling seems wrong here. If there is an error with `UnmarshalPublicKey`, return err. If not, move on to `SendOfflineMessage`. This should also avoid one level of code nesting. Similar issues below.

core/net.go
```diff
@@ -209,7 +213,20 @@ func (n *OpenBazaarNode) SendCancel(peerId, orderId string) error {
...
 	err = n.Service.SendMessage(ctx, p, &m)
 	if err != nil {
-		if err := n.SendOfflineMessage(p, &m); err != nil {
+		//try to get public key from order
+		order, _, _, _, _, err := n.Datastore.Purchases().GetByOrderId(orderId)
+		var kp *libp2p.PubKey
+		if err != nil { //probably implies we can't find the order in the Datastore
+			kp = nil //instead SendOfflineMessage can try to get the key from the peerId
+		} else {
+			k, err := libp2p.UnmarshalPublicKey(order.GetVendorListings()[0].GetVendorID().GetPubkeys().Guid)
```
JustinDrake: Should this use `BuyerID`?

core/net.go
```diff
@@ -233,7 +250,20 @@ func (n *OpenBazaarNode) SendReject(peerId string, rejectMessage *pb.OrderReject
...
 	err = n.Service.SendMessage(ctx, p, &m)
 	if err != nil {
-		if err := n.SendOfflineMessage(p, &m); err != nil {
+		var kp *libp2p.PubKey
+		//try to get public key from order
+		order, _, _, _, _, err := n.Datastore.Sales().GetByOrderId(rejectMessage.OrderID)
+		if err != nil { //probably implies we can't find the order in the Datastore
+			kp = nil //instead SendOfflineMessage can try to get the key from the peerId
+		} else {
+			k, err := libp2p.UnmarshalPublicKey(order.GetVendorListings()[0].GetVendorID().GetPubkeys().Guid)
```
JustinDrake: Should this use `BuyerID`?

core/net.go
jackkleeman: SendCancel is called only by order.go:
n.SendCancel(contract.VendorListings[0].VendorID.Guid, orderId)
So VendorID is correct I believe.

JustinDrake: Cool, my bad 👍

cpacia:
Probably will hold off on this until after the dispute resolution is merged because there might be merge conflicts. Should be in the next few days.

jackkleeman:
I have rebased. There were no conflicts but I needed to change the new dispute functions to use the changed SendOfflineMessage function. I am currently passing in nil so that a lookup is done but I would prefer that the functions that send disputes could take a public key as input, or otherwise an indicator of whether the function is being used to send to moderator, buyer or seller. This could then be used to get the public key from the contract. Any thoughts on making that change as well?

cpacia:
@jackkleeman you just want to rebase these two and i'll look to merge them in the next day or two.

jackkleeman:
Rebased!

Issue #282, opened 2016-12-19
Buyer should use refundaddress in contract on cancel, rejection, or refund (like seller)


jackkleeman:
On rejection or refund, the seller initiates a multisig transaction with the refund address stored in purchase. The buyer, however, always uses an address stored in their wallet, either the one marked as internal or external. Firstly, the **internal vs external address choice is inconsistent** - order.go provides the external address to the purchase (the seller will use this), while SweepMultisig() (used by buyer) uses internal for both wallet types, which may lead to buyer and seller trying to make different multisig transactions. 

Secondly, it is better for the buyer to use the refund address stored in the purchase, such that for rejection or refund this is guaranteed to be the same as the address the seller picks. I also suggest that this refund address is used on offline cancel.

Not only is this more consistent, it also means that a fork (eg Duo) can use OpenBazaar without holding any btc; it can use btc keys for signing, but send all returned money to an external address, with a simple change to the order function that allows a custom refund address.

Issue #283, opened 2016-12-19
Allow custom refund address to be provided on purchase


jackkleeman:
This requires #282 also.

PR #286, opened 2016-12-19
Allow optional user-supplied refund address on purchase, make internal vs external address use consistent, always use refund address from contract

jackkleeman:
SweepMultisig functions in bitcoind and spvwallet can now optionally take specific address, defaulting to their internal address as before. Purchase() can now also optionally take a refund address in its json input, defaulting to the internal address - previously the external address was used, which likely would lead to bugs. All buyer calls of SweepMultisig now use the address stored in the purchase contract, instead of letting the Sweep function supply the internal address. I have not added an equivalent ability for sellers to provide a custom address on confirmation of offline orders, but this isn't a bad idea.

This resolves issues #282 and #283


jackkleeman:
Rebased!

PR #295, opened 2016-12-24
Chat

cpacia:


JustinDrake:
Looks good. A few comments:

- Should there be a `ValidateChat` function that checks things like message length?
- I suggest adding a timestamp field to the `Chat` message protobuf (for when the sender sent the message). This will help ordering messages to the recipient, and the timestamp can be displayed to convey "freshness" of messages.
- It is common for messaging apps to display a subtle notification when the recipient has received the message and/or when the recipient has read the message. Is that something we want?
- As I understand, two peers can have several parallel "conversations". For example, there might be two separate moderated disputes on two different orders. Should there be a `conversation` field in the message protobuf, so that messages can to assigned to the right conversation?

cpacia:
> Should there be a ValidateChat function that checks things like message length?

Maybe. Do other app, like facebook enforce a max length?
@drwasho @jjeffryes @rmisio 

> I suggest adding a timestamp field to the Chat message protobuf (for when the sender sent the message). This will help ordering messages to the recipient, and the timestamp can be displayed to convey "freshness" of messages.

Yeah now that I think of it, it might be necessary for offline messages. My worry though is that messages will come in with a timestamp that is before the last one you sent and the UI will stick it before that message, making it seem buggy. Maybe we'll just have to see how it works in practice. 

> It is common for messaging apps to display a subtle notification when the recipient has received the message and/or when the recipient has read the message. Is that something we want?

I like that feature, but it requires more messages to be sent. Right now there is an ack on both the online and offline messages that could be used to at least ack that the person received it. 

> As I understand, two peers can have several parallel "conversations". For example, there might be two separate moderated disputes on two different orders. Should there be a conversation field in the message protobuf, so that messages can to assigned to the right conversation?

I'm mostly using the subject field for that purpose. A subject of "" is a general chat message that gets routed to the general chat. Whereas an order ID in the subject line will go to the specific order.

jjeffryes:
@cpacia I'm not sure what the max length is in other apps (a quick search says either 20k or 55k for FB), but I'm pretty sure they all have one. If you don't, someone could send a message with a gigabyte of text or something else absurd.

I think 20k characters would be a good limit.

net/service/handlers.go
```diff
@@ -840,3 +843,44 @@ func (service *OpenBazaarService) handleDisputeClose(p peer.ID, pmes *pb.Message
...
+
+	// Unmarshall
+	chat := new(pb.Chat)
+	err := ptypes.UnmarshalAny(pmes.Payload, chat)
+	if err != nil {
+		return nil, err
+	}
+
+	// Validate
+	if len(chat.Subject) > 500 || len(chat.Message) > 20000 {
```
tyler-smith: This same logic is used elsewhere. It should be abstracted so if it changes we don't have to hunt all over. The error message is also ambiguous as to what property has the issue.

net/service/handlers.go
```diff
@@ -840,3 +843,44 @@ func (service *OpenBazaarService) handleDisputeClose(p peer.ID, pmes *pb.Message
 
 	return nil, nil
 }
+
+func (service *OpenBazaarService) handleChat(p peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
```
tyler-smith: Why is options an interface{}?

PR #311, opened 2017-01-08
Images refactoring

### LINKED ISSUE #289: Support more image formats
OpenBazaar/openbazaar-go/core.SetHeaderImage for example only uses the image/jpeg codec which means image formats .gif and .png won't work. We should fix this.

rodentrabies:
Fixes #289 :
- Added support for GIF and PNG images (both formats supported in standard library, but require registering via `init()` functions in corresponding packages).
- Rewrote image handling fixing some errors such as missing file `Close()`s etc.
- Added test for supported image formats (also, first test in `core` package, should encourage to add more).


yurizhykin:
Hm, coverage really suffers:)

cpacia:
Thanks! I'll take a look.

PR #312, opened 2017-01-09
Stealth Transactions

cpacia:


jackkleeman:
Is this aimed at vendors withdrawing money? Very cool.

cpacia:
It allows someone to send bitcoins to another user's peerID. The coins arrive at a different bitcoin address each time preventing address reuse.

jackkleeman:
Will this be used as part of the protocol or is it more of an extra useful feature?

cpacia:
It's only going to be used in the wallet as far as I know.

JustinDrake:
Can stealth transactions be used for direct OpenBazaar payments (i.e. no moderator)?

cpacia:
Given that not all wallet implementations can handle stealth transactions (for example bitcoind can't) it's probably not a good idea.

repo/db/keys.go
```diff
@@ -92,6 +111,25 @@ func (k *KeysDB) GetPathForScript(scriptPubKey []byte) (spvwallet.KeyPath, error
 	return p, nil
 }
 
+func (k *KeysDB) GetKeyForScript(scriptPubKey []byte) (*btcec.PrivateKey, error) {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+
+	stmt, err := k.db.Prepare("select key from keys where scriptPubKey=? and purpose=-1")
```
tyler-smith: We should catch this to avoid a nil pointer panic on 121

ghost (changes requested):
Just one small fix, but I think it should be made.

PR #321, opened 2017-01-11
Use contract public key for entire contract state machine

jackkleeman:
This PR completes earlier work I did to change net functions to use public keys when they are known. Public keys for `SendOrderFulfillment`, `SendOrderCompletion`, `SendDisputeOpen`, `SendDisputeClose` can now be passed in; in all cases, the places where where they are called have access to contract holding the relevant public key (the only exception is `SendDisputeOpen` when sending to a moderator). 

The most obvious benefit of this is that a vendor can interact with a buyer who is not publishing their public key to ipfs; all interactions with the buyer now use the public key as provided in the initial BuyerOrder.

The only functions still forced to do an ipfs lookup are:
`SendChat`, `SendOfflineAck`, `Follow`, `Unfollow`, all of which are not connected to a contract holding a public key.
And `SendDisputeUpdate`, which only ever goes to the moderator; moderator public keys are not stored in contracts.

### DIFF
--- a/core/completion.go
+++ b/core/completion.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	libp2p "gx/ipfs/QmUWER4r4qMvaCnX5zREcfyiWN7cXN9g3a7fkRqNz8qWPP/go-libp2p-crypto"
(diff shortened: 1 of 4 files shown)


JustinDrake:
> moderator public keys are not stored in contracts

Is that something we should consider changing? cc @cpacia

jackkleeman:
@JustinDrake yes I think it would be ideal to have the contract store all information needed to complete the transaction without ipfs or any kind of internet connection needed except directly between the relevant parties.

PR #322, opened 2017-01-11
Ts docker

ghost:
- Extract build code out of deploy code.
- Allow building specific platforms.
- Add Dockerfile which uses linux/amd64 binary
- Add Makefile to make managing the above easier

@hoffmabc I think you should review this one because you setup the build stuff. `deploy.sh` should act exactly as before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ _testmain.go
 .idea/
 *.iml
 .gx/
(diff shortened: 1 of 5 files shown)


cpacia:
Not sure why the travis errors. It's saying it can't find a package?

Issue #314, opened 2017-01-10
Add PATCH ability to Profile API


rmisio:
Please add in the ability to save to the profile via HTTP PATCH so that only a partial data set could be sent in.

drwasho:
Thank you 👍

PR #329, opened 2017-01-15
PATCH method for Profile API

yurizhykin:
Solves #314 
`PatchProfile` function accepts patch data from request as a `map[string]interface{}` and reads old profile data, merging both maps using [mergo (https://github.com/imdario/mergo)](https://github.com/imdario/mergo) library. I hope that added dep is not an issue, but if it is, I will be glad to reimplement it on demand.

Also, I failed to add `mergo` package to `vendors/` because I'm facing an error that says `godep: Package (github.com/OpenBazaar/go-blockstackclient) not found`. Any suggestions?

### DIFF
--- a/api/endpoints.go
+++ b/api/endpoints.go
@@ -134,6 +134,8 @@ func patch(i *jsonAPIHandler, path string, w http.ResponseWriter, r *http.Reques
 	switch path {
 	case "/ob/settings", "/ob/settings/":
 		i.PATCHSettings(w, r)
+	case "/ob/profile", "/ob/profile/":
+		i.PATCHProfile(w, r)
 	default:
 		ErrorResponse(w, http.StatusNotFound, "Not Found")
 	}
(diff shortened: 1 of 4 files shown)


hoffmabc:
Is the godep issue a problem @cpacia

cpacia:
I don't know why it would give that error but I can add the dependency afterwards.

yurizhykin:
@cpacia , thanks for adding dependency. Just so I know how to do it myself in future, how is it done? `godep save` didn't do the trick and I'm pretty new to golang's infrastructure to understand why.

cpacia:
yeah I'm thinking it didn't work for you because you didn't have all the dependencies on your machines (and if you did they might be different versions). 

I think `godep update` might work to just add a single dependency, but I'd have to look it up to know for sure.

yurizhykin:
Yes, I didn't. I started adding them one by one with `go get ...` but then I thought there must be a better way. Couldn't loop over deps list though, because `go` tool printed them with `vendor/` directory prepended. Anyway, thanks a lot.

PR #332, opened 2017-01-19
SMTP (and other) notifications mechanism

yurizhykin:
Attempt to abstract out notification mechanism into separate part of the service. This PR is currently just for progress tracking purpose. Most commits will be squashed/changed later. Hope to receive comments when it's more or less ready. 

*Update*: Spent some time trying to change as little as possible, but it seems that notifications do need some refactoring (for example interface for integration of dangling `notifications.Notification` type). So started from that.

Task list (checked ones are ready for review):
- [x] Unify notifications in `core`, `listener` and `service` packages
- [x] Implement notifier mechanism
- [x] Implement SMTP notifier
- [x] Implement header/body generator for notifications
- [x] ~~Notification interface type~~ seems unnecessary
- [x] Testing

### DIFF
--- a/api/gateway.go
+++ b/api/gateway.go
@@ -33,7 +33,7 @@ func NewGateway(n *core.OpenBazaarNode, authCookie http.Cookie, l net.Listener, 
 	if err != nil {
 		return nil, err
 	}
-	n.Broadcast = wsAPI.h.Broadcast
+	n.Broadcast = manageNotifications(n, wsAPI.h.Broadcast)
 
 	topMux.Handle("/ob/", jsonAPI)
 	topMux.Handle("/wallet/", jsonAPI)
--- a/api/notifications/notifications.go
+++ b/api/notifications/notifications.go
@@ -124,6 +124,10 @@ type UnfollowNotification struct {
 	Unfollow string `json:"unfollow"`
 }
 
+type StatusNotification struct {
+	Status string `json:"status"`
+}
+
 type ChatMessage struct {
 	MessageId string    `json:"messageId"`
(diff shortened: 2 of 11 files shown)


yurizhykin:
Spent some time trying to change as little as possible, but it seems that notifications do need some refactoring (for example interface for integration of dangling `notifications.Notification` type). So started from that.

hoffmabc:
@yurizhykin can we resolve the conflicts so we can move forwarrd on this? ping @cpacia

yurizhykin:
@hoffmabc , sure. Just rebased on 'master'. Can be merged now, and I will add tests in another PR. But I think, review would be useful, since my commits change some core parts.

hoffmabc:
Yes of course. Thank you for the rebase.

yurizhykin:
Final touch: added template for notification emails. Also added `ServerPort` field to `SMTPSettings` struct, as most popular services seem to break "port 25" convention. Tested in [`MailHog`](https://github.com/mailhog/MailHog), everything seems alright. 

QA test is not done yet: spent whole evening trying to read from *python*'s `subprocess.Popen.stdout` on `smtpd.DebuggingServer` and failed. In REPL I observe that first `read()` blocks indefinitely, and when it is interrupted from keyboard, subsequent one actually returns expected data. Have no idea what it is. Have anyone seen something like that before?

Anyway, QA test can be rewritten to use `MailHog` and its JSON API but it seems like a lot of overhead to me.

cpacia:
Looks pretty good to me. Would just probably remove the `status` from the describe switch. 

Other than that I just need to test locally.

PR #336, opened 2017-01-24
Rebase to ipfs v0.4.5-pre2

cpacia:


JustinDrake:
Awesome. Is there a way to diff only the files outside of `vendor/`?

cpacia:
The diff for the OB files should be listed first.

PR #340, opened 2017-01-24
Switch identity keys to ed25519

cpacia:
This commit switches the default OpenBazaar identity key to ed25519 which is
now accepted as part of libp2p. It also changes the offline message encryption
and decryption algorithms to accept curve25519(converted from ed25519) via nacl
box in addition to RSA.


JustinDrake:
This is awesome, great job @cpacia.

PR #342, opened 2017-01-25
Tor

### LINKED ISSUE #99: Tor integration
As of August 26, the `openbazaar-go` codebase doesn't seem to have code for Tor integration. The intent of this issue is to push forward discussion about Tor integration, as this one of the most requested features for OpenBazaar. I'll start with a couple questions:
1. Is Tor integration planned for the OB2.0 initial release?
2. I tried to find a Tor library in Go but couldn't find anything satisfactory. Which Tor library is planned to be used for the integration?

cpacia:
- [x] Make it work over the onion transport
- [x] Create command line options for --tor and --dualstack
- [x] Make external API calls (blockstack resolver, exchange rates, message-retriever when fetching https) go out over Tor
- [x] Make bitcoin wallet implementations use Tor

closes #99


JustinDrake:
I think we discussed this already, but what's the advantage of dual stack over no Tor at all?

cpacia:
@JustinDrake you can't communicate with stores running exclusively on Tor if you are only using the clearnet.

jbenet:
This is awesome! 👍  

Reminder: let's audit {ipfs, OB, and this transport} before people rely on it to be fully secure.

hoffmabc:
Hey @El--Presidente interested in reviewing this with us?

bitcoin/bitcoind/wallet.go
```diff
@@ -77,13 +79,15 @@ func NewBitcoindWallet(mnemonic string, params *chaincfg.Params, repoPath string
 		masterPrivateKey: mPrivKey,
 		masterPublicKey:  mPubKey,
 		binary:           binary,
+		controlPort:      torControlPort,
+		userTor:          useTor,
```
gubatron: #style
`userTor`? typo here? (reads weird) `if w.userTor`.

bitcoin/bitcoind/wallet.go
```diff
@@ -92,13 +96,22 @@ func (w *BitcoindWallet) Start() {
 	if w.trustedPeer != "" {
 		args = append(args, "-connect="+w.trustedPeer)
 	}
+	if w.userTor {
+		var port int
+		if w.controlPort == 9151 {
```
gubatron: #styleOptimization
```
if w.controlPort == 9151 || w.control == 9051 {
    port--
}
```

gubatron: Q: what is the port number adjustment for?
Q2: is it ok to do this after what happens online 90
`args := []string{"-walletnotify='" + path.Join(w.repoPath, "notify.sh") + " %s'", "-server", "-torcontrol=127.0.0.1:" + strconv.Itoa(w.controlPort)}`

bitcoin/bitcoind/wallet.go
```diff
@@ -473,7 +487,7 @@ func (w *BitcoindWallet) ReSyncBlockchain(fromHeight int32) {
 	w.rpcClient.RawRequest("stop", []json.RawMessage{})
 	w.rpcClient.Shutdown()
 	time.Sleep(5 * time.Second)
-	args := []string{"-walletnotify='" + path.Join(w.repoPath, "notify.sh") + " %s'", "-server", "-rescan"}
+	args := []string{"-walletnotify='" + path.Join(w.repoPath, "notify.sh") + " %s'", "-server", "-rescan", "-torcontrol=127.0.0.1:" + strconv.Itoa(w.controlPort)}
```
gubatron: repetition?

bitcoin/bitcoind/wallet.go
```diff
@@ -482,6 +496,15 @@ func (w *BitcoindWallet) ReSyncBlockchain(fromHeight int32) {
...
 	}
+	if w.userTor {
+		var port int
+		if w.controlPort == 9151 {
+			port = 9150
+		} else if w.controlPort == 9051 {
+			port = 9050
+		}
+		args = append(args, "-listen", "-proxy:127.0.0.1:"+strconv.Itoa(port), "-onlynet=onion")
+	}
```
gubatron: repetition?

gubatron: #DRY

bitcoin/exchange/bitcoinprices.go
```diff
@@ -89,11 +96,12 @@ func (b *BitcoinPriceFetcher) fetchCurrentRates() error {
 }
 
 type BitcoinAverage struct {
-	cache map[string]float64
+	cache  map[string]float64
+	client *http.Client
 }
 
 func (b *BitcoinAverage) fetch() (err error) {
```
gubatron: perhaps this client should be passed to the fetch() method instead of adding it to that struct.

bitcoin/exchange/bitcoinprices.go
```diff
@@ -89,11 +96,12 @@ func (b *BitcoinPriceFetcher) fetchCurrentRates() error {
 }
 
 type BitcoinAverage struct {
-	cache map[string]float64
+	cache  map[string]float64
+	client *http.Client
```
gubatron: I think this is wrong, this might be better as part of `BitcoinPriceFetcher`

gubatron:
by looking at the code patterns, this interface
```go
type ExchangeRateProvider interface {
	fetch() error
}
```

could use being like this:
```go
type ExchangeRateProvider interface {
        client() *http.Client
        fetchUrl() string
        decode(body io.ReadCloser) error
	fetch() error
}
```

the code on `fetch()` is identical, except for its return type, but since all the `ExchangeRateProvider` implement `decode(body)`, `fetch()` could reuse a single generic implementation like this:

```
func (b *ExchangeRateProvider) fetch() (err error) {
	resp, err := b.client().Get(b.fetchUrl())
	if err != nil {
		return err
	}
	return b.decode(resp.Body)
}
```

since I believe they're initialized as `ExchangeRateProvider` 

```go
func NewBitcoinPriceFetcher(dialer proxy.Dialer) *BitcoinPriceFetcher {
...
b.providers = []ExchangeRateProvider{&BitcoinAverage{b.cache, client}, &BitPay{b.cache, client}, &BlockchainInfo{b.cache, client}, &BitcoinCharts{b.cache, client}}

b.run()
...
}
```

my guess is by doing that you don't have to implement `fetch()` so many times and avoid repetition. (warning: I'm not a go programmer, just curious)

bitcoin/exchange/bitcoinprices.go
cpacia: I can't look now, but I remember refactoring all this to make it easier to test.

bitcoin/bitcoind/wallet.go
cpacia: Good catch. That's a typo that it looks like the ide replaced everywhere. I'll fix.

bitcoin/bitcoind/wallet.go
cpacia: port is the socks port. I can make it more explicit.

bitcoin/exchange/bitcoinprices.go
cpacia: For the prices, I would like to standardize the API response there and have a list of price servers (maybe OB1, DUO) that all return the same thing rather than have to handle responses from multiple sources. 

I think that's a much better approach than this.

cpacia:
OK merging. 

Angel we will take a look at refactoring that price fetcher and possibly create a standard API schema so it's not so cumbersome to handle multiple APIs. 

Other than that we plan to pay someone to review this (and the IPFS code) and the broader code base for any privacy and security issues before we take the disclaimer off and let people go wild.

gubatron:
nuclear_meltdown

(reading over this code got me pumped about go again, if you won't work on that refactoring right away, as I bet you have a bunch more stuff to do, perhaps I'll be able to submit a PR before you get to it)

PR #352, opened 2017-01-30
Fix broken ADD_PROVIDER

JustinDrake:
As it stands `handleAddProvider` is (completely!) broken for pointers. The reason is that `p != pi.ID` for pointers. This means that `AddProvider` adds the wrong provider addresses for pointers. @jackkleeman and I spent many hours on this bug. I was surprised the OpenBazaar testbench has no test for pointers.

This PR also fixes another bug, by restricting the addition of provider records to the originator. The relevant lines of code were incorrectly removed from the [original code](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L279).

I have also added validation that len(pi.Addrs) == 1 for pointers.

### DIFF
--- a/vendor/github.com/ipfs/go-ipfs/routing/dht/handlers.go
+++ b/vendor/github.com/ipfs/go-ipfs/routing/dht/handlers.go
@@ -278,19 +278,31 @@ func (dht *IpfsDHT) handleAddProvider(ctx context.Context, p peer.ID, pmes *pb.M
 	// add provider should use the address given in the message
 	pinfos := pb.PBPeersToPeerInfos(pmes.GetProviderPeers())
 	for _, pi := range pinfos {
-		if len(pi.Addrs) < 1 {
-			log.Debugf("%s got no valid addresses for provider %s. Ignore.", dht.self, p)
+		if pi.ID != p && !isPointer(pi.ID) {
+			// we should ignore this provider reccord! not from originator.
+			// (we chould sign them and check signature later...)
+			log.Debugf("handleAddProvider received provider %s from %s. Ignore.", pi.ID, p)
+			continue
+		}
+		
+		if len(pi.Addrs) < 1 && !isPointer(pi.ID) {
+			log.Debugf("%s got no valid addresses for provider %s. Ignore.", dht.self, pi.ID)
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


jackkleeman:
As I discussed with @cpacia on friday, I think the code just currently adds the sender as a provider for the prefix hash, instead of adding the ciphertext hash as a provider address. This led to some very confusing results when we tried to get our relay to pull messages from the dht - we could find the prefix hash record but it didn't point to the ciphertext

cpacia:
I have to spend sometime debugging this. 

The final line of code:
> dht.providers.AddProvider(ctx, c, pi.ID)

I changed to `p` because the qa tests we're bombing trying to get an offline message. 

But just looking at the code it looks like it should be p if `!isPointer` otherwise p.ID. I'll test it out later. 

Did you run the qa tests with this patch?

jackkleeman:
I haven't tried QA testing, but I have managed to send messages to the relay using a bootstrap node with this patch applied (and not otherwise). I couldn't really piece together why qa doesn't fail before this patch.

JustinDrake:
> I changed to `p` because the qa tests we're bombing trying to get an offline message.

I believe `p` is wrong.

> But just looking at the code it looks like it should be p if !isPointer otherwise p.ID.

If !isPointer then `p == pi.ID`, so it doesn't matter. My code uses `p.ID`, which works in both cases.

> Did you run the qa tests with this patch?

No. The [README](https://github.com/OpenBazaar/openbazaar-go/blob/master/qa/README.md) states `test scripts in this package run automatically on all pull requests and commits.` so I guess it's all good. On this note, could the documentation to run the qa tests be beefed up? The only documentation is `./runtests.sh /path/to/openbazaar-go-binary /path/to/bitcoind-binary`, but I don't have bitcoind installed on my system.

cpacia:
OK this is my fault I commited this change to OpenBazaar/go-ipfs but apparently it failed to make it into the vendor package.

cpacia:
Fixed. Updated godeps which includes this patch.

JustinDrake:
As explained above, there's another bug:

> This PR also fixes another bug, by restricting the addition of provider records to the originator. The relevant lines of code were incorrectly removed from the original code.

Also,

> I have also added validation that len(pi.Addrs) == 1 for pointers.

PR #359, opened 2017-02-09
Add a max of 70 characters to the coupon code.

jjeffryes:
The coupon code should have a maximum, otherwise users can add very long strings (I tested with two paragraphs of lorem ipsum). 

This sets the max to 20.

If this PR is accepted, we will need to do the same validation on the client.

### DIFF
--- a/core/listings.go
+++ b/core/listings.go
@@ -890,8 +890,8 @@ func validateListing(listing *pb.Listing) (err error) {
 			return fmt.Errorf("Coupon title length must be less than the max of %d", SentenceMaxCharacters)
 		}
 		if len(coupon.GetDiscountCode()) > SentenceMaxCharacters {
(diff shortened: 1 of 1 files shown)


core/listings.go
```diff
@@ -889,6 +889,9 @@ func validateListing(listing *pb.Listing) (err error) {
 		if len(coupon.Title) > SentenceMaxCharacters {
 			return fmt.Errorf("Coupon title length must be less than the max of %d", SentenceMaxCharacters)
 		}
+		if len(coupon.GetDiscountCode()) > SentenceMaxCharacters {
+            return fmt.Errorf("Coupon code length must be less than the max of %d", SentenceMaxCharacters)
+        }
```
jjeffryes: I'm not sure why the spacing is off here, it looks correct in my IDE.

PR #383, opened 2017-02-21
Clean up listings and ratings indexes; fixes #381

### LINKED ISSUE #381: What happens if a listing has the title "Index"?
If a listing has the title `Index` the slug for it is then `index` and the file name is `index.json` which collides with the main listings `index.json`. Is there code to protect against such name collisions?

JustinDrake:
As per @hoffmabc's request.

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -1094,7 +1094,7 @@ func (i *jsonAPIHandler) GETListings(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 		}
-		listingsBytes, err := ipfs.ResolveThenCat(i.node.Context, ipnspath.FromString(path.Join(peerId, "listings", "index.json")))
+		listingsBytes, err := ipfs.ResolveThenCat(i.node.Context, ipnspath.FromString(path.Join(peerId, "listingsIndex.json")))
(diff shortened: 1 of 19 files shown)


cpacia:
@JustinDrake this is probably better to just put a check on the slug when saving that it cant be named index rather than refactoring the structure of the directory.

JustinDrake:
Why accumulate technical debt in the protocol at this early stage (pre-release)? Separation of concerns in the `listings` and `ratings` folder is clearly superior than the current approach.

cpacia:
I don't know what you mean. Those folders are already separated. This now sticks listing data in two separate folders... the root and the listing folder.

JustinDrake:
I mean that the `listings` folder does precisely one thing, namely store listings.

tyler-smith:
I support the proposal but have not reviewed the PR (it was assigned to Chris so I never checked it out).

@cpacia What do you think? I don't mind reviewing the PR if you'd rather not but do you have a yay or nay on the concept? Or more discussion?

cpacia:
I don't mind revisiting later, but it doesn't seem like an imperative change at this time. We would need a rebase to revisit as well.

PR #385, opened 2017-02-23
Listen on websocket transport

jackkleeman:
The ipfs websocket transport allows the browser to connect directly to ipfs nodes for online communication. The ipfs basichost has this transport enabled by default, but we need to add the relevant address to the swarm. This means that Duo browser clients, or anyone else for that matter, can engage in online comms. I am in the process of testing this and have already had a lot of success sending messages from js to an OB node with this websocket address enabled. This will greatly improve the usability of browser clients.

Fortunately, we can enable this transport in two lines!

### DIFF
--- a/openbazaard.go
+++ b/openbazaard.go
@@ -481,9 +481,6 @@ func (x *Start) Execute(args []string) error {
 		return err
 	}
 	onionAddrString := "/onion/" + onionAddr + ":4003"
-	// add websocket support for browser peers
-	cfg.Addresses.Swarm = append(cfg.Addresses.Swarm, "/ip4/0.0.0.0/tcp/9005/ws")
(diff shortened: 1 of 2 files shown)


openbazaard.go
```diff
@@ -481,6 +481,9 @@ func (x *Start) Execute(args []string) error {
 		return err
 	}
 	onionAddrString := "/onion/" + onionAddr + ":4003"
+	// add websocket support for browser peers
+	cfg.Addresses.Swarm = append(cfg.Addresses.Swarm, "/ip4/0.0.0.0/tcp/9005/ws")
+	cfg.Addresses.Swarm = append(cfg.Addresses.Swarm, "/ip6/::/tcp/9005/ws")
```
cpacia: @jackkleeman can you remove these two lines and add these addresses to the config in repo/config/InitConfig

jackkleeman:
changes made

PR #386, opened 2017-02-23
[DRY] Wallet cmd arguments preparation refactor

gubatron:
Simplified `Start()` and `ReSyncBlockchain(fromHeight int32)`, should be easier to maintain in the future if new parameters will need to be passed to the wallet by putting all the logic that was copied and pasted in both methods in a single place.

Cheers.

### DIFF
--- a/bitcoin/bitcoind/wallet.go
+++ b/bitcoin/bitcoind/wallet.go
@@ -85,9 +85,13 @@ func NewBitcoindWallet(mnemonic string, params *chaincfg.Params, repoPath string
 	return &w
 }
 
-func (w *BitcoindWallet) Start() {
-	w.shutdownIfActive()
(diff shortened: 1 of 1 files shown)


cpacia:
Lgtm. Also I think there's a way we can grab the socks port from the tor control. At some point I'd like to try to do that rather than assuming the port.

gubatron:
Thanks for the quick merge.

Re: port from Tor Control, 
aren't you already doing that with: `strconv.Itoa(w.controlPort)`, or is that `.controlPort` field being hardcoded elsewhere?

cpacia:
well the control port and socks port are different. Right now the code just subtracts 1 from the control port to the get socks port because that is the standard tor configuration but they could have modified their tor ports so it's better to get the actual socks port.

gubatron:
true true, I remember seeing that code when you first submitted the Tor patch. got ya.
Thanks.

Issue #344, opened 2017-01-26
Instructions to recompile the protobufs


JustinDrake:
What are the instructions for recompiling the protobufs? I am getting errors. The first problem I encountered was that it could not find the timestamp package.

Then I had a problem with the package name:

> net/networkservice.go:8:2: found packages pb (api.pb.go) and contracts (contracts.pb.go) in /Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/pb

I renamed the package to `pb` and now I get more errors

> Justins-MacBook-Pro:openbazaar-go justin$ go run openbazaard.go start
> import cycle not allowed
> package main
> 	imports github.com/OpenBazaar/openbazaar-go/api
> 	imports github.com/OpenBazaar/openbazaar-go/core
> 	imports github.com/OpenBazaar/openbazaar-go/net
> 	imports github.com/OpenBazaar/openbazaar-go/pb
> 	imports .
> 	imports .
> pb/api.pb.go:21:8: cannot find package "_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/proto" in any of:
> 	/usr/local/go/src/_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/proto (from $GOROOT)
> 	/Users/justin/go/src/_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/proto (from $GOPATH)
> pb/message.pb.go:21:8: cannot find package "_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/ptypes/any" in any of:
> 	/usr/local/go/src/_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/ptypes/any (from $GOROOT)
> 	/Users/justin/go/src/_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/ptypes/any (from $GOPATH)
> pb/api.pb.go:24:8: cannot find package "_/Users/justin/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/golang/protobuf/ptypes/timestamp" in any of:
[...]

PR #387, opened 2017-02-23
Protobuf compilation

yurizhykin:
Fix for #344 
Added package declaration to proto-files and a target to makefile.

Note: recompiling protobufs seems to produce a lot of diff, maybe `*.pb.go` files should be updated as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,10 @@ build:
 build_linux:
 	./build.sh linux/amd64
 
+protos:
+	cd pb/protos && protoc --go_out=.. *.proto
+
+
 ##
 ## docker
 ##
--- a/pb/protos/api.proto
+++ b/pb/protos/api.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
 
 import "contracts.proto";
 import "orders.proto";
--- a/pb/protos/contracts.proto
+++ b/pb/protos/contracts.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
 
 import "countrycodes.proto";
 import "google/protobuf/timestamp.proto";
--- a/pb/protos/countrycodes.proto
+++ b/pb/protos/countrycodes.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
 
 enum CountryCode {
     NA                             = 0;
--- a/pb/protos/message.proto
+++ b/pb/protos/message.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
 
 import "google/protobuf/any.proto";
 import "google/protobuf/timestamp.proto";
--- a/pb/protos/moderator.proto
+++ b/pb/protos/moderator.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
 
 import "contracts.proto";
 
--- a/pb/protos/orders.proto
+++ b/pb/protos/orders.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
 
 enum OrderState {
     // New order, has not yet been accepted by the vendor
--- a/pb/protos/profile.proto
+++ b/pb/protos/profile.proto
@@ -1,4 +1,6 @@
 syntax = "proto3";
+option go_package = "pb";
+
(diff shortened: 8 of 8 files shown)


yurizhykin:
Strange check fail:
`Error parsing coverage: open coverage.out: no such file or directory`

cpacia:
@yurizhykin unfortunately compiling the protobufs is a little more complicated due to how I have the packages structured. If we want to do it automatically we'd need to refactor the packages.

Primary issue is I think the protobuf compiler assumes each protobuf go file will be going in it's own package. I move them into the same package so it creates the following issues:

1) The package needs to be renamed to pb
2) The import need to be fixed after moving the file
3) Some method calls need the package removed. For example `*contracts.RicardianContract` to `*RicardianContract`
4) The compiler uses `fileDescriptor0` for all the file descriptors but only one `fileDescriptor0` can be in the package so the number needs to be incremented with a find/replace. 

I suppose we could put them all in the different packages to make it easier, but it's kind of annoying we'd have to do that.

cpacia:
Yeah @yurizhykin looks like you have the right idea, but when I tried that command there were a couple things messed up in the generated go file

1) The imports were messed up
2) The file descriptor was 0 but there are others in the package with the same file descriptor. 

It would be great if we could do this in one command. I've always been modifying the go file to fix these issues after it's generated.

yurizhykin:
@cpacia , OK, thanks for guidelines. I'll try to figure out a solution.

tyler-smith:
Anybody mind giving me a basic run-down of the issue with the file descriptors? It appears that each object has multiple fileDescriptors of the form `fileDescriptors{i}`. What needs to be changed with them?

cpacia:
@tyler-smith I think the protobuf compiler assumes the generated go files will all be in separate packages. 

Each one defines a separate variable called `fileDescriptor0`. Obviously there can only be one `fileDescriptor0` in each package. So you have to change the 0 to something else with a find/replace.

The protoc compiler also screws up the imports and doesn't import the correct `timestamp` and `any` packages.

yurizhykin:
I found out there is also another way to do this, with protobuf `option go_import` declaration. From looking at generated `*.pb.go` files, it seems to be ok, unless I misunderstand the problem. I reoredred and squashed commits.

cpacia:
@yurizhykin looks like this got it pretty close. Only remaining issue is the imports of the the `timestamp`, and `any` packages come out messed up. I'll merge it as they can be fixed manually, but maybe we could have this script do it.

yurizhykin:
@cpacia , is it this `google_protobuf` / `google_protobuf1` switch?

cpacia:
Well the path following `google_protobuf` is just google/protobuf which I don't have on my system and doesn't seem to be a valid path. I end up changing them to:
```
import google_protobuf "github.com/golang/protobuf/ptypes/any"
import google_protobuf1 "github.com/golang/protobuf/ptypes/timestamp"
```

yurizhykin:
It's strange, as this the diff I'm getting after saying `make protos`: 
2017-03-03-014541_494x100_scrot
It seems to be fine. 
Do you have this same lines removed in your version (due to recently updated protobuf)? It may be related:
2017-03-03-014814_507x125_scrot

cpacia:
I'm getting:
```
import google_protobuf1 "google/protobuf"
import google_protobuf "google/protobuf"
```
What version are you using? I'm on 3.0. Let me try upgrading.

cpacia:
Proto 0.3.2

yurizhykin:
Did you figure out the problem yet? I updated both `github.com/golang/protobuf/{proto,protoc-gen-go}` packages and my `.pb.go` output is the same as before.

cpacia:
I updated to protoc 0.3.2 and the imports are still the same when I run `make protos`

```
import google_protobuf1 "google/protobuf"
import google_protobuf "google/protobuf"
```

yurizhykin:
You mean *protoc* command line tool or go packages? I think the problem is with go packages. Did you try following:
```bash
go get -u github.com/golang/protobuf/proto
go get -u github.com/golang/protobuf/protoc-gen-go
```

cpacia:
I'm not having any luck. I updated both of those packages and get the same result.

yurizhykin:
Found something in golang/protobuf#43 : it seems like you have to have those packages installed locally for *protoc* to able to guess correct path.

Alternatively, it can be mapped with command line parameters.

PR #390, opened 2017-02-25
Simplify swarm addresses logic

JustinDrake:


cpacia:
The point of having those strings there was someone could use the --dualstack flag without having to edit their config file. 

There are two ways to use tor and dual stack mode... a command line option and via the config file. 

This commit removes the command line option for people who might have modified their config file in some way.

JustinDrake:
Cool, thanks for the explanation.

Issue #368, opened 2017-02-16
Variants Schema Redesign


cpacia:
We need to redesign the contract schema to match the design below.
![image](https://cloud.githubusercontent.com/assets/4504708/23044762/9425f04a-f46f-11e6-8344-235fe712db75.png)

Only exception to the design is I think that Wolf is going to add a Sku to the table as well.

So here's my initial proposal for the new schema:

```
repeated Option options    = 12;
repeated Sku skus          = 13;

message Option {
    string name                = 1;
    string description         = 2;
    repeated Variants variants = 3;

    message Variants {
        string name          = 1;
        Image image          = 2;
    }
}

message Sku {
	string combinations = 1;
	uint64 surcharge     = 2;
	string id                   = 3;
}
```

The `options` field is used primarily to enumerate what options are available so the UI can populate the drop down boxes for option selection during checkout. 

We would eliminate the current string `sku` field and replace it with the repeated `Sku` object. This object would contain a string similar to what is used in inventory to identify which combination it represents. For example `/large/blue`. 

The `surcharge` would be added to the base price found earlier in the listing. 

`id` is a sku number. 

### Issues
1. In the above schema the `image` is attached to each variant (ie: large or blue but not /large/blue). Do we want to keep it this way or should the image be per combination?

2. We need better names for `combinations` and `id`. I called it `id` above just because having `sku` inside `Sku` seems weird. So help would be nice naming them. 

3. What happens if they include a list of options but no list of skus?

4. Does this seem to work?

cpacia:
@rmisio @JustinDrake @jackkleeman @jjeffryes @drwasho

JustinDrake:
This is what I proposed during the last technical call. You have options (e.g. size and color). Each option has variants (e.g. small, medium, large). The SKUs enumerate the variant combinations with a well-defined defined ordering.

Each SKU has a price, quantity, image. This is maximally flexible. If the frontend wants to reuse images across SKUs, it can! If the frontend wants absolute prices, percentage surchages, absolute surchages, or something crazy, it can!

In the special case where there are no options, there's a single SKU corresponding to the empty combination. In general, if there are `n` options and `v_i` is the number of variants for option `i`, then there are `v_1 * ... v_n` SKUs.

```
repeated Option options    = 12; // e.g. [sizeOption, colorOption]
repeated Sku skus          = 13; // e.g. [smallBlueSKU, smallGreenSKU, mediumBlueSKU, ...]

message Option {
[...]

PR #392, opened 2017-02-28
Variant refactor

cpacia:


repo/db/inventory.go
```diff
@@ -42,7 +42,7 @@ func (i *InventoryDB) Get(slug string) (map[string]int, error) {
 	i.lock.RLock()
 	defer i.lock.RUnlock()
 	ret := make(map[string]int)
-	stm := `select * from inventory where slug like "` + slug + `%";`
+	stm := `select * from inventory where slug="` + slug + `";`
```
tyler-smith: As long as this function is called without user input it should be safe but I think we should always use parameterization for stuff like this so somebody down the line doesn't pass in user input without realizing how dangerous it is.

repo/db/inventory.go
```diff
@@ -70,23 +72,31 @@ func (i *InventoryDB) GetAll() (map[string]int, error) {
...
-	_, err := i.db.Exec("delete from inventory where slug=?", path)
+	_, err := i.db.Exec("delete from inventory where slug=? and variant=?", slug, variant)
 	return err
 }
 
 func (i *InventoryDB) DeleteAll(slug string) error {
 	i.lock.Lock()
 	defer i.lock.Unlock()
-	stm := `delete from inventory where slug like "` + slug + `%";`
+	stm := `delete from inventory where slug="` + slug + `";`
```
tyler-smith: This should also be parameterized

api/jsonapi.go
```diff
@@ -1016,7 +1028,12 @@ func (i *jsonAPIHandler) POSTInventory(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	for _, in := range invList {
-		err := i.node.Datastore.Inventory().Put(in.Slug, in.Quantity)
+		formatted, err := json.Marshal(in.Variants)
```
tyler-smith: Why store things as opaque blobs instead of normalized tables? It prevents atomicity, queryability, and data integrity.

JustinDrake:
What design did you end up picking @cpacia? Would it be worth having a spec to limit the reverse-engineering effort for code reviewers? Would you care updating #368?

Briefly looking at [the protobuf definition](https://github.com/OpenBazaar/openbazaar-go/blob/8182617ef68ab549fb89e64ca217239be3814f18/pb/protos/contracts.proto#L65), the `surcharge` feels unnatural:

1. Why use `surchage` instead of absolute prices? `surchage` shapes the thinking to a specific pricing scheme. Surchages can be the pricing metaphor openbazaar-desktop adopts without imposing it to the protocol.
2. Is it a percentage surcharge? Is it a absolute surcharge?
3. How are rounding errors handled? If percentage surcharges are used, rounding issues are a can of worms.

What do @tyler-smith and @jjeffryes think of absolute prices versus surcharges at the protocol level?

cpacia:
> Why use surchage instead of absolute prices? surchage shapes the thinking to a specific pricing scheme. Surchages can be the pricing metaphor openbazaar-desktop adopts without imposing it to the protocol.

It's added (or subtracted) from the base price of the listing. It would be weird if it was the total because then the base price of the item would not be used for anything. 

> Is it a percentage surcharge? Is it a absolute surcharge?
Absolute value

> How are rounding errors handled? If percentage surcharges are used, rounding issues are a can of worms.

Everything is converted into satoshis when calculating the total price. This is the same as it was before, just with a different schema.

JustinDrake:
> It would be weird if it was the total because then the base price of the item would not be used for anything.

The base price can be removed, completely. As you noticed, having `n+1` amounts to describe `n` prices is redundant. Using `n` prices to describes `n` prices seems the most natural.

cpacia:
The issue is we need a single price to display in the UI. It's probably not appropriate to randomly pick the /small/blue/cotton/ variant and use that as the display price.

JustinDrake:
> It's probably not appropriate to randomly pick 

Sure, a random pick is wrong. The default Sku could be implicit by using the first Sku in the `repeated` list. (The list is ordered, right?)

PR #393, opened 2017-02-28
Change Duo Bootstrap Peer ID

jackkleeman:
Looks like we've had the wrong PeerID for a while
@JustinDrake


JustinDrake:
Thanks @jackkleeman. I wrongly copy-pasted the root hash... Actually I think you were also tripped up by the output. Maybe the peer id and the root hash need to be clearly labelled and displayed next to each other.

PR #400, opened 2017-03-02
Reduce default prefix length to 14 bits

JustinDrake:
Reasons for reducing the prefix length to 14 bits:

* Increase deniability (Realistically the network will grow to less than 100,000 active nodes within the first year. Using 16 bits gives a very poor deniability ratio of 100,000/2^16 < 1!)
* Maintain room for growth (I am assuming nodes can bear an 8-to-1 deniability ratio. With 14 bits, that means the network can grow to 8*2^14 = 131,072 active users before the default prefix length will need changing. This is at least one year after launch.)
* It reduces the number of prefixes which makes DHT crawling for prefixes significantly easier for partners such as Duo.

### DIFF
--- a/core/net.go
+++ b/core/net.go
@@ -16,6 +16,7 @@ import (
 const (
 	CHAT_MESSAGE_MAX_CHARACTERS = 20000
 	CHAT_SUBJECT_MAX_CHARACTERS = 500
+	DefaultPointerPrefixLength  = 14
 )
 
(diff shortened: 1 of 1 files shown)


jackkleeman:
This isn't the only place we need to move to using the DefaultPointerPrefixLength variable. eg:
https://github.com/OpenBazaar/openbazaar-go/blob/master/openbazaard.go#L800
I don't think there are others, but I could be wrong

JustinDrake:
Oh shit. Could you fix that for me?

PR #403, opened 2017-03-04
Protoc package mapping

yurizhykin:
@cpacia , continuing #387 thread, here I added mapping between protobuf and golang packages, which is used in `protoc` command to generate correct imports. Any future protobuf dependencies should also be included in the makefile to ensure successful compilation in case *protoc* is unable to infer golang packages. 

I also added generated go files to make the diff visible, but if it is ok, those can be merged too, otherwise I'll remove the commit.


yurizhykin:
Note that a lot of selectors have been generated for different message types.

PR #406, opened 2017-03-05
Only send read message if read status changes

cpacia:
To prevent the server from sending read messages to the other chat party if the
UI calls MarkChatAsRead more than once, let's check to make sure the status
changed before sending the message.

### DIFF
--- a/repo/db/chat.go
+++ b/repo/db/chat.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"database/sql"
+	"fmt"
 	"github.com/OpenBazaar/openbazaar-go/repo"
 	"strconv"
 	"sync"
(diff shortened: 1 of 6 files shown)


repo/db/chat.go
```diff
@@ -150,44 +150,63 @@ func (c *ChatDB) GetMessages(peerID string, subject string, offsetId string, lim
...
 	}
 	tx, err := c.db.Begin()
 	if err != nil {
-		return "", err
+		return "", updated, err
 	}
 	var stmt *sql.Stmt
 	if messageId != "" {
 		stmt, _ = tx.Prepare("update chat set read=1 where peerID=? and subject=? and outgoing=? and timestamp<=(select timestamp from chat where messageID=?)")
+		stm := "select messageID from chat where peerID=? and subject=? and outgoing=? and read=0 and timestamp<=(select timestamp from chat where messageID=?)"
```
tyler-smith: I would personally so an EXISTS query and scan the result  into `&updated, which is clearer IMO but a very tiny detail and not a problem.

tyler-smith:
If you don't use EXISTS, at least add a `limit 1` and not worry about `break`ing.

PR #411, opened 2017-03-07
Moderator Notifications

cpacia:
Sends a message to a moderator whenever you add or remove them from your store. 

A notification is pushed to websockets. Should be trivial for the client to display as it's basically identical to the follow/unfollow notification just with different wording. 

A list of current stores you moderate is kept in the db. Since the UI doesn't have any designs for this, it will just sit here until we use it later in the future.


jjeffryes:
If the data for the stores you moderate is available, I think it would be good to display that somewhere.

Is there a way to interact with that data, and remove yourself from a store? I know we've talked about that before, and it would be difficult, plus we'd have to prevent that user from just adding you again.

cpacia:
Not at the moment

jjeffryes:
I'll put something in the backlog to keep track of it for future releases. I think we can skip it for 2.0.

core/moderation.go
```diff
@@ -218,3 +218,35 @@ func (n *OpenBazaarNode) SetModeratorsOnListings(moderators []string) error {
...
+	if settings.StoreModerators != nil {
+		for _, mod := range *settings.StoreModerators {
+			currentMods[mod] = true
+		}
+	}
+	var addedMods []string
+	for _, mod := range moderators {
+		if !currentMods[mod] {
+			addedMods = append(addedMods, mod)
+			delete(currentMods, mod)
```
tyler-smith: @cpacia Perhaps I'm thinking incorrectly, but shouldn't this `delete` call be outside the conditional? We want to end up with `currentMods` being a slice of mods not present in `moderators` correct?

cpacia: Yeah good catch

Issue #432, opened 2017-03-08
The dummy is not up-to-date


elopio:
I'm getting the following errors because in the dummy file there are statements not passing arguments recently added:
```

# github.com/OpenBazaar/openbazaar-go/dummy
github.com/OpenBazaar/openbazaar-go/dummy/main.go:262: not enough arguments in call to blockstackclient.NewBlockStackClient
github.com/OpenBazaar/openbazaar-go/dummy/main.go:263: not enough arguments in call to "github.com/OpenBazaar/openbazaar-go/bitcoin/exchange".NewBitcoinPriceFetcher
github.com/OpenBazaar/openbazaar-go/dummy/main.go:264: not enough arguments in call to selfhosted.NewSelfHostedStorage
github.com/OpenBazaar/openbazaar-go/dummy/main.go:267: not enough arguments in call to "github.com/OpenBazaar/openbazaar-go/net/repointer".NewPointerRepublisher
github.com/OpenBazaar/openbazaar-go/dummy/main.go:271: not enough arguments in call to "github.com/OpenBazaar/openbazaar-go/net/retriever".NewMessageRetriever
```

tyler-smith:
I'll have to rebase it. I've mainly only updated it when there were network-breaking changes. I need to get a CI flow going for our test swarm.

PR #433, opened 2017-03-08
snap: use go 1.7 from a remote part, remove the custom plugin

elopio:


snapcraft.yaml
```diff
@@ -20,5 +20,9 @@ parts:
     source: .
     plugin: go
     go-importpath: github.com/OpenBazaar/openbazaar-go
+    go-packages: [github.com/OpenBazaar/openbazaar-go]
```
elopio: This is needed because of #432, but it doesn't hurt. Otherwise it will try to install all the packages, when only this one is required.

elopio:
After our discussion about ipfs requiring go1.7, we started here discussing about how to support build dependencies that are not in the xenial archive. The custom plugin+PPA was obviously an ugly and unsafe solution, so this is a better approach. It uses the remote part from https://github.com/elopio/go/blob/master/snapcraft.yaml to get more recent versions of go, even 1.8 or master.

The downside is that it compiles go, so it's slightly slower. Now we are working on using snaps as build dependencies, so they will be precompiled already.

Any feedback will be appreciated.

snapcraft.yaml
tyler-smith: I think that sounds ideal anyways.

tyler-smith:
Thanks for the contribution

PR #436, opened 2017-03-08
Sanitize API

cpacia:


api/sanitize.go
```diff
@@ -0,0 +1,59 @@
...
+	var i interface{}
+	err := d.Decode(&i)
+	if err != nil {
+		return nil, err
+	}
+	sanitize(i)
+	return json.MarshalIndent(i, "", "    ")
+}
+
+func sanitize(data interface{}) {
```
tyler-smith: Since this accepts all types but doesn't work with all types I'd think returning an error for unhanded types would be nice.  If we can't get compile-time integrity we could at least run-time.

api/sanitize.go
```diff
@@ -0,0 +1,65 @@
...
+					return sanitize(t)
+				}
+			case []interface{}:
+				for _, t := range d {
+					return sanitize(t)
+				}
+			}
+		}
+	default:
+		return fmt.Errorf("Unsupported type: %t", d)
```
tyler-smith: I think this a good way to handle it

PR #448, opened 2017-03-10
Add allowedIPs in the API and gateway

### LINKED ISSUE #287: Remove misleading documentation about an `-a, --allowip=` command line option
There is documentation accessible via the command line (-h) and in the README.md of this repo that indicates an `allowip` command line option is available to configure which addresses a server will allow API connections from.

Apparently, this is not a valid option (perhaps the documentation is a hold-over from v1) and the correct way to do this is to set the Addresses.Gateway setting in the config file.

Expected results:
- mentions of the `allowip` flag are removed from the documentation
- it would be nice if there was some documentation that the `Addresses.Gateway` setting should be used to accomplish this (perhaps in general some documentation is needed on how to configure a node to work with a remote client)

cpacia:
closes #287

### DIFF
--- a/repo/testdata/config
+++ b/repo/testdata/config
@@ -49,11 +49,15 @@
     "PrivKey": "testKey"
   },
   "Ipns": {
+    "QuerySize": 0,
     "RecordLifetime": "7d",
     "RepublishPeriod": "24h",
     "ResolveCacheSize": 128
   },
   "JSON-API": {
+    "AllowedIPs": [
+      "127.0.0.1"
+    ],
     "Authenticated": true,
     "CORS": "*",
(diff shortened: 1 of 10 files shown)


drwasho:
LGTM

repo/testdata/config
```diff
@@ -49,11 +49,15 @@
     "PrivKey": "testKey"
   },
   "Ipns": {
+    "QuerySize": 0,
```
tyler-smith: @cpacia what is the reason for this change?

cpacia: That test data needs to be a properly formatted object for the test to pass. So it needs to have the QuerySize in there.

PR #455, opened 2017-03-13
Bumpfee

cpacia:


api/jsonapi.go
```diff
@@ -2488,3 +2492,36 @@ func (i *jsonAPIHandler) DELETEBlockNode(w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		ErrorResponse(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+	newTxid, err := i.node.Wallet.BumpFee(*txHash)
+	if err != nil {
+		ErrorResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	m, _ := i.node.Datastore.TxMetadata().Get(txid)
```
tyler-smith: I think you intended to assign to err here since you check it later on.

tyler-smith: Nvm, I misread. Handling it would be good though.

PR #460, opened 2017-03-15
Reuse streams for all messaging

jackkleeman:
As discussed with @cpacia, add a new mechanism for stream reuse; when a peer sends an online message to you, this stream is never closed (barring errors), and all future communication can go back and forth over this stream.

This means that the following groups can now receive online messages where previously they could not:
1. Users behind heavy nats
2. Users connecting from a browser implementation of libp2p 😉 

All they need to do is initiate a connection first, eg with a chat message. Given that it is mostly buyers with networking challenges, and mostly buyers initiating communications, this is a great solution.

The code is modelled (read: shamelessly taken) from the implementation of stream reuse used by the libp2p kademlia-dht implementation; all peers are assigned 'messageSender' objects which act as caretakers for the stream. The only difference is that in this implementation incoming streams will also be reused for outgoing messages to the same peer, while in dht the same outgoing stream is used repeatedly. You can see the libp2p implementation [here](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/dht_net.go)

We could implement a timeout for streams, to save memory; however I am not sure I think this is worthwhile on desktop implementations, especially considering how worthwhile it is to be able to have duplex online communication many minutes after the inital connection, even behind a NAT. Furthermore, the DHT implementation connects to many more peers than OpenBazaar needs to, and has no timeout.

### DIFF
--- a/net/service/messagesender.go
+++ b/net/service/messagesender.go
@@ -4,13 +4,11 @@ import (
 	"context"
 	"fmt"
 	inet "gx/ipfs/QmQx1dHDDYENugYgqA22BaBrRfuv1coSsuPiM7rYh1wwGH/go-libp2p-net"
+	ggio "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/io"
 	protocol "gx/ipfs/QmZNkThpqfVXs9GNbexPrfBbXSLNYeKrE7jwFM2oqHbyqN/go-libp2p-protocol"
 	peer "gx/ipfs/QmfMmLGoKzCHDN7cGgk64PJr4iipzidDRME8HABSJqvmhC/go-libp2p-peer"
 	"sync"
 	"time"
-
-	ggio "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/io"
-	proto "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/proto"
 
 	"github.com/OpenBazaar/openbazaar-go/pb"
 )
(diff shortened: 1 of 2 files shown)


hoffmabc:
Looks like the Travis unit tests are failing on the api module @jackkleeman

cpacia:
Thanks Jack. Looks mostly good. We don't need to do it now but I think it would make sense to at least remove the peer from the map if they go offline.

Other than that you just need to change the ggio url to: `ggio "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/io"`

and I believe the proto url to `"github.com/OpenBazaar/openbazaar-go/pb"`

jackkleeman:
Fixed. I used "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/proto" for proto, may as well use the interface; the whole messagesender.go file is entirely generic (except for hardcoded OpenBazaarProtocol).

cpacia:
I'm only a little worried about using gogo protobuf for the proto.message argument because of compatibility with our pb package. I've had errors I couldn't figure out what the problem was only to find out I imported gogo proto instead of the other one.

jackkleeman:
Good point. In general, I think you tend to get problems because proto.Message and pb.Message are not equivalent, you need to use *pb.Message. I have replaced with pb.

jackkleeman:
I need to add some more considerations to this, so don't merge yet. One thing I need to do is register a handler on all created streams to expect  (unrelated) messages to be returned over them. For requests, though, this will be a problem as the request and the stream handler would be listening to the stream for messages: this problem also arises if we receive a message from a peer and then reply with a request over the stream.

I can solve this in two ways: either we have requests always go over a newly created (and then closed) stream instead of reusing old or incoming streams, or we modify the request framework such that the responses to request messages get parsed by the stream handler and then forwarded to the requesting function. I am unsure If this would require changes to the request protobufs to add a way to link outgoing and ongoing messages with an id.

If we always make new streams for requests, then it is likely that we can't send request messages to users behind nats even if they initiate a connection: but the key request message is used for ordering which is from buyer to vendor so has less of a nat problem. I think that having everything work through the stream handler is a cleaner solution, and means that we essentially have a duplex protocol that only needs one stream for any communication.

@cpacia @JustinDrake which of these two options seems more appropriate?

jackkleeman:
Third solution: use a sync primitive to stop the stream handler listening while we send request messages. This means that we couldn't listen for other messages from the peer until the request response comes in, but i guess that if the peer is able to send new messages then they are able to respond to the request immediately, removing the lock. This might be the cleanest option.

PR #461, opened 2017-03-15
Add NewAddress wallet function

cpacia:
Orders need to go to a unique address in order for the listener to determine
what order they are for. This would normally be OK except scenarios where someone
has placed a direct order but hasn't paid (or two direct orders come in at the same time).

So we need a wallet function that returns a new address each time. Sadly this
can make it such that malicous buyers can place unfunded orders and run up
the size of the lookahead window in the keychain making it unlikely if not impossible
for all transactions will be detected when restoring from seed.

This can be fixed by either adding the lookahead window size to the config (TODO)
and implementing better DoS protection for unfunded orders.

### DIFF
--- a/bitcoin/bitcoind/wallet.go
+++ b/bitcoin/bitcoind/wallet.go
@@ -164,6 +164,11 @@ func (w *BitcoindWallet) CurrentAddress(purpose spvwallet.KeyPurpose) btc.Addres
 	return addr
 }
 
+func (w *BitcoindWallet) NewAddress(purpose spvwallet.KeyPurpose) btc.Address {
+	addr, _ := w.rpcClient.GetNewAddress(Account)
+	return addr
+}
+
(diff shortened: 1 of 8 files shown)


bitcoin/bitcoind/wallet.go
```diff
@@ -164,6 +164,11 @@ func (w *BitcoindWallet) CurrentAddress(purpose spvwallet.KeyPurpose) btc.Addres
 	return addr
 }
 
+func (w *BitcoindWallet) NewAddress(purpose spvwallet.KeyPurpose) btc.Address {
+	addr, _ := w.rpcClient.GetNewAddress(Account)
```
tyler-smith: Why not propagate this error? Other than that lgtm.

cpacia: I think I need to fix the errors in the interface in general. Will do that as a separate issue.

PR #464, opened 2017-03-16
Handle streams created for outgoing messages

jackkleeman:
Follows discussion at the end of #460 
Register handlers on created outbound streams, prevents message sending while we handle incoming messages, prevents message handling while we send outgoing messages.
Message senders cancel the context on the streamhandlers reader, and then lock the stream for sending. The streamhandler locks sending while it handles incoming messages.

Would appreciate some review @cpacia and @JustinDrake as this really made my head hurt working out whether this means we can't miss messages or handle them in the wrong place.

This is the last step needed for duplex communication over a single stream, which will improve nat availability hugely.


cpacia:
Ok I think it will mostly work but as it seems to me there could still be a race issue. If, in the time between when Alice sends a request to Bob and Bob receives the request off the wire, Bob fires off an outgoing message, it looks like Bob's outgoing message will acquire lock first (before the incoming read acquires it). 

This would cause Bob's message to be treated as the response to Alice's request, which is incorrect. 

We're only talking milliseconds, so it should be unlikely but still possible. 
@tyler-smith do you have any thoughts on this?

I'm not actually sure how to make things match up perfectly with tracking messages IDs in a map.

jackkleeman:
That's a good point. I have no big ideas about how to prevent that. I guess we need to wrap request messages with some kind of ephemeral id for matching request and response and have everything be read through the handler. This does somewhat move away from the normal libp2p pattern but I suppose we are already doing that.

cpacia:
That's probably the best way. I think we can add a messageID to the message object. The response to a request can contain the same messageID. 

There can be a map which stores a messageID/channel mapping. When a message comes in it can look up the channel in the map then pass the message into it. 

The send request function could block until the data arrives on the channel (or a timeout is triggered) and then return once it has the data.

jackkleeman:
Ok i was thinking the same thing. I'll set that up tomorrow and update this PR. Until then, the master branch will probably have problems with repeated online communication, as it's not handling the streams it makes for online messages.

jackkleeman:
this should be good now @cpacia take a look

cpacia:
OK @jackkleeman I think we're getting there. I think I would just make a couple changes...
- It might be better to have the request map and the lock be part of the messageSender object rather than part of service. Reason being so that messages from other peers don't have to wait to acquire lock while it's processing a message from someone else. 
- I'm not sure we need to boolean in the message proto. Since you're already doing:
```
ch, ok := service.requests[pmes.RequestId]
if ok {
```
I think that's sufficient to determine if it's a response correct?

(Also I'm not sure if you're aware but we can run `make protos` now to compile the protobufs and it works).

cpacia:
@tyler-smith you have any thoughts?

jackkleeman:
If we don't have a boolean, we can't distinguish a response from a new request (both will have an id that isn't in the map). I think we want to prevent responses for which the request has timed out from being interpreted as new requests. Everything else i agree on.

How did you know i didn't use make protos? I just ran the command directly as I'm not on a Unix system so make would be a pain to set up.

cpacia:
When we send a request we put the outgoing id in the map right? messagesender.go line 126. And then we send it back with the same ID. Or am I missing something.

jackkleeman:
Yeah but i was removing from the map when we time out waiting for a response. Then if the response comes after that, it's indistinguishable from a new request unless we have that Boolean.

cpacia:
hmm. So in that event it just calls continue it looks like.

jackkleeman:
Yes, we don't want to process responses that we aren't interested in any more. And we can't distinguish those from new requests in a general way without the boolean

cpacia:
I can see that. On the other hand I do think we should increase that timeout to like 10 seconds or long. In practice the only thing the timeout should be doing is triggering a delete from the map so the length doesn't matter much. I could even be 30 seconds. 

I would also put the timeout handling inside the sendRequest function so that all service has to do is:
```
c, ok := ms.requests[pmes.RequestID]
if ok {
    c <- pmes
}
```
In that scenario it seems like the only way a node would get a message after the timeout (and thus need the response bool) is if there's some serious misbehavior. But it's possible a peer could misbehave at any time (where the bool is there or not) so we will definitely need to make sure we handle misbehavior in any case. 

So for me I think I would still remove the bool from the message and use a longer timeout period (and probably change requestID to messageID).

jackkleeman:
The timeout is one minute, and it is in the sendrequest code in messagesender.go, in a select statement in the ctxreadmsg function. The select statement with a one second timeout in the handler is just in case the request function has not started trying to pull from the response channel by the time the response comes in, which is extraordinarily unlikely but i figured it's worth ensuring there are no odd edge cases, and also i didn't want to block the stream handler forever if something goes wrong on the sendrequest side which means the channel is still open but there's nothing receiving on it. Does that make sense?

Given that it is one minute, yes i suppose we could assume that responses won't be missed, and they'll just error if one is missed and gets processed probably.

jackkleeman:
I don't want to use messageId as this is already a field in the chat protobuf

PR #468, opened 2017-03-20
embrace timestamp conversion functions

jackkleeman:
this allows nanosecond granularity in message ids, which fixes the very regular message id collision errors


cpacia:
Good to know about them. @jackkleeman I think there's more places where timestamps are manually created.

jackkleeman:
Yes, I count another 20 or so. Will update this PR tomorrow

jackkleeman:
this should be good to go now

PR #469, opened 2017-03-20
Validate profile

cpacia:


tyler-smith:
Ideally these validation errors would all be exported so the validation can be used from other packages with error checking support but it's not a show-stopping issue.

core/profile.go
```diff
@@ -178,3 +187,118 @@ func (n *OpenBazaarNode) updateProfileCounts() error {
...
+		}
+		for _, l := range profile.ModInfo.Languages {
+			if len(l) > WordMaxCharacters {
+				return fmt.Errorf("Moderator language character length is greater than the max of %d", WordMaxCharacters)
+			}
+		}
+		if profile.ModInfo.Fee != nil {
+			if profile.ModInfo.Fee.FixedFee != nil {
+				if len(profile.ModInfo.Fee.FixedFee.CurrencyCode) > WordMaxCharacters {
+					return fmt.Errorf("Moderator fee currency code character length is greater than the max of %d", WordMaxCharacters)
```
rmisio: Should we be validating that the user is providing an actual valid currency code as opposed to some string of no more than 40 characters?

It would mean that the client and server have to be in sync on what the valid list is... but it would ensure the format is what both are expecting.

cpacia: We don't have a standard set of currency codes at the moment.

rmisio: @jjeffryes What do you think about getting some back-end stories in the backlog so the server could maintain a valid list of countries and currency codes and then validate that any related fields give real values?

Otherwise, as it is now, someone could enter 'pickles' for a currency and 'marmalade' for a country.

It would be a slight pain to have to keep server and client lists in sync, but I think it would be less so than the client having to deal with bogus values.

jjeffryes: We do have a list of valid country codes in the server. If they aren't being validated against, then they probably should be.

Adding a canonical list of currency codes to validate against would be a good idea.

PR #471, opened 2017-03-21
Don't close outbound streams

jackkleeman:
Apologies for yet another stream reuse fix. This prevents the closing of outbound streams after a message is sent.

### DIFF
--- a/net/service/messagesender.go
+++ b/net/service/messagesender.go
@@ -69,7 +69,7 @@ func (ms *messageSender) prep() error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


jackkleeman:
This has a failed check but it seems to be an unknown error running coveralls...

hoffmabc:
I went ahead and tried to restart it since it seemed like a borked Coveralls issue.

hoffmabc:
Ok @jackkleeman still failing so we can just wait a few hours and it'll probably resolve itself. It's not the code obviously.

PR #478, opened 2017-03-22
Push wallet transactions to websockets

cpacia:


cpacia:
Both of them don't cause any panics so I think we're OK. It's probably not appropriate to trigger an error up the chain of calls here as what they're doing is cosmetic. Though we could log the error. I'll do that later.

Issue #484, opened 2017-03-31
Could magic ids be deterministic?


jackkleeman:
Is it necessary for magic ids on outgoing offline messages to have a random component? Could they be deterministic based on the ciphertext hash, for example. This would mean, for example, that trying to post the exact same ciphertext to DHT twice would have the same magicID, which is probably desirable.

If there's support for this I can do a PR

cpacia:
I don't see any reason why they couldn't be.

PR #500, opened 2017-04-11
Use deterministic magic ids

jackkleeman:
See #484, this has magic ids derived from the message ciphertext hash for offline messages, and the peerID for moderator pointers.


JustinDrake:
The word `entropy` suggests randomness. What about `payload`?

PR #503, opened 2017-04-12
Sign ratings

cpacia:
When leaving a rating for an order that ended in a dispute, the moderator's signature on the rating key is included in the rating. Since the vendor signs the moderator key as part of the order confirmation, this allows us to validate a review for an order which a vendor never fulfilled.


hoffmabc:
@tyler-smith how we looking on this review?

qa/complete_direct_online.py
```diff
@@ -162,12 +162,17 @@ def run_test(self):
...
-            "Review": "I love it!"
+            "ratings": [
+                {
+                    "slug": slug,
+                    "overall": 4,
+                    "quality": 5,
+                    "description": 5,
+                    "customerService": 4,
+                    "deliverySpeed": 3,
+                    "Review": "I love it!"
```
tyler-smith: @cpacia Shouldn't "Review" be "review"?

PR #507, opened 2017-04-13
Listing format refactor

cpacia:


qa/mail.dump
```diff
@@ -0,0 +1,6 @@
+From: openbazaar@test.org
```
tyler-smith: Is this file used? If so I'm missing where.

PR #519, opened 2017-04-27
Add profile caching

cpacia:
Given the slow speed of ipns calls there are some places where we can probably
get away with loading a profile from cache.

This commit adds a `usecache` argument to the `FetchProfile()` function which will
load the last IPNS record from the datastore and use that in an IPFS cat call
(which should pull from the disk cache). At the same time a goroutine fires
fetching the most recent profile and updating the cache so that the next time
`FetchProfile()` is queried the cache will have the update. In this sense the `usecache`
argument returns a profile that is always one update behind (though given that
profiles don't update with high frequency most of the time it will be the latest).


drwasho:
🍾

PR #534, opened 2017-05-19
Mark orders as unread when chat message comes in

cpacia:
@tyler-smith can you take a look at this and see if you can think of any cleaner way to do this. 

Since we want orders to go to the top of the transactions list view when they are "unread" AND since we want them to be set as unread when they have unread chat messages, we need to mark an order as unread when chat messages for that order come in (so that the sql filters sort by unread work).

However, the chat messages only have an orderID and it doesn't tell me what table it's for. So it seems like unless there are some other major changes the only way to mark an order as unread when chat messages come in is to call `markAsRead` on each of the sales, purchases, and cases tables and let them error out if that order doesn't exist in the table.


jjeffryes:
@cpacia the client has been updated to use this PR, we're ready for you to merge it as soon as @tyler-smith has reviewed it.

rmisio:
@cpacia I'm still getting this error when marking a chat as read:
`POST http://localhost:4002/ob/markchatasread?subject=QmTyAyY45yUkStTc7NpdbqYtTZLqvowK6fYARPJdo6rkaz`

`sql: Scan error on column index 0: converting driver.Value type \u003cnil\u003e ('\u003cnil\u003e') to a int: invalid syntax`

rmisio:
@tyler-smith @cpacia Aside from the error above... is this one almost ready? We've mistakingly merged a client PR that's dependent on this one.

PR #536, opened 2017-05-30
Refactor notification format

cpacia:


jjeffryes:
I created https://github.com/OpenBazaar/openbazaar-desktop/issues/500 on the client repo to track the client side of this. Once that's done we can merge both of them together.

cpacia:
@jjeffryes I think you're going to need to change the format on the payment notification if you're relying on it already.

jjeffryes:
@cpacia yes, that's expected.

PR #560, opened 2017-06-29
Add optional pointer from gateway

cpacia:
Add a routine in the MessageRetrieve to fetch pointers from the gateway in addition
to the dht.


core/net.go
```diff
@@ -75,6 +81,27 @@ func (n *OpenBazaarNode) SendOfflineMessage(p peer.ID, k *libp2p.PubKey, m *pb.M
...
+			for _, g := range n.CrosspostGateways {
+				go func() {
+					dial := gonet.Dial
+					if n.TorDialer != nil {
+						dial = n.TorDialer.Dial
+					}
+					tbTransport := &http.Transport{Dial: dial}
+					client := &http.Client{Transport: tbTransport, Timeout: time.Minute}
+					client.Post(g.String()+"ipfs/providers", "application/x-www-form-urlencoded", bytes.NewReader(ser))
+				}()
```
tyler-smith: because the goroutine is async the value of `g` likely won't be what you expect. All the schedule goroutines could end up using the same value for `g` (the last one).

Make the func take g as a param so it's copied to its stack.

tyler-smith:
```
go func(g *url.URL) {
  // ...
}(g)
```

cpacia: good catch

PR #568, opened 2017-07-13
Segwit prep: refactor keys db

cpacia:
Most of the change is to the spvwallet lib, but it does push up a refactor of the keys db implementation. 

Keys used to be fetched from the db by giving the corresponding scriptPubKey. Since segwit introduces a new kind of scriptPubKey, and if a user resyncs the chain from seed we wont know which scriptPubKey to use. 

So this changes it to save the hash160 instead of the scriptPubKey since segwit addresses use the same hash160. 

Keys can now be fetched by giving the hash160, which makes the keys db agnostic to segwit. 

This will likely require resetting the data directory.

### DIFF
--- a/repo/db/keys.go
+++ b/repo/db/keys.go
@@ -15,16 +15,16 @@ type KeysDB struct {
 	lock sync.RWMutex
 }
 
-func (k *KeysDB) Put(scriptPubKey []byte, keyPath spvwallet.KeyPath) error {
+func (k *KeysDB) Put(scriptAddress []byte, keyPath spvwallet.KeyPath) error {
 	k.lock.Lock()
 	defer k.lock.Unlock()
 	tx, err := k.db.Begin()
 	if err != nil {
 		return err
 	}
-	stmt, _ := tx.Prepare("insert into keys(scriptPubKey, purpose, keyIndex, used) values(?,?,?,?)")
+	stmt, _ := tx.Prepare("insert into keys(scriptAddress, purpose, keyIndex, used) values(?,?,?,?)")
 	defer stmt.Close()
-	_, err = stmt.Exec(hex.EncodeToString(scriptPubKey), int(keyPath.Purpose), keyPath.Index, 0)
+	_, err = stmt.Exec(hex.EncodeToString(scriptAddress), int(keyPath.Purpose), keyPath.Index, 0)
 	if err != nil {
 		tx.Rollback()
 		return err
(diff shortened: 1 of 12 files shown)


repo/db/keys.go
```diff
@@ -15,16 +15,16 @@ type KeysDB struct {
...
-func (k *KeysDB) Put(scriptPubKey []byte, keyPath spvwallet.KeyPath) error {
+func (k *KeysDB) Put(scriptAddress []byte, keyPath spvwallet.KeyPath) error {
 	k.lock.Lock()
 	defer k.lock.Unlock()
 	tx, err := k.db.Begin()
 	if err != nil {
 		return err
 	}
-	stmt, _ := tx.Prepare("insert into keys(scriptPubKey, purpose, keyIndex, used) values(?,?,?,?)")
+	stmt, _ := tx.Prepare("insert into keys(scriptAddress, purpose, keyIndex, used) values(?,?,?,?)")
```
tyler-smith: This error should be checked or line 27 will panic.

repo/db/keys.go
```diff
@@ -33,16 +33,16 @@ func (k *KeysDB) Put(scriptPubKey []byte, keyPath spvwallet.KeyPath) error {
...
-func (k *KeysDB) ImportKey(scriptPubKey []byte, key *btcec.PrivateKey) error {
+func (k *KeysDB) ImportKey(scriptAddress []byte, key *btcec.PrivateKey) error {
 	k.lock.Lock()
 	defer k.lock.Unlock()
 	tx, err := k.db.Begin()
 	if err != nil {
 		return err
 	}
-	stmt, _ := tx.Prepare("insert into keys(scriptPubKey, purpose, used, key) values(?,?,?,?)")
+	stmt, _ := tx.Prepare("insert into keys(scriptAddress, purpose, used, key) values(?,?,?,?)")
```
tyler-smith: This error should be checked or line 45 will panic.

repo/db/keys.go
```diff
@@ -51,17 +51,17 @@ func (k *KeysDB) ImportKey(scriptPubKey []byte, key *btcec.PrivateKey) error {
...
-func (k *KeysDB) MarkKeyAsUsed(scriptPubKey []byte) error {
+func (k *KeysDB) MarkKeyAsUsed(scriptAddress []byte) error {
 	k.lock.Lock()
 	defer k.lock.Unlock()
 	tx, err := k.db.Begin()
 	if err != nil {
 		return err
 	}
-	stmt, _ := tx.Prepare("update keys set used=1 where scriptPubKey=?")
+	stmt, _ := tx.Prepare("update keys set used=1 where scriptAddress=?")
```
tyler-smith: This error should be checked or line 64 will panic.

Issue #188, opened 2016-10-31
Allow for multiple accepted currencies


JustinDrake:
The `Purchase` function in `order.go` checks `AcceptedCurrency`:

		if strings.ToLower(listing.Metadata.AcceptedCurrency) != strings.ToLower(n.Wallet.CurrencyCode()) {
			return "", "", 0, false, fmt.Errorf("Contract only accepts %s, our wallet uses %s", listing.Metadata.AcceptedCurrency, n.Wallet.CurrencyCode())
		}

I am assuming this is for future proofing, to allow more than BTC. If so, shouldn't we change it to `AcceptedCurrencies` (note plural) to allow for several accepted currencies?

The relevant line in `contracts.proto` would become

    repeated string acceptedCurrencies          = 5;

cpacia:
Multiple currencies would require multiple wallets. I tend to think running multiple wallets and handling all that logic is probably too much for an end user application. Possibly sometime in the future it could be handled but I don't have any plans for it in the short term.

JustinDrake:
@cpacia To be clear, I am not suggesting that we support multiple currencies. Instead, I am suggesting we give ourselves the option for multiple currencies in the future. The main change required is to make `acceptedCurrencies` a `repeated string` in the protobuf.

tyler-smith:
I think it makes sense to keep the protocol flexible where possible and reasonable, even if the reference client makes no use of it.

From the buyer's point of view we're already ensuring the accepted currency is correct. I don't see a reason we couldn't just check a slice of currencies instead. It's almost no change. From the vendor's standpoint there's no reason we need to have the feature surfaced as @JustinDrake says.

JustinDrake:
I still thing this is a good idea, and it's a very easy code change to make.

JustinDrake:
Implemented https://github.com/OpenBazaar/openbazaar-go/pull/584

PR #584, opened 2017-07-20
Change AcceptedCurrency to list

cpacia:
Change the AcceptedCurrency field to a list of AcceptedCurrencies so that the
contract can potentially support multiple coins in the future.


cpacia:
Addresses #188

core/listings.go
```diff
@@ -691,6 +691,17 @@ func validateListing(listing *pb.Listing) (err error) {
 	if len(listing.Metadata.Language) > WordMaxCharacters {
 		return fmt.Errorf("Language is longer than the max of %d characters", WordMaxCharacters)
 	}
+	if len(listing.Metadata.AcceptedCurrencies) == 0 {
+		return errors.New("At least one accepted currency must be provided")
```
tyler-smith: Making errors like this named (i.e. package-level vars) would be super cool but not required.

PR #610, opened 2017-08-01
Spvwallet - Improve blockchain handling

cpacia:
Adds an in-memory cache of 100 blocks to improve validation speeds and make it easier to calculate MTP in the future. 

It also fixes a bug calculating the common ancestor in certain types of reorgs.

### DIFF
--- a/vendor/github.com/OpenBazaar/spvwallet/blockchain.go
+++ b/vendor/github.com/OpenBazaar/spvwallet/blockchain.go
@@ -41,10 +41,14 @@ type Blockchain struct {
 }
 
 func NewBlockchain(filePath string, walletCreationDate time.Time, params *chaincfg.Params) (*Blockchain, error) {
+	hdb, err := NewHeaderDB(filePath)
+	if err != nil {
+		return nil, err
+	}
 	b := &Blockchain{
 		lock:   new(sync.Mutex),
 		params: params,
-		db:     NewHeaderDB(filePath),
+		db:     hdb,
 	}
 
 	h, err := b.db.Height()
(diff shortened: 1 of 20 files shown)


vendor/github.com/OpenBazaar/spvwallet/blockchain.go
```diff
@@ -199,6 +201,28 @@ func (b *Blockchain) calcRequiredWork(header wire.BlockHeader, height int32, pre
...
+		iterNode, err = b.db.GetHeader(header.BlockHash())
+		if err != nil {
+			return time.Time{}, err
+		}
+	}
+	timestamps = timestamps[:numNodes]
+	sort.Sort(timeSorter(timestamps))
+	medianTimestamp := timestamps[numNodes/2]
+	return time.Unix(medianTimestamp, 0), nil
+	return time.Now(), nil
```
tyler-smith: I don't think this is intentional.

vendor/github.com/OpenBazaar/spvwallet/headers.go
```diff
@@ -63,11 +74,15 @@ var (
 )
 
 func NewHeaderDB(filePath string) *HeaderDB {
+	if !strings.Contains(filePath, ".bin") {
+		filePath = path.Join(filePath, "headers.bin")
+	}
 	h := new(HeaderDB)
-	db, _ := bolt.Open(path.Join(filePath, "headers.bin"), 0644, &bolt.Options{InitialMmapSize: 5000000})
+	db, _ := bolt.Open(filePath, 0644, &bolt.Options{InitialMmapSize: 5000000})
```
tyler-smith: If this error is ignored it will panic on line 87

vendor/github.com/OpenBazaar/spvwallet/headers.go
```diff
@@ -303,3 +358,35 @@ func deserializeHeader(b []byte) (sh StoredHeader, err error) {
 	}
 	return sh, nil
 }
+
+type HeaderCache struct {
+	headers *ordered_map.OrderedMap
+	sync.Mutex
```
tyler-smith: Not an issue really, but a RWMutex would be good here instead of sync.Mutex.

PR #662, opened 2017-09-05
Fix bug calculating change value

cpacia:
The NewUnsignedTransaction in btcd doesn't work right. The "fee" it takes as an input is used both to calculate the change amount AND to determine if the change output is dust... which obviously is not correct as dust should use the network default relay fee not the amount paid.

So this duplicates the fuction inside spvwallet and fixes the issue.

### DIFF
--- a/vendor/github.com/OpenBazaar/spvwallet/sortsignsend.go
+++ b/vendor/github.com/OpenBazaar/spvwallet/sortsignsend.go
@@ -598,7 +598,7 @@ func (w *SPVWallet) buildTx(amount int64, addr btc.Address, feeLevel wallet.FeeL
 	if optionalOutput != nil {
 		outputs = append(outputs, optionalOutput)
 	}
-	authoredTx, err := txauthor.NewUnsignedTransaction(outputs, btc.Amount(feePerKB), inputSource, changeSource)
+	authoredTx, err := NewUnsignedTransaction(outputs, btc.Amount(feePerKB), inputSource, changeSource)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


vendor/github.com/OpenBazaar/spvwallet/sortsignsend.go
```diff
@@ -629,6 +629,66 @@ func (w *SPVWallet) buildTx(amount int64, addr btc.Address, feeLevel wallet.FeeL
...
 }
 
+func NewUnsignedTransaction(outputs []*wire.TxOut, feePerKb btc.Amount, fetchInputs txauthor.InputSource, fetchChange txauthor.ChangeSource) (*txauthor.AuthoredTx, error) {
+
+	var targetAmount btc.Amount
+	for _, txOut := range outputs {
+		targetAmount += btc.Amount(txOut.Value)
+	}
+
+	estimatedSize := EstimateSerializeSize(1, outputs, true, P2PKH)
```
tyler-smith: Why P2PKH instead of P2SWH? Am I misunderstand how this works?

vendor/github.com/OpenBazaar/spvwallet/sortsignsend.go
```diff
@@ -629,6 +629,66 @@ func (w *SPVWallet) buildTx(amount int64, addr btc.Address, feeLevel wallet.FeeL
...
+	targetFee := txrules.FeeForSerializeSize(feePerKb, estimatedSize)
+
+	for {
```
tyler-smith: It seems like this could be done without a potentially infinite loop. It seems that the call to `fetchInputs` will only ever get two different inputs and should only return two different values.

vendor/github.com/OpenBazaar/spvwallet/sortsignsend.go
cpacia: This function only spends from p2pkh atm.

vendor/github.com/OpenBazaar/spvwallet/sortsignsend.go
cpacia: This is how it is written in btcd. I'll probably keep it for now and consider refactoring later.

PR #680, opened 2017-09-11
Fix bug validating dispute contracts

cpacia:


hoffmabc:
Will this fix contracts already in effect?

PR #689, opened 2017-09-15
Allow & in sanitized strings

cpacia:


jjeffryes:
@cpacia when I try to run this branch, I get this error:

api\sanitize.go:29: cannot use i (type interface {}) as type []byte in argument to bytes.Replace: need type assertion

PR #694, opened 2017-09-20
Refactor import listings function

cpacia:


core/import.go
```diff
@@ -17,32 +17,50 @@ import (
...
 
-	respCh := make(chan Result, len(records)-1)
-	defer close(respCh)
+	indexLock := new(sync.Mutex)
+
+	for {
+		select {
+		case err := <-errChan:
+			return err
+		case <-done:
```
cpacia: If the last 5 goroutines are in flight and one finishes, next time through the loop it will try to read from the CSV and get an end of file error then `done` will be called. I suppose it's possible `done` can be called before the remaining four goroutines finish.

How can we make sure the remaining four goroutines finish so we can check for errors before returning.?

PR #701, opened 2017-09-22
IPFS Rebase

cpacia:


cpacia:
Currently I save IPNS records in the database and if the IPNS query fails because the data dropped out of the DHT then it pull the last record from the database. 

But this makes it hard for the search to know if listings have dropped out of the DHT since it always returns from the database. So this is just an option to disable the cache.

PR #708, opened 2017-09-29
Add gencerts command

cpacia:
Add a new command line command to automatically generate an ssl certificate.
The command also has an option to automatically set up ssl in the config file.


cpacia:
Can't get the client to accept the generated certs.

PR #710, opened 2017-09-29
Implement classified listing type

cpacia:


JustinDrake:
I personally really like the idea of classifieds.

hoffmabc:
There is an OBIP that can be talked through as well.

https://github.com/OpenBazaar/obips/pull/5

michaelfolkson:
I agree with @hoffmabc. No problem with idea but needs discussion. I'm really struggling to keep up with what priorities are. Are we at the stage where we start discussion on this new feature or is the discussion postponed until a later date? My guess is we are starting to draft OBIPs but they won't be seriously discussed until after full release?

michaelfolkson:
We need to discuss how the variable `Listing_Metadata_ContractType_name` in `pb/contracts.pb.go` gets changed and what it represents. My understanding is that `3: CROWD_FUND` isn't being used at all and there are no plans to implement anything in the near future. The contract code for CROWD_FUND would be completely distinct from say `0: PHYSICAL_GOOD` whilst `1: DIGITAL_GOOD` shares much of the same code as `0:PHYSICAL_GOOD`. At the moment I'm not sure why classified listings need their own contract type. You could have both classified physical goods and classified digital goods? In which case CLASSIFIED needs to be embedded into both of those rather than having its own contract type?

placer14:
Archived for later discussion.

PR #711, opened 2017-10-01
Refactor data sharing

cpacia:


hoffmabc:
This has a potential complaint from DUO on deprecating this feature. I'm not sure how Mubiz or others feel either, but we should pause to get feedback.

@JustinDrake this is a concern right?

JustinDrake:
In the context of an OpenBazaar client running in the browser, Duo has built infrastructure to cache offline messages. Right now we are assuming that the Duo gateway will be notified of offline message pointers. I want to confirm that the offline pointers will still be broadcasted to the gateways.

cpacia:
Pointers are pushed using the DHT putProvider call instead of http. But, yes, they are still sent.

PR #727, opened 2017-10-11
[WIP] Update mobile code and add badger-ds support

hoffmabc:
Got some questions here for @cpacia.

We can't merge this immediately because it changes the datastore config, but I need to figure out how to best do this. Should I create a mobile config go file and reference that from node.go? Would that be best so as not to conflict with the desktop server? Or would you like to do something else? Also the dsync pass for the compiler is only relevant for mobile and doesn't need to be skipped on desktop so we'll need to think about that perhaps in the future.

### DIFF
--- a/mobile/node.go
+++ b/mobile/node.go
@@ -64,8 +64,30 @@ type Node struct {
 	ipfsConfig *ipfscore.BuildCfg
 	apiConfig  *repo.APIConfig
 }
+type Mobile struct { }
 
-func NewNode(config NodeConfig) (*Node, error) {
+
+func NewOB(repoPath string, authenticationToken string, testnet bool, userAgent string, walletTrustedPeer string, password string, mnemonic string) *Node {
+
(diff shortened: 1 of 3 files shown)


mobile/node.go
```diff
@@ -64,8 +64,30 @@ type Node struct {
 	ipfsConfig *ipfscore.BuildCfg
 	apiConfig  *repo.APIConfig
 }
+type Mobile struct { }
 
-func NewNode(config NodeConfig) (*Node, error) {
+
+func NewOB(repoPath string, authenticationToken string, testnet bool, userAgent string, walletTrustedPeer string, password string, mnemonic string) *Node {
```
hoffmabc: Objective-C does not like allocating structure objects to send into these bridge methods so I had to create a standard proxy function.

cpacia:
Doesn't look like this helped that much. Also needs a major rebase.

PR #729, opened 2017-10-11
Update Makefile for ios_framework

drwasho:
Also ignore .DS_Store files for mac


drwasho:
Upstream was incorrect

PR #734, opened 2017-10-15
Fix issues with misspell

groksrc:
This fixes the remaining issues with misspell. It also picks up a couple of changes applied by my editor to reorder the imports correctly.


cpacia:
Thanks

PR #736, opened 2017-10-19
Add additional shipping rule validation

cpacia:
Max rules must be less than MaxListItems (30) and shipping rule ranges must
not overlap.


jjeffryes:
I think this looks fine. It should cover any overlaps.

PR #742, opened 2017-10-23
Unfunded order resync

cpacia:
Adds a db migration to add a column to the sales table to track if a resync is needed.

Each hour it will loop through orders that:
1) Are unfuned
2) The order timestamp is before the received time (indicating we might have synced passed the payment).
3) Have never been subjected to a resync.

If any orders meet the condition the wallet will resync the blockchain from the timestamp of the earliest order - 24 hours. 

This is only done for the spvwallet and not bitcoind because bitcoind does not allow rescan from a height/timestamp and it's a very heavyweight operation (takes longer than the 1 hour resync interval).


jjeffryes:
This is a nice addition.

cpacia:
@jjeffryes you guys should probably remove that pop up notification from the client.

jjeffryes:
@cpacia once this is merged we'll do that.

repo/db/sales.go
```diff
@@ -273,3 +278,41 @@ func (s *SalesDB) Count() int {
...
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	var ret []repo.UnfundedSale
+	rows, err := s.db.Query(`select orderID, timestamp from sales where state=? and needsSync=?`, 1, 1)
+	if err != nil {
+		return ret, err
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var orderID string
```
tyler-smith: Not a big deal here probably, but in general if you keep these variables out of the loop you won't have to allocate new memory each cycle.

PR #744, opened 2017-10-24
Prevent overlapping shipping rule range for quantity discount

### LINKED ISSUE #743: Shipping rule should not allow a min greater than the max
- minRange should not be larger than maxRange

cpacia:
closes #743

### DIFF
--- a/core/listings.go
+++ b/core/listings.go
@@ -968,7 +968,9 @@ func validateListing(listing *pb.Listing, testnet bool) (err error) {
 				return fmt.Errorf("Shipping rules exceeds max of %d", MaxListItems)
 			}
 			for i, rule := range shippingOption.ShippingRules.Rules {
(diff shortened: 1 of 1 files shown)


core/listings.go
```diff
@@ -968,7 +968,9 @@ func validateListing(listing *pb.Listing, testnet bool) (err error) {
 				return fmt.Errorf("Shipping rules exceeds max of %d", MaxListItems)
 			}
 			for i, rule := range shippingOption.ShippingRules.Rules {
-				if (shippingOption.ShippingRules.RuleType == pb.Listing_ShippingOption_ShippingRules_FLAT_FEE_QUANTITY_RANGE || shippingOption.ShippingRules.RuleType == pb.Listing_ShippingOption_ShippingRules_FLAT_FEE_WEIGHT_RANGE) && rule.MaxRange <= rule.MinRange {
+				if (shippingOption.ShippingRules.RuleType == pb.Listing_ShippingOption_ShippingRules_FLAT_FEE_QUANTITY_RANGE ||
```
rmisio: > rule.MaxRange <= rule.MinRange

Should we not allow the equal to? What if you wanted to give a 1.00 flat fee shipping for a quantity of 3 items? You would set the min and max both to 3. Is that not a valid scenario?


Why are we doing this validation only for certain types? In what scenario would you want a max < min for COMBINED_SHIPPING_ADD || COMBINED_SHIPPING_SUBTRACT?

PR #752, opened 2017-10-30
Tails

cpacia:
Refactor of the go-onion-transport to allow for optionally using a manually created hidden service.


n6p0u1:
Is this issue still being worked on? It would be amazing if I could use OpenBazaar in Tails.

Issue #761, opened 2017-11-11
API Preflight Headers Need Adjustment


christroutner:
I've got SSL setup between my remote server and local client, as well as API login and password working. I can connect just fine with my OB client. Trying to make an AJAX call to the OB server however, doesn't work. 

Here is a snippet of the AJAX request I'm making, using jQuery:
```
      var settings = {
        "url": "https://104.207.555.555:4002/ob/config",
        "method": "GET",
        "headers": {'Authorization': readyCredential},
      }

      $.ajax(settings).done(function (response) {
        debugger;
        console.log(response);
      });
```

According to [this stack overflow thread](https://stackoverflow.com/questions/8685678/cors-how-do-preflight-an-httprequest) the Server should be responding with `GET` listed in the `access-control-allow-methods`. I believe that is the reason my GET call to `/ob/config` is failing.

It's important to keep in mind that this is a *preflight* header, not the actual request. That's why it's listed as *OPTIONS* and not *GET*.

## Request Headers
```
:authority:104.207.158.166:4002
:method:OPTIONS
:path:/ob/config
:scheme:https
accept:*/*
accept-encoding:gzip, deflate, br
accept-language:en-US,en;q=0.9
access-control-request-headers:authorization
access-control-request-method:GET
origin:http://127.0.0.1:56238
user-agent:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.89 Safari/537.36
```

## Response Headers
```
access-control-allow-headers:Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization
access-control-allow-methods:PUT,POST,DELETE
access-control-allow-origin:*
content-length:15
content-type:text/plain; charset=utf-8
date:Sat, 11 Nov 2017 00:53:53 GMT
status:403
```

christroutner:
I think I found the source of the issue, but I'm not good enough with GO to make the edits necessary. It looks to me that everything controlling this issue takes place in [jsonapi.go](https://github.com/OpenBazaar/openbazaar-go/blob/0e44b51d2efcfe499a3a95777882a54c0a6ec327/api/jsonapi.go).

Line 108 sets the value for `Access-Control-Allow-Methods`. This should be expanded to include `GET` and `OPTIONS`.

[...]

PR #763, opened 2017-11-13
Allow CORS access to the API by fixing Prelight Header Handling.

christroutner:
The current implementation of the JSON API does not handle true Cross Origin Resource Sharing (CORS) support. 

When making an AJAX call using Basic Authentiation, a preflight OPTIONS header is sent before the actual GET request. An OPTIONS header does not/can not contain the Authorization Token. Only the GET request can contain the token. The current implementation does not respond to CORS OPTIONS or GET calls.

There are two changes in the PR:

* Change the `Access-Control-Allow-Methods` to allow GET and OPTIONS HTTP methods.

* Respond to OPTIONS headers with an HTTP status of `200 - OK`. This permits the client to *then* send the actual GET request with the authorization token.

The changes in this PR solve Issue #761

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -105,7 +105,7 @@ func (i *jsonAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if i.config.Cors != nil {
 		w.Header().Set("Access-Control-Allow-Origin", *i.config.Cors)
-		w.Header().Set("Access-Control-Allow-Methods", "PUT,POST,DELETE")
+		w.Header().Set("Access-Control-Allow-Methods", "PUT,POST,DELETE,GET,OPTIONS")
 		w.Header().Set("Access-Control-Allow-Headers", "Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization")
 	}
 
@@ -127,6 +127,13 @@ func (i *jsonAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 		} else {
+      
+      if r.Method == "OPTIONS" {
+        w.WriteHeader(http.StatusOK)
+        fmt.Fprint(w, "200 - OK")
+        return
+      }
+      
 			username, password, ok := r.BasicAuth()
 			h := sha256.Sum256([]byte(password))
 			password = hex.EncodeToString(h[:])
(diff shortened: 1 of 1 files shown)


rmisio:
@cpacia This is a slight tangent here, but isn't allowing the server to be run on CORS a big security risk? For example, if I set a stand-alone node to be allow cross domain requests any site I visit would potentially have access to the api.

I think Chrome prevents cross domain calls on localhost, but I'm not sure about other browsers.

At the very least, if CORS is on, does the server require some type of authentication to be specified as well? It would seem really risky to me to allow the server to be started without authentication and CORS on.

christroutner:
@rmisio The person still needs to use Basic Authentication in order use the API. Everything in [the security doc](https://github.com/OpenBazaar/openbazaar-go/blob/0e44b51d2efcfe499a3a95777882a54c0a6ec327/docs/security.md) still applies.

Also, CORS can be limited to specific domains or IP addresses. It doesn't just have to be "*", which allows everything.

The main change being implemented by the PR is to fix the current implementation, so that it conforms preflight requests, [as per the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).

tyler-smith:
If exposing GET and OPTIONS requests to CORS is a vulnerability over only allowing POST, PUT< and DELETE, we have a more serious issue.

This change looks good to me.

rmisio:
> If exposing GET and OPTIONS requests to CORS is a vulnerability over only allowing POST, PUT< and DELETE, we have a more serious issue.

I wasn't asking that. I was questioning if allowing CORS at all without requiring auth (if that's what we indeed do) is a security issue?

tyler-smith:
Yes absolutely if the settings allow any domain  without auth it's a giant open door.

cpacia:
I think this is OK. Just have to run `gofmt`

christroutner:
Please make any changes you'd like. I don't even know Go. I just hacked it together from the code I could find and using Google. But I tested it on my remote server and I can access the API now. And it appears to be working as per the CORS spec.

christroutner:
Hey @cpacia are you referring to line 133? I should use `gofmt` instead of `fmt`?

hoffmabc:
@christroutner just run gofmt in the code directory from terminal so that it properly formats your code submission and Travis doesn't barf.

https://blog.golang.org/go-fmt-your-code

hoffmabc:
If you do this, then we can merge the code.

PR #813, opened 2018-01-06
Fix bugs in Bitcoin Cash singing lib

cpacia:


net/retriever/retriever.go
```diff
@@ -406,11 +406,11 @@ func (m *MessageRetriever) handleMessage(env pb.Envelope, id *peer.ID) error {
 		// Get the peer ID from the public key
 		pubkey, err := libp2p.UnmarshalPublicKey(env.Pubkey)
 		if err != nil {
-			return nil
+			return err
 		}
 		i, err := peer.IDFromPublicKey(pubkey)
 		if err != nil {
```
hoffmabc: Are these the only suppressed errors in the retriever?

cpacia: yeah, but this just makes it passed up to the log. I don't think either of these two errors were firing but I guess we'll see.

PR #822, opened 2018-01-12
Make sure each offline message is only downloaded once

cpacia:


hoffmabc:
Can you explain why would messages be out of order? We’re we not handling this before? @cpacia

cpacia:
The should have been processed in order but some of the recent changes made it possible for them to be out of order.

PR #824, opened 2018-01-14
Create command to convert wallet type

cpacia:


cmd/convert.go
```diff
@@ -0,0 +1,367 @@
...
+		if err != nil {
+			return err
+		}
+		sl := new(pb.SignedListing)
+		err = jsonpb.UnmarshalString(string(file), sl)
+		if err != nil {
+			return err
+		}
+		sl.Listing.Metadata.AcceptedCurrencies = []string{currencyCode}
+
```
ab10460ef3: // Clear old moderators 
`sl.Listing.moderators = []`

ab10460ef3:
Old moderators are still associated with the store and end up in the listings when they are re-saved.  

1. Created a listing with BTC added 5 moderators. 
2. Converted store, which removed moderators from listing.
3. Re-saved listing, BTC moderators were added back.

In checkout, there is no option to select these moderators, but there is no way to removed BTC moderators from a converted store.

cmd/convert.go
```diff
@@ -0,0 +1,368 @@
...
+	identityKey, err := sqliteDB.Config().GetIdentityKey()
+	if err != nil {
+		PrintError(err.Error())
+		return err
+	}
+	identity, err := ipfs.IdentityFromKey(identityKey)
+	if err != nil {
+		return err
+	}
+
```
ab10460ef3: `	// Check for old moderators
	settings, err := sqliteDB.Settings().Get()
	if err != nil && err != db.SettingsNotSetError {
		log.Error(err)
		return err
	}

	if len(*settings.StoreModerators) > 0 {
        fmt.Println("Opps!, it looks like you still have moderators using your old currency")
        fmt.Println("You'll have to remove them manually from their home pages, they will not appear in the client.")
        for _, mid := range *settings.StoreModerators {
            fmt.Printf("  ob://%s/home  \n", mid)
		}
	}`

I couldn't see a way to easily modify the settings.  It would be best just to remove them automatically.

cpacia: Gotcha. I think they should still be able to remove them in settings but you're right it probably should clear out of the settings.

ab10460ef3: ATM. In all UI lists, moderators are filtered by accepted currency. If the store is associated with several moderators of a different currency, they do not appear in the moderator list in settings, nor in available moderators in checkout.

The only way to I've found to remove moderators of a different currency is from their home page.

FYI it is also possible to add a moderator of a different currency from their homepage. Which is not a server problem

Issue #843, opened 2018-01-25
Changes to the 45 Day Timeout


jjeffryes:
Add the new features discussed in https://github.com/OpenBazaar/openbazaar-desktop/issues/923

- When a dispute is started, reset the timeout to 45 days.
- Provide notifications to the buyer at 15, 40, and 44 days.
- Provide notifications to the buyer and the moderator at 15, 40, and 44 days when a dispute is started.
- Provide a notification to the buyer and seller when the timeout has been reached.
- The api should prevent opening a dispute after the 45 day timeout has expired.
- The api should prevent resolving a dispute after the additional 45 day timeout (from when the dispute was opened) expires.
- the releaseescrow api should be updated to allow releasing the funds when both the order state is Disputed and the dispute escrow timeout has expired. Currently, it only allows it if the order state is Fulfilled.
- The wallet GETBalance api should return the current block height.
- Provide a socket message to the buyer when the vendor has claimed the payment (successfully called ob/releaseescrow).
- the POSTOrderComplete api should be updated to allow the buyer to complete the order (i.e. leave a review) when the order is in the PAYMENT_FINALIZED state.
- Nodes using the ZEC wallet should return an error when attempting to POSTReleaseEscrow saying that releasing escrow is not supported for ZEC transactions.

New Notification Schema

`buyerDisputeTimeout`:  for buyer at 15, 40, 44 and 45 days that you have X remaining days to open a dispute
```
"notification": {
  "notificationId": "idstring", //string, guid
  "orderId": "12345", //string, points to purchase ID in DB
  "type": "buyerDisputeTimeout", //string, always the same for this type
  "expiresIn": 0, //int, seconds before timeout
  "thumbnail": {
      "small": "zb2rhZhsmsd1nehk1GXD2zxF49xsYgVH7Lap4bVLmG1LJ2SKs", //string, hash to content
      "tiny": "zb2rhX9MsdkausCnYD4LH9Hjw2hQZmZ5S1fQAQ6DjBhoqjq2D", //string, hash to content
  }
}
```

`vendorDisputeTimeout`:  for vendor at 45 days that they can now claim funds
```
"notification": {
  "notificationId": "idstring", //string, guid
  "orderId": "12345", //string, points to sale ID in DB
  "type": "vendorDisputeTimeout", //string, always the same for this type
  "expiresIn": 0, //int, seconds before timeout
  "thumbnail": {
[...]

PR #852, opened 2018-02-06
[#843] Guard escrow during dispute

placer14:
This abstracts some methods for testing which mode the node is running in. This begs the question whether the modes are mutually exclusive or not. Then we use the abstractions to protect escrow from being released while the order is in dispute.

### DIFF
--- a/core/order.go
+++ b/core/order.go
@@ -513,7 +513,7 @@ func (n *OpenBazaarNode) createContractWithOrder(data *PurchaseData) (*pb.Ricard
 			if err := validateVendorID(sl.Listing); err != nil {
 				return nil, err
 			}
-			if err := validateListing(sl.Listing, n.TestNetworkEnabled()); err != nil {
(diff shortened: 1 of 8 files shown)


core/order.go
```diff
@@ -514,11 +513,7 @@ func (n *OpenBazaarNode) createContractWithOrder(data *PurchaseData) (*pb.Ricard
 			if err := validateVendorID(sl.Listing); err != nil {
 				return nil, err
 			}
-			testnet := false
-			if n.Wallet.Params().Name != chaincfg.MainNetParams.Name {
-				testnet = true
-			}
-			if err := validateListing(sl.Listing, testnet); err != nil {
+			if err := validateListing(sl.Listing, n.TestNetworkEnabled()); err != nil {
```
cpacia: This line needs to be `n.TestNetworkEnabled() || n.RegressionTestEnabled()` to make it pass the QA test.

cpacia:
Everything else looks good

core/order.go
placer14: That does not appear to be enough to get QA to pass. Have you tried it?

PR #856, opened 2018-02-08
Extract modelStore and Queryable interface

placer14:


repo/datastore.go
```diff
@@ -13,24 +14,31 @@ import (
...
+	Settings() ConfigurationStore
+	Inventory() InventoryStore
+	Purchases() PurchaseStore
+	Sales() SaleStore
+	Cases() CaseStore
+	Chat() ChatStore
+	Notifications() NotificationStore
+	Coupons() CouponStore
+	TxMetadata() TransactionMetadataStore
+	ModeratedStores() ModeratedStore
```
placer14: These names are up for discussion, but typically I try to call Interfaces a verb-like name to indicate the sort of actions it can take. Unfortunately, because these are interfaces for exactly one type of object (the interface isn't shared across different structs) the name for these end up being a little more specific indicating the type of Store they are. Naming things is hard.

repo/datastore.go
```diff
@@ -13,24 +14,31 @@ import (
...
 	Close()
 }
 
+type Queryable interface {
+	BeginTransaction() (*sql.Tx, error)
+	PrepareQuery(string) (*sql.Stmt, error)
+	PrepareAndExecuteQuery(string, ...interface{}) (*sql.Rows, error)
+	ExecuteQuery(string, ...interface{}) (sql.Result, error)
+}
+
```
placer14: This is how we will share future shared behavior across the Stores. Eventually, we'll migrate things like Put into here once we sort out the typecasting. We would approach this in steps and this is the foundation.

repo/datastore.go
```diff
@@ -341,3 +373,28 @@ type ModeratedStores interface {
...
+type UnspentTransactionOutputStore interface {
+	Queryable
+	wallet.Utxos
+}
+
+type WatchedScriptStore interface {
+	Queryable
+	wallet.WatchedScripts
+}
```
placer14: This will let us abstract our Datastore internals from the wallet interface... making it easier to decouple these packages later.

repo/db/cases.go
```diff
@@ -12,8 +12,11 @@ import (
 )
 
 type CasesDB struct {
-	db   *sql.DB
-	lock *sync.Mutex
+	modelStore
+}
+
+func NewCaseStore(db *sql.DB, lock *sync.Mutex) repo.CaseStore {
+	return &CasesDB{modelStore{db, lock}}
```
placer14: `modelStore` acts as a base struct which we can capture all of the database-specific internals

repo/db/db.go
```diff
@@ -16,25 +16,25 @@ var log = logging.MustGetLogger("db")
...
+	settings        repo.ConfigurationStore
+	inventory       repo.InventoryStore
+	purchases       repo.PurchaseStore
+	sales           repo.SaleStore
+	cases           repo.CaseStore
+	chat            repo.ChatStore
+	notifications   repo.NotificationStore
+	coupons         repo.CouponStore
+	txMetadata      repo.TransactionMetadataStore
+	moderatedStores repo.ModeratedStore
```
placer14: No more wallet interfaces to worry about here.

placer14:
This is ready to merge.

cpacia:
lgtm

PR #860, opened 2018-02-09
Handel errors processing offline orders

cpacia:


net/service/handlers.go
```diff
@@ -245,32 +247,48 @@ func (service *OpenBazaarService) handleOfflineRelay(p peer.ID, pmes *pb.Message
...
 
-	err = service.node.ValidateOrder(contract, !offline)
+	orderId, err = service.node.CalcOrderId(contract.BuyerOrder)
 	if err != nil {
 		return errorResponse(err.Error()), err
 	}
+
+	pro, _ := service.node.GetProfile()
+	if !pro.Vendor {
+		return errorResponse("The vendor turn his store off and is not accepting orders at this time"), errors.New("Store is turned off")
```
rmisio: turn => turned

rmisio:
@cpacia What can I do to make an order go into the PROCESSING_ERROR state, so that I could see what the server is sending the client, including the error string?

qa/purchase_offline_error.py
```diff
@@ -0,0 +1,129 @@
...
+        api_url = alice["gateway_url"] + "ob/shutdown"
+        requests.post(api_url, data="")
+        time.sleep(4)
+
+        # bob send order
+        with open('testdata/order_direct.json') as order_file:
+            order_json = json.load(order_file, object_pairs_hook=OrderedDict)
+        order_json["items"][0]["listingHash"] = listingId
+
+        # set empty shipping address to trigger error
```
rmisio: If an empty shipping address would produce a PROCESSING_ERROR, shouldn't the purchase API be rejecting that purchase with an error in the first place?

rmisio:
Don't we need to update `POSTOrderCancel` to allow the buyer to cancel unmoderated payments in this state?

rmisio:
Similar to above, wouldn't we need to update POSTOpenDispute to allow a buyer to open a dispute if the order is moderated and in a PROCESSING_ERROR state?

rmisio:
@cpacia Currently, the server will not allow the buyer to cancel a partially funded order. Is that by design?

Currently, it fails with `order must be PENDING or partially funded and only a direct payment to cancel`.

Either the error message or the behavior are incorrect.

cpacia:
> Currently, the server will not allow the buyer to cancel a partially funded order. Is that by design?

Correct

rmisio:
> Currently, the server will not allow the buyer to cancel a partially funded order. Is that by design?

> Correct

Ok, can we update the inaccurate error message, which implies that "partially funded" can be cancelled.

"order must be PENDING or partially funded and only a direct payment to cancel"

rmisio:
> I think this addresses all those issues. Anything else @rmisio ?

I'll get back to you. I need to test the order flow when an order is canceled / disputed after being in the PROCESSING_ORDER order state and make sure there are no surprises.

rmisio:
> order must be PENDING or PROCESSING_ERROR and only a direct payment to cancel

Based on what I'm observing, if we want the error message to reflect the behavior, it should say:

order must be PENDING or a partially funded PROCESSING_ERROR and only a direct payment to cancel.

api/jsonapi.go
```diff
@@ -2023,8 +2023,8 @@ func (i *jsonAPIHandler) POSTOpenDispute(w http.ResponseWriter, r *http.Request)
 		ErrorResponse(w, http.StatusBadRequest, "Order must be either AWAITING_FULFILLMENT or FULFILLED to start a dispute")
 		return
 	}
-	if !isSale && (state != pb.OrderState_AWAITING_FULFILLMENT && state != pb.OrderState_PENDING && state != pb.OrderState_FULFILLED) {
-		ErrorResponse(w, http.StatusBadRequest, "Order must be either AWAITING_FULFILLMENT, PENDING, or FULFILLED to start a dispute")
+	if !isSale && !(state == pb.OrderState_AWAITING_FULFILLMENT || state == pb.OrderState_PENDING || state == pb.OrderState_FULFILLED || state == pb.OrderState_PROCESSING_ERROR) {
```
rmisio: If the state is PROCESSING_ERROR, shouldn't we also ensure the order is at least partially funded? In other words, do we want to let them open a dispute if the order is not funded at all?

core/order.go
```diff
@@ -380,7 +390,15 @@ func (n *OpenBazaarNode) Purchase(data *PurchaseData) (orderId string, paymentAd
...
-				return "", "", 0, false, fmt.Errorf("Vendor rejected order, reason: %s", string(resp.Payload.Value))
+				var errStr string
+				errMsg := new(pb.Error)
+				err := ptypes.UnmarshalAny(resp.Payload, errMsg)
+				if err == nil {
+					errStr = errMsg.ErrorMessage
+				} else { // For backwards compatibility check for a string payload
+					errStr = string(resp.Payload.Value)
+				}
+				return "", "", 0, false, fmt.Errorf("Vendor rejected order, reason: %s", errStr)
```
placer14: The second time you write the same exact code should be an indication that it should be extracted into its own function.

`str, err := extractErrorString(resp)`

core/order.go
```diff
@@ -549,6 +571,19 @@ func (n *OpenBazaarNode) createContractWithOrder(data *PurchaseData) (*pb.Ricard
 			return nil, fmt.Errorf("Contract only accepts %s, our wallet uses %s", listing.Metadata.AcceptedCurrencies[0], n.Wallet.CurrencyCode())
 		}
 
+		// Make sure shipping fields are filled if the order contains a physical good
+		if containsPhysicalGood && !(n.TestNetworkEnabled() || n.RegressionNetworkEnabled()) {
```
placer14: I would like you to stop using this pattern:
```
var thing bool
for {
  if condition {
    thing = true
  }
}
```

The mutation is difficult to reason about and creates unexpected bugs when you forget certain cases (which are spread over a 50-100 lines of code). Instead, these variables want to be a function which accepts the dependancies of that condition.

`containsPhysicalGood(sl)`
```
func containsPhysicalGoods(sl) bool {
  if sl.Listing.Metadata.ContractType == pb.Listing_Metadata_PHYSICAL_GOOD {
    return true
  }
  return false
}

Comment (placer14) [reply core/order.go]: Or even better:
`sl.containsPhysicalGood()`

File: core/order.go [hunk 6]
```diff
@@ -1028,9 +1063,10 @@ collectListings:
 	}
 
 	// Validate the each item in the order is for sale
+	knownListings := true
 	for _, listing := range contract.VendorListings {
 		if !n.IsItemForSale(listing) {
-			return errors.New("Contract contained item that is not for sale")
+			knownListings = false
```
placer14: Again, this condition depends on one thing `n`. Instead, let's do `n.AnyListings()` or `n.HasListings()`.

core/order.go
```diff
@@ -1183,6 +1219,10 @@ collectListings:
 	if err != nil {
 		return err
 	}
+
+	if !knownListings {
+		return UnknowListingError
+	}
```
placer14: And maybe since the only usage so far is a negative, maybe it should be `n.HasNoListings()` or maybe what this REALLY depends on in the domain and name it `n.IsVendor()` but I don't know if that's too broad for this case. This is how we can capture domain knowledge without having to pass it around by word-of-mouth or trying to keep up with out-of-date docs.

net/retriever/retriever.go
```diff
@@ -58,14 +59,14 @@ type offlineMessage struct {
 	env  pb.Envelope
 }
 
-func NewMessageRetriever(db repo.Datastore, ctx commands.Context, node *core.IpfsNode, bm *net.BanManager, service net.NetworkService, prefixLen int, pushNodes []peer.ID, dialer proxy.Dialer, sendAck func(peerId string, pointerID peer.ID) error) *MessageRetriever {
+func NewMessageRetriever(db repo.Datastore, ctx commands.Context, node *core.IpfsNode, bm *net.BanManager, service net.NetworkService, prefixLen int, pushNodes []peer.ID, dialer proxy.Dialer, sendAck func(peerId string, pointerID peer.ID) error, sendError func(peerId string, k *libp2p.PubKey, errorMessage pb.Message) error) *MessageRetriever {
```
placer14: Option struct as an argument to this function please? This function signature is out of control. If you can't write a function call without using intellisense, there's probably cognitive overhead that can be reduced. The reason I like struct options is because they are self-documenting and (with good field names) make the arguments easier to understand without referring to how they were made.

net/retriever/retriever.go
```diff
@@ -58,14 +59,14 @@ type offlineMessage struct {
...
 	dial := gonet.Dial
 	if dialer != nil {
 		dial = dialer.Dial
 	}
 	tbTransport := &http.Transport{Dial: dial}
 	client := &http.Client{Transport: tbTransport, Timeout: time.Second * 30}
-	mr := MessageRetriever{db, node, bm, ctx, service, prefixLen, sendAck, client, pushNodes, new(sync.Mutex), make(chan struct{}), make(chan struct{}, 5), new(sync.WaitGroup)}
+	mr := MessageRetriever{db, node, bm, ctx, service, prefixLen, sendAck, sendError, client, pushNodes, new(sync.Mutex), make(chan struct{}), make(chan struct{}, 5), new(sync.WaitGroup)}
```
placer14: A good practice for declaring struct literals is to use the field names if there are more than 2 or maybe 3 arguments at most. It makes changes to the code in the future cleaner (a field value changing does not mess with other fields because its on its own line), easier to use (as you don't mix up values, especially if you have two string fields next to each other and accidentally reverse them), and you get a better sense of what the object will do (since you can see the intentions of how that function will use the values passed in).

net/retriever/retriever.go
```diff
@@ -349,7 +350,7 @@ func (m *MessageRetriever) handleMessage(env pb.Envelope, addr string, id *peer.
 	}
 
 	// Dispatch handler
-	_, err := handler(*id, env.Message, true)
+	resp, err := handler(*id, env.Message, true)
 	if err != nil && err == net.OutOfOrderMessage {
```
placer14: Every single case here depends on `err != nil`. Pull that out of the condition to a separate level:
```
if err != nil {
  if otherConditions {
  }
}
```

net/retriever/retriever.go
```diff
@@ -360,6 +361,10 @@ func (m *MessageRetriever) handleMessage(env pb.Envelope, addr string, id *peer.
 		} else {
 			log.Errorf("Error serializing offline message %s for storage")
 		}
+	} else if env.Message.MessageType == pb.Message_ORDER && err != nil && resp != nil {
```
placer14: The net retriever probably shouldn't be aware of the structure of your messages. To hide this from the retriever, you can:
`env.IsOrderMessage()` 
If this is the sort of thing that's happening often and we're checking types of messages from outside the envelope, then the envelope should have `env.GetMessageType()` which returns a type which matches whatever `pb.Message_ORDER` is and then do your `==` check.

net/service/handlers.go
```diff
@@ -245,32 +247,48 @@ func (service *OpenBazaarService) handleOfflineRelay(p peer.ID, pmes *pb.Message
 
 func (service *OpenBazaarService) handleOrder(peer peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
 	offline, _ := options.(bool)
```
placer14: Can you help me understand why `options` is a bool inside the function but an interface{} outside the function? And maybe why it's named `options` and not `isOffline` or something? If we're overloading this function to accept multiple versions of options, then we should extract the behavior specific to this case and call from within the function. If this is a handler or more generic interface, it should route to other behavior and be as flat as possible. Overloading functions like this make them impossible to test.

net/service/handlers.go
```diff
@@ -1453,3 +1462,47 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
 	}
 	return m, nil
 }
+
+func (service *OpenBazaarService) handleError(peer peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
```
placer14: Same with `options` here.

qa/purchase_offline_error.py
```diff
@@ -0,0 +1,129 @@
...
+        with open('testdata/profile.json') as profile_file:
+            profile_json = json.load(profile_file, object_pairs_hook=OrderedDict)
+        api_url = alice["gateway_url"] + "ob/profile"
+        requests.post(api_url, data=json.dumps(profile_json, indent=4))
+
+        # post listing to alice
+        with open('testdata/listing.json') as listing_file:
+            listing_json = json.load(listing_file, object_pairs_hook=OrderedDict)
+        if self.bitcoincash:
+            listing_json["metadata"]["pricingCurrency"] = "tbch"
```
placer14: Would love these o be extracted into `test_framework` to work like:
`listing_json = self.return_bitcoin_listing(bch=true)`
This could be done throughout our QA tests.

net/service/handlers.go
cpacia: At the time we had need for different types of data to be passed into there that depended in the specific type of message being handled. But we needed it to conform to the interface. 

Currently on the offline flag is used and only by some of the handlers.

api/jsonapi.go
```diff
@@ -1613,8 +1613,8 @@ func (i *jsonAPIHandler) POSTOrderCancel(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	if !(state == pb.OrderState_PENDING && len(records) > 0) || state != pb.OrderState_PENDING || contract.BuyerOrder.Payment.Method == pb.Order_Payment_MODERATED {
-		ErrorResponse(w, http.StatusBadRequest, "order must be PENDING or partially funded and only a direct payment to cancel")
+	if !((state == pb.OrderState_PENDING || state == pb.OrderState_PROCESSING_ERROR) && len(records) > 0) || !(state == pb.OrderState_PENDING || state == pb.OrderState_PROCESSING_ERROR) || contract.BuyerOrder.Payment.Method == pb.Order_Payment_MODERATED {
```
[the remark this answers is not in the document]
cpacia: I think `contract.CanCancel()` might be ideal but this would require a pretty large refactor. `contract` is a protobuf object and to add a `CanCancel()` method would require extending it... which would require changing the type everywhere `pb.RicardianContract` is used in the code. 

Might be worth it later but I'd say that would be worthy of a separate PR.

qa/purchase_offline_error.py
cpacia: Good suggestion

cpacia:
All tests pass. Going to merge.

PR #863, opened 2018-02-15
Add moderators to listing

placer14:


jjeffryes:
@placer14 this is working as expected when I use it with my verified moderators branch.

cpacia:
`RepoVersion` in repo/init.go will need to be bumped to 7.

repo/init.go
```diff
@@ -17,7 +17,7 @@ import (
 	"time"
 )
 
-const RepoVersion = "6"
+const RepoVersion = "7"
```
placer14: Does this value represent the desired migration version? Is there any reason we couldn't use the filenames of the migrations to determine this?

cpacia: This is just what the file is set to when initializing a new node. I suppose you could use the migration  index+1 for it.

placer14:
@cpacia I'm happy to have this merged when you're ready. Thanks for the assist!

repo/init.go
placer14: I'll make an issue for this to track for later fixing.

PR #867, opened 2018-02-16
Remove unexpected error cases

camelmasa:
It's tiny changes.

I thought the `verifySignature` function (`core/signatures.go`) doesn't return `noSigError` type.


cpacia:
Thanks!

PR #878, opened 2018-03-06
gomobile fixes

drwasho:
This PR contains changes for gomobile to build frameworks for iOS and Android. It is currently configured to build the framework using a Bitcoin Cash SPV wallet. 

Additional changes are required to make the selection of the wallet configurable in `\mobile\config.go`.


cmd/start.go
```diff
@@ -496,6 +496,12 @@ func (x *Start) Execute(args []string) error {
 	proto.Unmarshal(dhtrec.GetValue(), e)
 
 	// Wallet
+	// Exchange rates
```
cpacia: I think this block should not be here. Exchange rates are already set in the `start`

repo/init.go
```diff
@@ -161,7 +161,7 @@ func addConfigExtensions(repoRoot string, testnet bool) error {
 		return err
 	}
 	var w WalletConfig = WalletConfig{
-		Type:             "spvwallet",
```
cpacia: We also do not want to initialize it with bitcoin cash by default here.

cmd/start.go
drwasho: Ok fixed in the latest commit.

repo/init.go
drwasho: Fixed

drwasho:
@cpacia changes made

hoffmabc:
@drwasho can you please address @cpacia issues?

drwasho:
@hoffmabc I already did, @cpacia hasn't updated his review

placer14:
@drwasho can you help me understand what delta here that still needs to get into master? Or can we close?

drwasho:
@placer14 the current mobile folder was never configured correctly, as Chris can attest to; the only way we ever managed to get it to work was by making changes that we part of this PR.. but no one ever got around to merging them at the time. After a while it became clear that we needed to wait for multiwallet to be finished and make similar adjustments if the mobile frameworks were failing to build or function properly.

We could close this PR and prepare a fresh new one if these changes are necessary.

placer14:
Archiving.

Issue #885, opened 2018-03-19
User should be able to post a crypto listing


placer14:
Completes #876 in part...

- [x] Express unique type of `metadata.contractType` to represent crypto listings. (replace `AUCTION` type)
- [x] Express unique type of `metadata.format` to represent that the price should reflect Market Value. (referred to as listing "Market Value" format)
- [x] Express new `metadata` field which indicates the type of coin for sale in the listing. (referred to as the listing coin type)
- [x] When `metadata.contractType` reflects a new crypt listing, the server should not persist values for the following fields:
    - pricingCurrency
    - price
    - condition
    - options
    - shippingOptions
    - coupons
- [x] When `metadata.contractType` reflects a crypto listing, the server should return an error if any of the following objects are sent with a value:
    - pricingCurrency
    - condition
    - options
    - shippingOptions
    - coupons
- [x] When `metadata.contractType` reflects a crypto listing, the server should return an error if the `coinType` is empty
- [x] When `metadata.format` is set to "market value", the server should return an error if `item.price` is not zero. 
 - [x] When `metadata.contractType` reflects a crypto listing, the server should set `format` to `MARKET_PRICE`

Example listing:
```
{
   "termsAndConditions":"test",
   "refundPolicy":"test",
   "item":{
      "title":"ZEC for Sale",
      "description":"I am selling 1.5 ZEC.",
      "tags":[
         "ZEC",
         "Zcash"
      ],
      "categories":[
         "crypto"
      ],
      "nsfw":false,
      "condition":"", // should be blank
      "images":[
         {
            "filename":"cryptoImage.jpg",
            "original":"zb2rhkeGuSRJe9eKdiPAoNXRtc2APEqR8YcKwLaer75bJZrkF",
            "large":"zb2rhkfkVTD1opwgSz3i6oMKm1Uwesqzxn4SHBhb5aFsyxjcY",
            "medium":"zb2rhkJXrP7Vp5oP4EMVDEhVPweZDw8GmBvQjkcw7pKPx1uCL",
            "small":"zb2rhf5SNn6MdZL8BXeg6WcTn8tBrN6PyZS3d2YdaYrPEXo1d",
            "tiny":"zb2rhdZ7MLdpQne12gigyowCzCRBZLtgUJz4zSfnzLkSHES1s"
         }
      ],
      "options":[], // should be empty
      "skus":[
         {
            "quantity": 150000000  // this represents 1.5 ZEC 
         }
      ],
[...]

PR #887, opened 2018-03-21
FEATURE: POST/PUT ob/listings crypto listings support

ghost:
Add cryptocurrency listing support to POST/PUT listings.

#885


test/factory/listing.go
```diff
@@ -0,0 +1,109 @@
...
+	listing.Metadata.ContractType = pb.Listing_Metadata_CRYPTOCURRENCY
+	listing.Item.Skus = []*pb.Listing_Item_Sku{}
+	listing.Metadata.PricingCurrency = ""
+	listing.ShippingOptions = nil
+	listing.Item.Condition = ""
+	listing.Item.Options = nil
+	listing.Item.Price = 0
+	listing.Coupons = nil
+	return listing
+}
```
placer14: I prefer instantiating the struct with all fields defined instead of setting them field-wise. Less verbose, too.

api/jsonapi_test.go
```diff
@@ -133,6 +138,18 @@ func TestModerator(t *testing.T) {
 }
 
 func TestListings(t *testing.T) {
+	goodListingJSON := jsonFor(t, factory.NewListing("ron-swanson-tshirt"))
+	updatedListing := factory.NewListing("ron-swanson-tshirt")
```
placer14: Factories. Thank you. <3

api/jsonapi_test.go
```diff
@@ -195,6 +195,73 @@ func TestListings(t *testing.T) {
...
+	listing.Item.Options = physicalListing.Item.Options
+	runTest(listing)
+
+	listing = factory.NewCryptoListing("crypto")
+	listing.ShippingOptions = physicalListing.ShippingOptions
+	runTest(listing)
+
+	listing = factory.NewCryptoListing("crypto")
+	listing.Coupons = physicalListing.Coupons
+	runTest(listing)
```
placer14: I like this pattern.

placer14 (approved):
LGTM w one minor comment.

test/factory/listing.go
tyler-smith: I do too, but in this case I'm trying to use the base factory listing and override the appropriate fields for crypto listings instead of copying the entire huge pb.Listing definition.

placer14: Oh, I think I overlooked this being a function call. Saw all the field setting and my mind's eye saw a struct declaration on the first line instead of a method call. Sounds good!

Issue #888, opened 2018-03-21
Cryptocurrency listing should define coin divisibility


placer14:
Completes in part #876...

Currently, all coins are divisible into 100,000,000 parts for purposes of sale. Because this may change in the future and we don't want to rely on some magic number for pricing maths, this value should be included in the listing struct within `metadata` and referenced when handling the market price.

PR #889, opened 2018-03-21
FEATURE: Add coin divisibility to crypto listings

ghost:
#888

Depends on PR #887


core/listings.go
```diff
@@ -43,6 +43,8 @@ const (
 	MaxCountryCodes          = 255
 	EscrowTimeout            = 1080
 	SlugBuffer               = 5
+
+	DefaultCoinDivisibility uint32 = 10e7
```
placer14: Suppose to be 100_000_000 instead of 10_000_000.

placer14: Suppose to be 100_000_000 instead of 10_000_000

placer14 (changes requested):
One fix. GTM!

core/listings.go
tyler-smith: It is 100,000,00 but I can change it to 1e8 instead of 10e7 if that's more readable.

tyler-smith:
@placer14 It wasn't incorrect, but I've changed it from 10e7 to 1e8 to be more clear.

placer14 (approved):

PR #890, opened 2018-03-21
TWEAK: Add GOPATH/bin to 'make protos'.

ghost:
Add GOPATH/bin to 'make protos'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ P_ANY = Mgoogle/protobuf/any.proto=github.com/golang/protobuf/ptypes/any
 PKGMAP = $(P_TIMESTAMP),$(P_ANY)
 
 protos:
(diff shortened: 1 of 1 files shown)


tyler-smith:
This fixes the case that protoc-gen-go is installed with `go get ...`, which puts the binary in $GOPATH/bin.

placer14:
A brief discussion on including `GOPATH/bin` in `PATH` resulted in agreeing that mikefiles should not care about the path. I guess the only question is if GOPATH/bin is an expected default. A quick research suggests it's not standard across OSes where it might be considered a security concern to allow `go install` to potentially replace other global executes (depending on GOPATH/bin's precedence in the PATH list).

For that point alone, I'll leave it as is.

placer14 (approved):
LGTM

Issue #892, opened 2018-03-22
GET ob/order should show cryptocurrency purchase fields


tyler-smith:
Completes #876 in part...

### GET ob/order
- [x] When getting the order for a crypto currency listing, the new crypto currency specific data should be returned with the other data. 
    - coinType
    - paymentAddress

tyler-smith:
merged in cryptolistings branch

Issue #886, opened 2018-03-21
User should be able to create a crypto listing order


tyler-smith:
Completes #876 in part...

### POST ob/purchase

- [x] When the purchase is for a crypto currency listing, the server should return an error if the crypto currency item in the item array does not have a value for `paymentAddress `.
- [x] When the purchase is for a crypto currency listing, the server should return an error if the crypto currency item in the item array has values in the options or coupons arrays.
- [x] When a crypto currency purchase is made, the amount returned by the server should be calculated based on the current exchange rates between the listing's coinType and the buyer's currency.

Example order:
```
{
   "shipTo":"",
   "address":" ",
   "city":"",
   "state":"",
   "postalCode":"",
   "countryCode":"UNITED_STATES",
   "addressNotes":"",
   "moderator":"QmX1GdRaaBCfcASyqEhisyY98SGdBBn3JLEUPzExaPJpJh",
   "items":[
      {
         "listingHash":"zb2rhmc3pUT6Nx2rHTLTaWrLJCRAx4zfkhYXdDoG1cTHb5CrJ",
         "quantity":300000000,
         "options":[], // should be empty
         "shipping":{
            "name":"",
         },
         "memo":"",
         "coupons":[], // should be empty
         "paymentAddress ": "qr5dfheu5udjdn2ez8t40ac46a9lnw7ypvdq06qj3c"
      }
   ],
   "memo":"",
   "alternateContactInfo":""
}
```

PR #895, opened 2018-03-22
FEATURE: Add paymentAddress to purchase data items for crypto listings

ghost:
Completes #886 and #892 

Depends on PR #889


core/order.go
```diff
@@ -751,14 +804,23 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
 	// Calculate the price of each item
 	for _, item := range contract.BuyerOrder.Items {
 		var itemTotal uint64
+		var itemQuantity uint32 = item.Quantity
```
placer14: `var` block?

core/order.go
```diff
@@ -751,14 +804,23 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
...
 		l, err := ParseContractForListing(item.ListingHash, contract)
 		if err != nil {
 			return 0, fmt.Errorf("Listing not found in contract for item %s", item.ListingHash)
 		}
 		if l.Metadata.ContractType == pb.Listing_Metadata_PHYSICAL_GOOD {
 			physicalGoods[item.ListingHash] = l
 		}
-		satoshis, err := n.getPriceInSatoshi(l.Metadata.PricingCurrency, l.Item.Price)
+
+		var satoshis uint64
```
placer14: I'd be okay with this in the var block above as well.

core/order.go
```diff
@@ -741,6 +781,19 @@ func (n *OpenBazaarNode) CalcOrderId(order *pb.Order) (string, error) {
 	return id.B58String(), nil
 }
 
+func (n *OpenBazaarNode) getMarketPriceInSatoshis(symbol string, quantity uint32) (uint64, error) {
```
placer14: What kind of symbol is this? (It's a currencySymbol. Or a currencyCode.)

core/order.go
```diff
@@ -741,6 +781,19 @@ func (n *OpenBazaarNode) CalcOrderId(order *pb.Order) (string, error) {
...
+	if n.ExchangeRates == nil {
+		return 0, ErrMarketPriceRequiresExchangeRates
+	}
```
placer14: If this can be nil, the other call `getPriceInSatoshi` could also crash if ExchangeRates are off. Can we improve the other method to include this check?

core/order.go
```diff
@@ -821,11 +883,14 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
...
-		itemTotal *= uint64(item.Quantity)
+		itemTotal *= uint64(itemQuantity)
 		total += itemTotal
 	}
 
-	// Add in shipping costs
+	// Add in shipping costs for physical items
+	if len(physicalGoods) == 0 {
+		return total, nil
+	}
```
placer14: Instead of short circuiting the whole function, could be just protect the shipping calculations inside an IF block? I don't want to have to be aware that non-physical goods won't get processed behind this point.

placer14: Or even better, extract this behavior into a function which returns the appropriate shipping value to be attached to the order?

qa/purchase_crypto_listing.py
```diff
@@ -0,0 +1,153 @@
...
+
+class PurchaseCryptoListingTest(OpenBazaarTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 2
+
+    def run_test(self):
+        alice = self.nodes[0]
+        bob = self.nodes[1]
```
placer14: Can we adopt more meaningful names in our QA tests which describe the person's role in the transaction? I don't want to have to remember if alice is the buyer or the vendor, let's just call them what they are.

qa/testdata/crypto_listing.json
```diff
@@ -0,0 +1,32 @@
+{
+    "slug": "",
+    "metadata": {
+        "version": 1,
+        "contractType": "CRYPTOCURRENCY",
+        "format": "MARKET_PRICE",
+        "expiry": "2030-08-17T04:52:19.000Z",
+        "coinType": "ETH"
```
placer14: Should we include coin divisibility here for completeness sake?

placer14 (commented):
Some general thoughts, but no deal breakers.

tyler-smith:
Moved to #900

PR #899, opened 2018-03-25
Add wallet support for Zcoin

t3rm1n4l:
Zcoin wallet can be configured by setting the wallet config as follows:
  "Wallet": {
    "Binary": "/Applications/Zcoin-Qt.app/Contents/MacOS/Zcoin-Qt",
    "FeeAPI": "",
    "HighFeeDefault": 160,
    "LowFeeDefault": 20,
    "MaxFee": 2000,
    "MediumFeeDefault": 60,
    "TrustedPeer": "",
    "Type": "zcoind"
  }

### DIFF
--- a/bitcoin/zcoind/address.go
+++ b/bitcoin/zcoind/address.go
@@ -1,0 +1,334 @@
+package zcoind
+
+import (
+	"errors"
+
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/base58"
+	"golang.org/x/crypto/ripemd160"
+)
+
+var (
+	// ErrChecksumMismatch describes an error where decoding failed due
+	// to a bad checksum.
+	ErrChecksumMismatch = errors.New("checksum mismatch")
+
+	// ErrUnknownAddressType describes an error where an address can not
+	// decoded as a specific address type due to the string encoding
+	// begining with an identifier byte unknown to any standard or
+	// registered (via chaincfg.Register) network.
+	ErrUnknownAddressType = errors.New("unknown address type")
+
+	// ErrAddressCollision describes an error where an address can not
+	// be uniquely determined as either a pay-to-pubkey-hash or
+	// pay-to-script-hash address since the leading identifier is used for
+	// describing both address kinds, but for different networks.  Rather
+	// than assuming or defaulting to one or the other, this error is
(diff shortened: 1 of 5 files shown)


t3rm1n4l:
Thanks for the quick feedback.
I looked at the qa tests and found that the current test framework tightly coupled with bitcoind. Looks like a regtest network is a requirement for the tests. Unfortunately, Zcoin does not have a regtest network. I was looking for an altcoin test for Zcash so that I could follow. At this point, I am not quite sure how to automate the tests for Zcoin.

placer14:
@t3rm1n4l For lack of an integration test pattern, I would recommend unit tests around your behavior which proves they function as intended. A simple example might be to write a unit test around `CheckEncode` which offers some example inputs and verifies their expected outputs. The inputs should be significant enough to exercise as many branches in your function as is reasonable to prove newly added behavior

Upon closer inspection, it appears that the implementation requires the external Zerocash binary. Given our experience with this approach on ZCash we're hesitant to merge any non-SPV implementation. Does Zerocoin allow an SPV wallet implementation? If so, would you be willing to reimplement this as an SPV wallet?

t3rm1n4l:
I will try to rework the patch to add integration into https://github.com/OpenBazaar/spvwallet. Hope, that is the expectation.

cpacia:
@t3rm1n4l I think what we're leaning towards is to have only a few coins natively integrated to maximize maintainability. 

However, we are going to work on a plugin architecture to make it easy to integrate new coins. It should work similar to this https://medium.com/learning-the-go-programming-language/writing-modular-go-programs-with-plugins-ec46381ee1a9

Basically your code would remain exactly the same. You would just compile a binary plugin file and drop it in a plugins folder and on next startup the coin would be available. 

Soon we're going to be starting a large refactor to make the app work with multiple coins at the same time. At this time we will probably build in the plugin system. 

Hope to start this in the next couple weeks and should be finished within a month or so.

t3rm1n4l:
Thanks. I will probably wait for the plugin refactoring and i will adapt the changes accordingly.

cpacia:
Cool. I don't think your code will need to change much. It will just need to be put in a separate package and compiled with a plugin flag.

plasma-crypto:
@cpacia Any estimated timeline on the coin integration interface refactor ?

plasma-crypto:
Are you guys planning to accept coin integrations ? Is there any update on the new integration model ?

placer14:
3rd party integrations will be possible with a plugin interface that we are still working on. Developer updates are published to our youtube channel every two-ish weeks. Update from June 21 had some wallet-related discussion.

cpacia:
We can revisit once plugins are available

Issue #893, opened 2018-03-22
POST ob/orderfulfillment crypto listings support


tyler-smith:
Completes #876 in part...

### POST ob/orderfulfillment
- [x] When the order being fulfilled is for a crypto currency listing, the API should require a transactionId value as a string, which represents the ID of the transaction where the seller transferred the crypto currency to the buyer.
- [x] When the order being fulfilled is for a crypto currency listing, if a trandactionId is not supplied, the server should return an error.
- [x] Check to ensure that the wrong fulfillment type is not provided for fulfillment.

The transactionId could be represented as an object in the protobuff:
```
message CryptoCurrencyDelivery: {
  string transactionId
}
```

tyler-smith:
merged into cryptolistings branch.

PR #900, opened 2018-03-26
Ts cryptolistings

ghost:
Continues from #895 

Completes #893


core/order.go
```diff
@@ -874,25 +877,35 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
 		total += itemTotal
 	}
 
-	// Add in shipping costs for physical items
-	if len(physicalGoods) == 0 {
-		return total, nil
+	shippingTotal, err := n.calculateShippingTotalForListings(contract, physicalGoods)
```
placer14: ty!

qa/purchase_crypto_listing.py
```diff
@@ -0,0 +1,176 @@
...
+        if resp["state"] != "FULFILLED":
+            raise TestFailure("PurchaseCryptoListingTest - FAIL: Buyer failed to detect order fulfillment")
+        if resp["contract"]["vendorOrderFulfillment"][0]["cryptocurrencyDelivery"][0]["transactionID"] != "crypto_transaction_id":
+            raise TestFailure("PurchaseCryptoListingTest - FAIL: Buyer failed to detect transactionID")
+
+        print("PurchaseCryptoListingTest - PASS")
+
+if __name__ == '__main__':
+    print("Running PurchaseCryptoListingTest")
+    PurchaseCryptoListingTest().main(["--regtest"])
```
placer14: I think all tests should be using this option? I'm surprised to find this here. Is there a better location which includes this option on every test instead of just this one?

placer14 (approved):
A few very minor thoughts, but GTM otherwise.

qa/purchase_crypto_listing.py
tyler-smith: By default flags are `["--disablewallet", "--testnet", "--disableexchangerates"]`. We could add `--regtest` but it would still need to be overridden in all the places that don't want --disablewallet. We need a better way to declaring which flags a test should use.

Maybe have the same defaults but be able to do like `PurchaseCryptoListingTest().WithWallet().WithExchangeRates()`

PR #909, opened 2018-03-30
Add conditions for new coins integrations

SamPatt:


cpacia:
I think this should probably also mention the plugin approach.

PR #926, opened 2018-04-20
Dispute Notifications

placer14:
Concerns #843


cpacia:
Dammit I was trying to go backwards on the page and it deleted my comments :/

placer14:
@cpacia This is because the worker intends to make the insert atomic for each type of notification and if the worker bombed in the middle of a job it would not leave the data in an undefined state. The transactions were not yet added to the Record abstraction and this was my compromise. There is some debt here to add a `(Record).PutRecordWithTransaction(record, tx)` or similar.

PR #937, opened 2018-05-04
TWEAK: Support larger order quantities by changing uint32 -> uint64.

tyler-smith:
@rmisio and @jjeffryes found a bug with ordering 5000 ARK tokens. The issue is the quantity (5000*100000000) is larger than the uint32 max so it wraps around resulting in an incorrect quantity. This change changes the quantity to use uint64 which increases the safe limit from 4294967295 to 18446744073709551615.

Long term we should probably replace all json numbers with strings and parse into uint64s so we can detect wrap around.

Note: Tests updated to use cheaper coin so we can test higher quantity without generating more blocks, which takes more time.


jjeffryes:
> Long term we should probably replace all json numbers with strings and parse into uint64s so we can detect wrap around.

Can we create a tracking issue for that? It sounds like something we should do.

rmisio:
> Long term we should probably replace all json numbers with strings and parse into uint64s so we can detect wrap around.

I strongly disagree with this approach. I've said this many times before, but I think it will introduce a lot of scope and refactoring for clients to have to support representing numbers as strings and support numbers beyond the top range of what js natively supports. You would have to bring in some type of library to manage that and run all numeric arithmetic operation through that lib (addition, subtraction, multiplication, etc...).

Just for some perspective. An uint64 has a max of 18446744073709551615, which in the context of quantity means we could support a quantity of 18446744073709551615 / 100,000,000 =  184,467,440,737. That sounds to me like a high enough limit.

PR #943, opened 2018-05-09
BUGFIX: Fix ipfs keystore caching.

tyler-smith:
1) It wasn't writing to cache after a cache miss
2) It wasn't reloading cache asynchronously after cache hit which causes a deadlock when issue #1 is fixed

### DIFF
--- a/repo/ipfs_keystore.go
+++ b/repo/ipfs_keystore.go
@@ -58,12 +58,7 @@ func GetObjectFromIPFS(ctx commands.Context, p peer.ID, name string, maxCacheLen
 	getObjectFromIPFSCacheMu.Lock()
 	defer getObjectFromIPFSCacheMu.Unlock()
 
-	entry, ok := getObjectFromIPFSCache[getIPFSCacheKey(p, name)]
-	if !ok {
-		return fetchObjectFromIPFS(ctx, p, name)
-	}
-
-	if entry.created.Add(maxCacheLen).Before(time.Now()) {
+	fetchAndUpdateCache := func() ([]byte, error) {
 		objBytes, err := fetchObjectFromIPFS(ctx, p, name)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


repo/ipfs_keystore.go
```diff
@@ -77,19 +72,16 @@ func GetObjectFromIPFS(ctx commands.Context, p peer.ID, name string, maxCacheLen
...
 	}
 
+	entry, ok := getObjectFromIPFSCache[getIPFSCacheKey(p, name)]
+	if !ok || entry.created.Add(maxCacheLen).Before(time.Now()) {
+		return fetchAndUpdateCache()
+	}
+
 	// Update cache in background after a successful read
-	func() {
+	go func() {
```
placer14: Nice

repo/ipfs_keystore.go
```diff
@@ -58,12 +58,7 @@ func GetObjectFromIPFS(ctx commands.Context, p peer.ID, name string, maxCacheLen
 	getObjectFromIPFSCacheMu.Lock()
 	defer getObjectFromIPFSCacheMu.Unlock()
```
placer14: This defer will release the lock before the goroutine below is necessarily done. You might get unreliable data.

placer14: Not sure if this is a good time to use this or how much of a change it is: https://golang.org/pkg/sync/#Map. It's a map with the semaphore built in with O(n) lookup/storage.

placer14: The function will continue and mutate the map after the lock has closed. https://play.golang.org/p/rHz6-LLFL4h

cpacia: But that's the behavior we want here isn't it?

placer14: The lock is named `getObjectFromIPFSCacheMu`. It would seem it's trying to protect the map from being mutated. So I would guess not. But @tyler-smith helped me notice we have the lock inside the go routine also, which I overlooked. The lock inside the go routine should be enough.

PR #966, opened 2018-05-31
Ensure calls to POST /ob/releaseescrow w ZEC sales fail

placer14:
Completes `return error on POSTReleaseEscrow when Node uses ZEC` in #843


placer14:
@cpacia Agree. I went this route given the wallet work you had in flight.

PR #967, opened 2018-06-04
Update spv wallet dependencies

cpacia:
The Bitcoin and Bitcoin Cash spvwallets have been updated with all new wire and blockchain handling code. 

Several of the btcd wallet libraries have been updated.

The bitcoind and zcashd packages have been updated to conform to the new btcd `InputSource` interface.

A small change has been made to the db wallet interface and the corresponding implementation in this package. (this change allows the use of block timestamps for transaction time after the tx confirms. Rather than using time.Now() which gives the wrong impression that the transaction is recent after a resync). 

The bchutil library has been updated to build cashaddrs without a prefix.

TODO:

- [x] Fix txns db test

- [x] Implement temporary fix to prevent purchase failure when an non-upgraded node purchases from an upgraded node (and vice versa) using the new bitcoin cash address format.


vendor/github.com/OpenBazaar/spvwallet/blockchain.go
```diff
@@ -299,7 +293,7 @@ func (b *Blockchain) GetBlockLocatorHashes() []*chainhash.Hash {
 		}
 		start += 1
 	}
-	return ret
+	return blockchain.BlockLocator(ret)
```
placer14: Where does this get used? Looks like it's exposing a large abstraction outside of the wallet and I'm curious what future dependencies this creates on the spvwallet interface.

vendor/github.com/OpenBazaar/spvwallet/peers.go
```diff
@@ -236,124 +209,75 @@ func (pm *PeerManager) onVerack(p *peer.Peer, msg *wire.MsgVerAck) {
...
-	}
-}
-
-func (pm *PeerManager) selectNewDownloadPeer() {
-	for peer := range pm.readyPeers {
-		pm.setDownloadPeer(peer)
-		break
+	log.Debugf("Peer %s disconnected", peer)
+	delete(pm.connectedPeers, req.ID())
+	if pm.msgChan != nil {
```
placer14: How can we get the `msgChan` nil check removed from every message? Any way we ensure that the `PeerManager` never has a nil `msgChan`? If not, then this should be in a private method that handles this for us quietly. Also, what does it mean for the msgChan to be nil and then drop the command it intended to send entirely...is this case handled?

vendor/github.com/OpenBazaar/spvwallet/peers.go
```diff
@@ -236,124 +209,75 @@ func (pm *PeerManager) onVerack(p *peer.Peer, msg *wire.MsgVerAck) {
...
-			return &net.TCPAddr{}, errors.New("Adder manager returned nil address")
-		}
-		knownAddress := ka.NetAddress()
-		addr := &net.TCPAddr{
-			Port: int(knownAddress.Port),
-			IP:   knownAddress.IP,
+			// Don't return addresses we're still connected to
+			for _, p := range pm.connectedPeers {
+				if p.NA().IP.String() == knownAddress.IP.String() {
+					continue loop
```
placer14: Why a labeled continue here and not extract this whole thing as a function call that returns an error you can check?

I tend to avoid labels as much as possible. An extracted function is more descriptive (added function name, clear input/return values), less opportunity for mutation errors should someone add behavior here later (like getting surprised with a continue that maintains state of an added variable versus as return which resets the value of the variable to a known state, zero, upon return), and much easier to reason about in your head instead of having to search around a largish function for the entry/exit points.

vendor/github.com/OpenBazaar/spvwallet/txstore.go
```diff
@@ -354,14 +354,15 @@ func (ts *TxStore) Ingest(tx *wire.MsgTx, height int32) (uint32, error) {
 		}
 		// Let's check the height before committing so we don't allow rogue peers to send us a lose
 		// tx that resets our height to zero.
-		if txn.Height <= 0 {
-			ts.Txns().UpdateHeight(tx.TxHash(), int(height))
+		if err == nil && txn.Height <= 0 {
```
placer14: I would've added an `else` to the prior block and nest the `Height` check inside so it's obvious the logic can only go one way or the other to the next reader instead of being surprised this separate block of logic is conditioned on the same data as the prior block.

vendor/github.com/btcsuite/btcd/blockchain/blockindex.go
```diff
@@ -101,33 +101,33 @@ type blockNode struct {
...
 		nonce:      blockHeader.Nonce,
 		timestamp:  blockHeader.Timestamp.Unix(),
 		merkleRoot: blockHeader.MerkleRoot,
 	}
+	if parent != nil {
+		node.parent = parent
+		node.height = parent.height + 1
+		node.workSum = node.workSum.Add(parent.workSum, node.workSum)
+	}
 }
```
placer14: Instead of a mutating initialization function, make a constructor which returns a valid `*blockNode`. Then you never have a node in a bad state or have to worry about concurrent access. :)

vendor/github.com/btcsuite/btcd/blockchain/blockindex.go
```diff
@@ -101,33 +101,33 @@ type blockNode struct {
...
-func newBlockNode(blockHeader *wire.BlockHeader, height int32) *blockNode {
+// newBlockNode returns a new block node for the given block header and parent
+// node, calculating the height and workSum from the respective fields on the
+// parent. This function is NOT safe for concurrent access.
+func newBlockNode(blockHeader *wire.BlockHeader, parent *blockNode) *blockNode {
 	var node blockNode
-	initBlockNode(&node, blockHeader, height)
+	initBlockNode(&node, blockHeader, parent)
 	return &node
 }
```
placer14: It seems like you did. I'm not sure why this function is unsafe for concurrent access...

vendor/github.com/btcsuite/btcd/blockchain/blockindex.go
```diff
@@ -241,6 +242,7 @@ func newBlockIndex(db database.DB, chainParams *chaincfg.Params) *blockIndex {
 		db:          db,
 		chainParams: chainParams,
 		index:       make(map[chainhash.Hash]*blockNode),
+		dirty:       make(map[*blockNode]struct{}),
```
placer14: This would make all of your function that touch `dirty` concurrently safe: https://godoc.org/golang.org/x/sync/syncmap

PR #968, opened 2018-06-07
Remove the params call

amangale:
Remove the redundant Params func call because params struct is already embedded inside the wallet implementation and is used everywhere else.


cpacia:
There's a number of other places where it's used besides these two. I think it's best to get a solid spec for the interface changes agreed on and then make a PR for all the interface changes.

PR #970, opened 2018-06-11
[#843] Produce purchase dispute expiry notifications

placer14:
Completes `generate buyerDisputeExpiry for buyer when dispute is opened after 15, 40, 44 days` in #843


cpacia:
Didn't see anything that stood out to me. Should be good.

PR #979, opened 2018-06-19
Remove dummy package

placer14:
The package has no obvious use and will fail `go test ./...` for the last time. 🍻 

Note that this also removes a `Dockerfile.dummy` with no obvious consumer. I'd like to clarify before merge what purpose that package had and that this missing docker image won't have any negative impact.


tyler-smith:
We used it create test nodes at one point but it's fallen behind and isn't used anymore. Purge it.

PR #1002, opened 2018-07-05
Add multiwallet configuration to config file

cpacia:
This commit sets up the config file for use in the multiwallet refactor.
Each coin has its own config added with the idea that, abset the config, the
defaults will be used.

This commit does not change the existing wallet config object which will be
deprecated when the full refactor is complete.

### DIFF
--- a/repo/init.go
+++ b/repo/init.go
@@ -168,6 +168,45 @@ func addConfigExtensions(repoRoot string, testnet bool) error {
 			LowFeeDefault:    20,
 			TrustedPeer:      "",
 		}
+		ws = schema.WalletsConfig{
+			BTC: schema.CoinConfig{
+				Type:             "API",
+				API:              "https://btc.bloqapi.net/insight-api",
+				APITestnet:       "https://test-insight.bitpay.com/api",
+				MaxFee:           200,
+				FeeAPI:           "https://btc.fees.openbazaar.org",
+				HighFeeDefault:   50,
+				MediumFeeDefault: 10,
+				LowFeeDefault:    1,
+			},
+			BCH: schema.CoinConfig{
+				Type:             "API",
+				API:              "https://bch-insight.bitpay.com/api",
+				APITestnet:       "https://test-bch-insight.bitpay.com/api",
+				MaxFee:           200,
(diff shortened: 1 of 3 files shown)


repo/init.go
```diff
@@ -168,6 +168,45 @@ func addConfigExtensions(repoRoot string, testnet bool) error {
 			LowFeeDefault:    20,
 			TrustedPeer:      "",
 		}
+		ws = schema.WalletsConfig{
+			BTC: schema.CoinConfig{
+				Type:             "API",
+				API:              "https://btc.bloqapi.net/insight-api",
+				APITestnet:       "https://test-insight.bitpay.com/api",
```
placer14: `API`, `APITestnet`, `FeeAPI` are all specific to the wallet in use and is not specific to the coin. Imagine where one API wallet might use one API and another wallet uses another API. I'd prefer that `[]WalletsConfig` contains `WalletConfig` and `[]CoinConfigs` contain `CoinConfig`. This way each wallet can be configured separately as needed.

repo/init.go
```diff
@@ -168,6 +168,45 @@ func addConfigExtensions(repoRoot string, testnet bool) error {
 			LowFeeDefault:    20,
 			TrustedPeer:      "",
 		}
```
placer14: I assume the `WalletConfig` above here will be deprecated? What's your plan with this?

cpacia: Yeah when the actual multiwallet is merged in that one can be removed.

repo/init.go
cpacia: What would go into `CoinConfig` in your mind? The data here is really specific to the implementation as some other implementation could require different settings.

placer14: Things which are universal to the coin: `economyFee`, `*Fee`, the Divisibility for crypto listings if we need to coin-specific values, perhaps?

cpacia: Things like divisibility which are unique to the coin should probably be part of the coin interface and hardcoded rather than a configurable option. 

I think the config is only for user settings how they want their wallet to run. 

On those lines, fees are not really coin-specific per se, users might want to modify the defaults (and I know a couple users who have).

PR #1009, opened 2018-07-09
Migrate wallet database tables

cpacia:
Migrates the db tables used by the wallet to have a `coin` column so that multiple coin implementations can share the same database.

Currently db structs are created as part of the `SQLiteDatastore` object but once the multiwallet is in we'll move the creation of the structs into the wallet constructor. Each wallet will create it's own db struct with the tables it needs. 

But for now this means the `cointype` needs to be determined on startup and passed into the db package so that the tables use the correct cointype. But this wont be necessary after the multiwallet refactor as each wallet knows its own cointype and will initialize the struct itself.


repo/migrations/Migration009_test.go
```diff
@@ -0,0 +1,377 @@
...
+		if watchOnly != 0 {
+			t.Error("Failed to return correct watchOnly")
+		}
+		if spendHeight != 150001 {
+			t.Error("Failed to return correct spendHeight")
+		}
+		if spendTxid != "1234" {
+			t.Error("Failed to return correct spendTxid")
+		}
+	}
```
placer14: Testing pattern to adopt: Memoize the asserted values to indicate the actual value doesn't matter and only needs to match the beginning, improves maintainability, and the variable names can be used to indicate what type of value it is suppose to be.

mobile/node.go
```diff
@@ -108,6 +101,23 @@ func NewNode(config NodeConfig) (*Node, error) {
...
+	}
+	migrations.WalletCoinType = ct
+
+	sqliteDB, err := initializeRepo(config.RepoPath, "", "", true, time.Now(), ct)
+	if err != nil && err != repo.ErrRepoExists {
+		return nil, err
+	}
+
+	// Get creation date. Ignore the error and use a default timestamp.
+	creationDate, _ := sqliteDB.Config().GetCreationDate()
```
placer14: Don't ignore the error, please.

cmd/start.go
```diff
@@ -279,6 +254,45 @@ func (x *Start) Execute(args []string) error {
...
+			return err
+		}
+		if sqliteDB.Config().IsEncrypted() {
+			log.Error("Invalid password")
+			os.Exit(3)
+		}
+	}
+
+	// Get creation date. Ignore the error and use a default timestamp.
+	creationDate, _ := sqliteDB.Config().GetCreationDate()
```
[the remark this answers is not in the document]
cpacia: ¯\_(ツ)_/¯ 

If it errors (or isn't set) then creationDate will just be the unix epoch. Which is fine as the wallet just uses it to decide if it can skip downloading some blocks to save time. Worst case is the sync just takes longer. Not sure if it's worth panicing here rather than just proceeding with a longer sync.

vendor/github.com/OpenBazaar/wallet-interface/datastore.go
```diff
@@ -32,6 +37,21 @@ func (c *CoinType) String() string {
...
+	case BitcoinCash:
+		return "BCH"
+	case Zcash:
+		return "ZEC"
+	case Litecoin:
+		return "LTC"
+	default:
+		return ""
+	}
+}
```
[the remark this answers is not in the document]
cpacia: That's a  good point. I think the answer is no as the `coinType` only specifies the specific coin whereas there are other things ... like the `Params()` and `CurrencyCode()` interface methods that can show if it's testnet or not.

mobile/node.go
cpacia: See above

cmd/start.go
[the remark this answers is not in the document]
placer14: I guess the problem is that it shouldn't return an error. Do we ever care about that value?

vendor/github.com/OpenBazaar/wallet-interface/datastore.go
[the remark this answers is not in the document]
placer14: I would really like to make sure we're consistent with our usage and representations everywhere. I know there are places where we DO express the cointype in terms of "TBTC" or similar. It may not be of consequence to any of our logic immediately, but it would good to represent this accurately if possible. This choice also requires us to use SQL AND Marshalling to inspect the serializedContract for the real type (testing vs mainnet) instead of just being able to `where coin = "TBTC"` on a query.

cmd/convert.go
```diff
@@ -173,7 +177,7 @@ func (x *Convert) Execute(args []string) error {
 	// Update listings with new coin type
 	creationDate := time.Now()
 	var sqliteDB *db.SQLiteDatastore
-	sqliteDB, err = InitializeRepo(repoPath, x.Password, "", x.Testnet, creationDate)
+	sqliteDB, err = InitializeRepo(repoPath, x.Password, "", x.Testnet, creationDate, ct)
```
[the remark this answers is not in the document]
placer14: Thinking about this further, this will have to be removed once we have multiple currencies active at once. How are you planning on switching currencies on a per-transaction basis with this design?

vendor/github.com/OpenBazaar/wallet-interface/datastore.go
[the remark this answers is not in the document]
cpacia: The only tables that are getting a column added are those used by the wallets. So purchases, sales, etc (which would have a serializedContract) are unaffected.

cmd/start.go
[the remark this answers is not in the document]
cpacia: I could see a scenario where I'd probably want to catch an error rather than return a default. But the start function isn't one of them.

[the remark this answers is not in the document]
tyler-smith: I would try to log that there was an error and that block sync will start from the beginning.

repo/migrations/Migration011.go
```diff
@@ -0,0 +1,190 @@
...
+		databaseFilePath string
+	)
+	if testnetEnabled {
+		databaseFilePath = path.Join(repoPath, "datastore", "testnet.db")
+	} else {
+		databaseFilePath = path.Join(repoPath, "datastore", "mainnet.db")
+	}
+
+	const (
+		alterKeysSQL                 = "alter table keys add coin text;"
```
placer14: Could we use `coinType` to match the column names we're using in `purchases` and `sales`?

repo/migrations/Migration011.go
```diff
@@ -0,0 +1,190 @@
...
+
+import (
+	"database/sql"
+	"fmt"
+	"github.com/OpenBazaar/wallet-interface"
+	"path"
+	"strings"
+)
+
+var WalletCoinType wallet.CoinType
```
placer14: We have been prefixing public vars in `migrations` with `MigrationXXX_ConstantName` pattern so we can avoid colliding vars inside this package. Ex: https://github.com/OpenBazaar/openbazaar-go/blob/master/repo/migrations/Migration009.go#L17

repo/migrations/Migration011.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+	_, err = db.Exec("update txns set coin = ?", WalletCoinType.CurrencyCode())
+	if err != nil {
+		return err
+	}
+	_, err = db.Exec("update watchedscripts set coin = ?", WalletCoinType.CurrencyCode())
+	if err != nil {
+		return err
+	}
```
placer14: Please put all of these into a single transaction so it reverts if it fails.

repo/migrations/Migration011.go
```diff
@@ -0,0 +1,190 @@
...
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+	if _, err = tx.Exec(migration); err != nil {
+		tx.Rollback()
+		return err
+	}
+	if err = tx.Commit(); err != nil {
+		return err
```
placer14: You should Rollback on Commit error, too.

repo/migrations/Migration011.go
```diff
@@ -0,0 +1,190 @@
...
+		alterKeysSQL                 = "alter table keys add coin text;"
+		createKeysIndexSQL           = "create index index_keys on keys (coin);"
+		alterUTXOSQL                 = "alter table utxos add coin text;"
+		createUTXOIndexSQL           = "create index index_utxos on utxos (coin);"
+		alterSTXOSQL                 = "alter table stxos add coin text;"
+		createSTXOIndexSQL           = "create index index_stxos on stxos (coin);"
+		alterTXNSSQL                 = "alter table txns add coin text;"
+		createTXNSIndexSQL           = "create index index_txns on txns (coin);"
+		alterWatchedScriptsSQL       = "alter table watchedscripts add coin text;"
+		createWatchedScriptsIndexSQL = "create index index_watchedscripts on watchedscripts (coin);"
```
placer14: Make the columns `NOT NULL DEFAULT ''`

repo/migrations/Migration011_test.go
```diff
@@ -0,0 +1,382 @@
...
+		databasePath = appSchema.DatabasePath()
+		schemaPath   = appSchema.DataPathJoin("repover")
+
+		schemaSql = "pragma key = 'foobarbaz';"
+
+		CreateTableKeysSQL                      = "create table keys (scriptAddress text primary key not null, purpose integer, keyIndex integer, used integer, key text);"
+		CreateTableUnspentTransactionOutputsSQL = "create table utxos (outpoint text primary key not null, value integer, height integer, scriptPubKey text, watchOnly integer);"
+		CreateTableSpentTransactionOutputsSQL   = "create table stxos (outpoint text primary key not null, value integer, height integer, scriptPubKey text, watchOnly integer, spendHeight integer, spendTxid text);"
+		CreateTableTransactionsSQL              = "create table txns (txid text primary key not null, value integer, height integer, timestamp integer, watchOnly integer, tx blob);"
+		CreatedTableWatchedScriptsSQL           = "create table watchedscripts (scriptPubKey text primary key not null);"
```
placer14: Typo: `Created`TableWatchedScriptsSQL

schema/constants.go
```diff
@@ -10,18 +10,23 @@ const (
...
+	CreateTableKeysSQL                      = "create table keys (coin text, scriptAddress text primary key not null, purpose integer, keyIndex integer, used integer, key text);"
+	CreateIndexKeysSQL                      = "create index index_keys on keys (coin);"
+	CreateTableUnspentTransactionOutputsSQL = "create table utxos (coin text, outpoint text primary key not null, value integer, height integer, scriptPubKey text, watchOnly integer);"
+	CreateIndexUnspentTransactionOutputsSQL = "create index index_utxos on utxos (coin);"
+	CreateTableSpentTransactionOutputsSQL   = "create table stxos (coin text, outpoint text primary key not null, value integer, height integer, scriptPubKey text, watchOnly integer, spendHeight integer, spendTxid text);"
+	CreateIndexSpentTransactionOutputsSQL   = "create index index_stxos on stxos (coin);"
+	CreateTableTransactionsSQL              = "create table txns (coin text, txid text primary key not null, value integer, height integer, timestamp integer, watchOnly integer, tx blob);"
+	CreateIndexTransactionsSQL              = "create index index_txns on txns (coin);"
```
placer14: I'm not sure if this is going to create any problems later on, but the column order in this schema is different from the column order that the migrations will produce. Which means that new nodes will have `coin` in the first column but migrated nodes will have `coin` in the last column. Could create a problem later on that would be very hard to detect and resolve.

schema/constants.go
```diff
@@ -10,18 +10,23 @@ const (
...
 	CreateIndexPurchasesSQL                 = "create index index_purchases on purchases (paymentAddr, timestamp);"
 	CreateTableSalesSQL                     = "create table sales (orderID text primary key not null, contract blob, state integer, read integer, timestamp integer, total integer, thumbnail text, buyerID text, buyerHandle text, title text, shippingName text, shippingAddress text, paymentAddr text, funded integer, transactions blob, needsSync integer, lastDisputeTimeoutNotifiedAt integer not null default 0, coinType not null default '', paymentCoin not null default '');"
 	CreateIndexSalesSQL                     = "create index index_sales on sales (paymentAddr, timestamp);"
-	CreatedTableWatchedScriptsSQL           = "create table watchedscripts (scriptPubKey text primary key not null);"
+	CreatedTableWatchedScriptsSQL           = "create table watchedscripts (coin text, scriptPubKey text primary key not null);"
```
placer14: Same here.

cmd/convert.go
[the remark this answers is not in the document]
placer14: I'm placated with fixing this later when the multiwallet work is delivered.

Issue #962, opened 2018-05-25
Support Crypto Listing Markups


jjeffryes:
## User Story
Support cryptocurrency listings having prices above or below the market rate.

### Listing API
- [x] Add a field for a price modifier ("priceModifier").
    - This field is only used for listings with type CRYPTOCURRENCY.
    - This field should have values between -99.99 and 1000.00, with two decimal places.
    - The actual value should be an integer. It represents a percentage between negative 99.99% (a discount) and 1000%.
    - The value should be validated to prevent discounts of 100% or more, or increases above 1000%.
    - The upper limit of 1000% is an arbitrary number, we can change it, presumably no one would successfully sell anything at a 10x markup.

### Listings API
- [x] Add in the acceptedCurrencies array so the client can determine what coin you are paying in (i.e. which coin you're trading for the coinType coin). (search already does this)
- [x] For cryptocurrency listings, add in the priceModifier field so the client can determine how to style the price.

### Sales, Purchases, and Cases APIs
- [x] Add the CoinType (the coin being sold) to these APIs so it can be displayed in the client.
- [x] Add the payment coin to these APIs as well. EDIT: remove incorrect note

rmisio:
> The actual value should be an integer. It represents a percentage between negative 99.99% (a discount) and 1000%.

This is not correct. If we follow suit with other percentage fields, e.g. the ones used for coupons, it should be a float.

rmisio:
> It's only strictly needed for the Cases API right now, but it would be best to add it to all three, so it's in place when we support multicoin wallets.

@jjeffryes Well for Sales it would be needed. For purchases, I suppose it could be inferred from the `cryptoCurrency` the current nodes wallet supports, but I totally agree with "it would be best to add it to all three, so it's in place when we support multicoin wallets".

BazaarDog:
This can be accommodated by the existing price object. 

A 5% markup on coin ABC is is just 1 ABC for 1.05 ABC. The markup an UI and logic can be contained in the client.

jjeffryes:
> This is not correct. If we follow suit with other percentage fields, e.g. like the ones used for coupons, it should be a float.

@cpacia @tyler-smith @mg would it be safe here to use a float like we do with coupons?

placer14:
@jjeffryes IMO, keeping floats away when accuracy of value is important (like with prices) is a good policy in general since limits and edge cases are hard to reason about.

jjeffryes:
[...]

PR #1025, opened 2018-07-19
Payment coin and coin type

ghost:
This PR completes #962 

Features:

- Add `paymentCoin` to orders
- Add `coinType` and `paymentCoin` to cases/sales/purchases db tables
- Set `coinType` and `paymentCoin` when creating cases/sales/purchases db entries
- Migrate existing cases/sales/purchases db entries to set `coinType` and `paymentCoin`
- Migrate existing listing indexes to include `acceptedCurrencies`


cpacia:
@tyler-smith where is the fix for the migration 09 -> 10 bug? Not sure if I see it.

placer14:
@cpacia The migration test for 10 was fixed prior to this. See https://github.com/OpenBazaar/openbazaar-go/commit/9fb3fb72d4a70986136557d56f7fffbf19bbc27f

cpacia:
got it

cpacia (approved):
I think this looks good to me. @placer14 can you also review?

repo/migrations/migration_helpers.go
```diff
@@ -0,0 +1,56 @@
+package migrations
+
+import (
+	"database/sql"
+	"os"
+	"path"
+	"strconv"
+)
+
+func NewDB(repoPath string, dbPassword string, testnet bool) (*sql.DB, error) {
```
placer14: Maybe more like `OpenDB`? Was surprised by the name and that it lived in `migrations` until I came to look.

placer14: One thing to be concerned with the idea of helpers in our migrations will be the problem that arises when the helper breaks across schema versions (if we were to move the location of the database file, for example). I tend to inline the helpers needed for each migration and write new ones. Something like:

`repo/migrations/migration_009.go`
```
func migration009_openDB (...) (...) {
   // actual open logic here
}
```

`repo/migrations/migration_010.go`
```
func migration010_openDB (...) (...) {
   return migration009_openDB(...)
}

// or explicitly show how this schema is changing 
// by manipulating the inputs as needed for migration009_openDB
func migration010_openDB (path string, ...) (...) {
   migration010_dbPath := migration010_dbPathChanges(path)
   return migration009_openDB(migration010_dbPath, ...)
}

```

test/factory/contract.go
```diff
@@ -3,9 +3,10 @@ package factory
 import (
 	"time"
 
+	crypto "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
+
 	"github.com/OpenBazaar/openbazaar-go/pb"
 	"github.com/golang/protobuf/ptypes"
-	crypto "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
```
placer14: Why can't `fmt` make up it's mind? I hate how this makes our history more noisy.

repo/migrations/Migration009.go
```diff
@@ -0,0 +1,386 @@
...
+	Migration009CreateSalesTable     = "create table sales (orderID text primary key not null, contract blob, state integer, read integer, timestamp integer, total integer, thumbnail text, buyerID text, buyerHandle text, title text, shippingName text, shippingAddress text, paymentAddr text, funded integer, transactions blob, needsSync integer, lastDisputeTimeoutNotifiedAt integer not null default 0);"
+	Migration009CreateSalesIndex     = "create index index_sales on sales (paymentAddr, timestamp);"
+	Migration009CreatePurchasesTable = "create table purchases (orderID text primary key not null, contract blob, state integer, read integer, timestamp integer, total integer, thumbnail text, vendorID text, vendorHandle text, title text, shippingName text, shippingAddress text, paymentAddr text, funded integer, transactions blob, lastDisputeTimeoutNotifiedAt integer not null default 0, lastDisputeExpiryNotifiedAt integer not null default 0, disputedAt integer not null default 0);"
```
placer14: I like to indicate in the variable name that these schemas are as they existed prior to this migration.

placer14 (approved):
Some thoughts but nothing that should hold this up. LGTM!

repo/db/util.go
```diff
@@ -0,0 +1,37 @@
...
+	paymentCoin := contract.BuyerOrder.Payment.Coin
+	if paymentCoin != "" {
+		return paymentCoin
+	}
+
+	if len(contract.VendorListings[0].Metadata.AcceptedCurrencies) > 0 {
+		paymentCoin = contract.VendorListings[0].Metadata.AcceptedCurrencies[0]
+	}
+
+	return paymentCoin
```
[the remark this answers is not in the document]
tyler-smith: VendorListings having a single element is already an invariant. We can make it return an error if there's no listing and fail to create the record if that's preferable.

PR #1027, opened 2018-07-19
Add optional asyncID query param to async API calls

### LINKED ISSUE #925: Add ID Parameter to Async Socket Calls
When making POST calls that return async data via web sockets, add an optional parameter for an ID, so clients can specify the ID for the socket messages they will watch for.

This will fix the issue where sometimes the POST data with the ID returns after socket data begins arriving, causing some of it to be missed.

cpacia:
This commit adds an optional query parameter allowing clients to pass in an
ID used for async requests where the response is returned over websockets.

The current approach of generating and returning a random ID may create a race
condition causing the client to miss the response.

closes #925


jjeffryes:
This worked as expected when I tested it.

PR #1051, opened 2018-07-25
BUGFIX: Fix Errorf call in schema/manager.go.

tyler-smith:


tyler-smith:
Rolling this PR into Go 1.10 changes.

PR #1053, opened 2018-07-25
BUGFIX: Use UTC times for notification timestamps as expected by the …

tyler-smith:
…api tests.


tyler-smith:
Failing Travis for some reason. Need to investigate before merging.

PR #1054, opened 2018-07-25
Use Go 1.10 for Travis builds.

ghost:
Updates .travis.yml to use 1.10.3 along with some tweaks to make things work correctly in 1.10.


test_compile.sh
```diff
@@ -1,27 +1,6 @@
...
-go test -coverprofile=repo.db.cover.out ./repo/db
-go test -coverprofile=repo.migrations.db.cover.out ./repo/migrations
-go test -coverprofile=schema.cover.out ./schema
-go test -coverprofile=storage.cover.out ./storage
-go test -coverprofile=storage.dropbox.cover.out ./storage/dropbox
-go test -coverprofile=storage.selfhosted.cover.out ./storage/selfhosted
-echo "mode: set" > coverage.out && cat *.cover.out | grep -v mode: | sort -r | \
-awk '{if($1 != last) {print $0;last=$1}}' >> coverage.out
-rm -rf *.cover.out
+go test -coverprofile=coverage.out ./...
```
placer14: I've been waiting for this! It finally works!

api/gateway.go
```diff
@@ -50,30 +48,15 @@ func NewGateway(n *core.OpenBazaarNode, authCookie http.Cookie, l net.Listener,
...
-		case <-g.shutdownCh:
-			return
-		case <-time.After(5 * time.Second):
-			log.Infof("waiting for server at %s to terminate...", g.listener.Addr())
-
-		}
-	}()
-
-	// Shutdown the listener
-	close(g.shutdownCh)
```
placer14: What was this protecting that it is no longer needed?

placer14 (approved):
GTM. Confused about the shutdownChannel that is no longer needed, though.

api/gateway.go
cpacia: Looks like this code does nothing to me. Amirite?

placer14: That's what I'm confused about it. It's really not obvious to me why it was there.

tyler-smith: It appears that the goroutine with the select was supposed to have a loop around the select, so that it would log every few seconds until is g.shutdownCh is closed. This is a good idiom if the the shutdown occurs somewhere else, but in this case the goroutine is set up and then the channel is closed immediately. The message will never fire in practice and even if it did it would only print once and quit instead of printing until shutdown.

The code never worked as intended and just made debugging test issues I ran into harder to debug because it caused me to focus on a spot in the code that didn't matter.

tyler-smith:
Hmm. Tests were working before but are now failing. Need to investigate.

tyler-smith:
This is showing as failing, but multiples retries have worked. It seems to have been a transient issue.

e.g. https://travis-ci.org/OpenBazaar/openbazaar-go/builds/408207750

tyler-smith:
All fixed and ready to go.

PR #1057, opened 2018-07-26
lint core/order.go

amangale:
Clean up and lint core/order.go


cpacia:
Thanks @amangale we'll review it.

PR #1059, opened 2018-07-26
[WIP] Multiwallet

cpacia:


repo/db/db.go
```diff
@@ -86,6 +86,15 @@ func NewSQLiteDatastore(db *sql.DB, l *sync.Mutex, coinType wallet.CoinType) *SQ
...
 
+type DB struct {
+	SqlDB *sql.DB
+	Lock  *sync.Mutex
+}
+
+func (d *SQLiteDatastore) DB() *DB {
+	return &DB{d.db, d.lock}
+}
+
```
placer14: This isn't consumed by anything yet, but I'm already concerned about what we're passing a whole DB into. I would prefer to keep the DB interactions inside the `db` package.

PR #1060, opened 2018-07-27
Fix bug starting Tor transport

cpacia:


cmd/restore.go
```diff
@@ -213,7 +213,11 @@ func (x *Restore) Execute(args []string) error {
 		if x.TorPassword != "" {
 			torPw = x.TorPassword
 		}
-		auth := &proxy.Auth{Password: torPw}
+		var auth *proxy.Auth
+		if x.TorPassword != "" {
+			torPw = x.TorPassword
+			auth = &proxy.Auth{Password: torPw}
+		}
 		onionTransport, err = oniontp.NewOnionTransport("tcp4", torControl, auth, repoPath, (usingTor && usingClearnet))
```
placer14: Can NewOnionTransport accept `nil`?

PR #1075, opened 2018-08-08
BUGFIX: Don't shadow wallet import in mobile package.

tyler-smith:
This prevents the mobile package from building which, among other things, breaks Travis builds.


tyler-smith:
This is fixed in PR #1072

PR #1081, opened 2018-08-13
Fix crypto markup versions

ghost:


repo/migrations/Migration012.go
```diff
@@ -0,0 +1,257 @@
...
+	AverageRating      float32  `json:"averageRating"`
+	RatingCount        uint32   `json:"ratingCount"`
+	ModeratorIDs       []string `json:"moderators"`
+	AcceptedCurrencies []string `json:"acceptedCurrencies"`
+	CoinType           string   `json:"coinType"`
+}
+
+func Migration012_listingHasNewFeaturesAndOldVersion(sl *pb.SignedListing) bool {
+	return sl.Listing.Metadata.Version == 3 &&
+		sl.Listing.Metadata.PriceModifier != 0
```
placer14: Is there any chance that any part of this chain is nil and this panics?

repo/migrations/Migration012.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// Finish early If no crypto listings
+	if len(cryptoListings) == 0 {
+		return writeRepoVer(repoPath, 13)
+	}
+
+	// Check each crypto listing for markup
+	markupListings := []*pb.SignedListing{}
+	for _, listingAbstract := range cryptoListings {
+		listingJSONBytes, err := ioutil.ReadFile(path.Join(repoPath, "root", "listings", migration012_filenameForSlug(listingAbstract.Slug)))
```
placer14: Put the whole `path.Join` inside the helper?

repo/migrations/Migration012_test.go
```diff
@@ -0,0 +1,564 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/ipfs/go-ipfs/repo/fsrepo"
+	"github.com/libp2p/go-libp2p-crypto"
+	peer "github.com/libp2p/go-libp2p-peer"
+)
+
+const testMigration012_IdentityKeyBase64 = "CAESYHwrVuRp5s2u0w5ykibsR77aHWBmvpcaDq+vU9pv8lOqae31NJYJbdDsOlxVRqQZS/eDfssdd7N/rJmoVbQvPytp7fU0lglt0Ow6XFVGpBlL94N+yx13s3+smahVtC8/Kw=="
+
+var (
+	testMigration012_datadir          = path.Join("/", "tmp", "openbazaar-test")
```
placer14: Can we build a new repo for each test with a random path? `github.com/OpenBazaar/openbazaar-go/schema.GenerateTempPath()`. The fixed directory lends itself to leakage between test runs.

repo/migrations/Migration012_test.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	// Check that correct listing's hashes changed
+	listingsIndex := []*migrations.Migration012_ListingData{}
+	err = json.Unmarshal(listingsIndexJSON, &listingsIndex)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedHashLengths := []int{0, 0, 0, 49, 49}
```
placer14: Are these lengths just known quantities are they based on something in the test?

placer14 (commented):
Some feedback after a tired first-pass.

repo/migrations/Migration012_test.go
tyler-smith: It's a known length - the length of a base58 encoded multihash.

repo/migrations/Migration012.go
tyler-smith: Not really, all listings should have a metadata, but since there's no static assurance I added a check anyways.

repo/migrations/Migration012_test.go
tyler-smith: I think they'll all be 49 but made this check into a looser sanity check that checks the hash length is >= 32.

repo/migrations/Migration012.go
```diff
@@ -0,0 +1,261 @@
...
+		out, err := m.MarshalToString(sl)
+		if err != nil {
+			return err
+		}
+
+		filename := migration012_listingFilePath(repoPath, sl.Listing.Slug)
+		if err := ioutil.WriteFile(filename, []byte(out), os.ModePerm); err != nil {
+			return err
+		}
+		h, err := ipfs.GetHashOfFile(nd, filename)
```
placer14: Is it worth ensuring this function produces a specific result in a test? If the lib changes the hashing, that test could alert us to the problem migration.

repo/migrations/Migration012_test.go
```diff
@@ -0,0 +1,660 @@
...
+		{4, -0.001, false},
+	} {
+		signedListing.Listing.Metadata.Version = test.version
+		signedListing.Listing.Metadata.PriceModifier = test.modifier
+		hasNewFeatures := migrations.Migration012_listingHasNewFeaturesAndOldVersion(signedListing)
+		if hasNewFeatures != test.expected {
+			t.Fatal("Test", i, "failed\nExpected:", test.expected, "\nGot:", hasNewFeatures)
+		}
+	}
+}
```
placer14: <3

placer14 (approved):
Aside from maybe testing the ipfs.Hash behavior with a fixed expectation, this looks GTM. 👍

repo/migrations/Migration012.go
```diff
@@ -0,0 +1,261 @@
...
+	cfg, err := r.Config()
+	if err != nil {
+		return err
+	}
+
+	cfg.Identity = identity
+
+	ncfg := &ipfscore.BuildCfg{
+		Repo:   r,
+		Online: true,
```
cpacia: Why don't you set this to  `Online: false`. Since this is not configuring Tor it would someone's privacy if it tries to make network connections.

tyler-smith: Good call. I was trying to stay as close the the example on cmd/convert.go but overlooked the Online flag when looking for things that should be changed.

cpacia (approved):

PR #1086, opened 2018-08-14
Fix chain sync bug in wallets

cpacia:


vendor/github.com/OpenBazaar/spvwallet/eight333.go
```diff
@@ -572,6 +576,10 @@ func (ws *WireService) handleInvMsg(imsg *invMsg) {
 	// Request the advertised inventory if we don't already have it
 	gdmsg := wire.NewMsgGetData()
 	numRequested := 0
+	shouldSendGetData := false
+	if len(state.requestQueue) == 0 {
+		shouldSendGetData = true
+	}
```
placer14: @tyler-smith had a good comment in chat: `shouldSend := len(state.requestQueue) == 0`

PR #1091, opened 2018-08-15
Fixed #1088

### LINKED ISSUE #1088: GetModeratorFee function is not wallet agnostic
The `GetModeratorFee` function in core/moderation.go currently contains lines like this for both FIXED and FIXED_PLUS_PERCENTAGE fee structures:

```go
var fixed uint64
if strings.ToLower(profile.ModeratorInfo.Fee.FixedFee.CurrencyCode) == "btc" {
	fixed = profile.ModeratorInfo.Fee.FixedFee.Amount
} else {
	fixed, err = n.getPriceInSatoshi(profile.ModeratorInfo.Fee.FixedFee.CurrencyCode, profile.ModeratorInfo.Fee.FixedFee.Amount)
	if err != nil {
		return 0, err
	}
}
```

It obviously should be comparing the currency code to n.Wallet.CurrencyCode() rather than to just "btc" as the user might not be running a btc wallet.

The only reason this isn't a bug is because `n.getPriceInSatoshi` checks the wallet type and just returns the satoshi value if the currency code matches the wallet. 

So either the first IF statement can be changed to something like:
```go
if NormalizeCurrencyCode(profile.ModeratorInfo.Fee.FixedFee.CurrencyCode) == NormalizeCurrencyCode(n.Wallet.CurrencyCode())
```

Or it could be removed entirely and just leave `n.getPriceInSatoshi` to do the work and make sure its commented.

devansh289:


cpacia:
@devansh289 if you look at the case right above it `case pb.Moderator_Fee_FIXED` there is another spot that needs to be updated as well.

devansh289:
@cpacia Oh alright. Will do.

devansh289:
My apologies. But, how do you check why the tests failed? It's one of my first pushes.

cpacia:
Looks like you just need to remove an unused import `core/moderation.go:10:2: imported and not used: "strings"`

And then run `gofmt`.

cpacia:
Thanks

PR #1064, opened 2018-07-31
Make wallet-interface use generic Addresses and Key handling

placer14:
Work in progress tracking #1063 .

PR #1096, opened 2018-08-17
[#1064] Migrate to use wallet-interface ChildKey()

placer14:
Please verify that I captured all of the `ChildKey` migrations within https://github.com/OpenBazaar/openbazaar-go/pull/1066

/cc @amangale


amangale:
@placer14 - The changes for the ChildKey all look fine except core/order.go . There seems to be something funny in that file as the number of occurences of ChildKey present is 2-3 more than what should have been.

PR #1121, opened 2018-08-22
Update /wallet/transactions endpoint with wallet-interface changes

cpacia:


vendor/github.com/OpenBazaar/wallet-interface/datastore.go
```diff
@@ -248,10 +248,33 @@ type Txn struct {
...
+type StatusCode string
+
+const (
+	Unconfirmed StatusCode = "UNCONFIRMED"
+	Pending                = "PENDING"
+	Confirmed              = "CONFIRMED"
+	Stuck                  = "STUCK"
+	Dead                   = "DEAD"
+	Error                  = "ERROR"
+)
```
placer14: I like indicating what these random constants are in their name so we don't mix up wallet-interface's Error with any other Error we might want to express in this package.
`StatusUnconfirmed`, `StatusPending`, `StatusConfirmed` would be appropriate. I would also accept putting them into a status package to consume like `status.Pending`.

api/jsonapi.go
```diff
@@ -2787,7 +2768,7 @@ func (i *jsonAPIHandler) GETTransactions(w http.ResponseWriter, r *http.Request)
 			tx.Thumbnail = m.Thumbnail
 			tx.CanBumpFee = m.CanBumpFee
 		}
-		if status == "DEAD" {
+		if t.Status == "DEAD" {
```
placer14: `wallet.StatusCode(t.Status) == wallet.StatusDead`

PR #1124, opened 2018-08-22
Optionally use counterparty outpoints in dispute resolution

### LINKED ISSUE #1117: Check outpoints exist when signing dispute payout
If for some reason a dispute is received by the moderator and is missing the outpoints (likely because the wallet was not fully synced), and _if_ the moderator attempts to award a majority of the funds to that user, the logic should use the other party's outpoints (which should exist) rather than signing a no-input transaction.

placer14:
Fixes #1117

### DIFF
--- a/repo/dispute_case_record.go
+++ b/repo/dispute_case_record.go
@@ -99,3 +99,20 @@ func (r *DisputeCaseRecord) Contract() *pb.RicardianContract {
 	}
 	return contract
 }
+
+// ResolutionPaymentOutpoints returns the preferred outpoints to be used when resolving
+// a pending DisputeCaseResolution based on the provided PayoutRatio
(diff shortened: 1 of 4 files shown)


repo/dispute_case_record.go
```diff
@@ -99,3 +99,43 @@ func (r *DisputeCaseRecord) Contract() *pb.RicardianContract {
...
+			outpoints = r.VendorOutpoints
+		}
+	}
+	return
+}
+
+// ResolutionPaymentContract returns the preferred contract to be used when resolving
+// a pending DisputeCaseRecord based on the provided PayoutRatio
+func (r *DisputeCaseRecord) ResolutionPaymentContract(ratio PayoutRatio) (contract *pb.RicardianContract) {
+	if ratio.VendorAny() {
```
cpacia: This should return whoever has the majority rather than returning the vendor contract if he receives any payout at all.

placer14:
@cpacia Fixed.

PR #1125, opened 2018-08-23
BUGFIX: Call deferred Close() after checking error.

tyler-smith:
`defer Close()`'s should be called after checking the potential error returned when the `io.Closer` was created.

Travis passed, but coverage fails because this code is uncovered and there's now more of it even though it's more correct.


tyler-smith:
Coverage failed because code is already untested. It's a problem, but not for this PR.

Issue #1129, opened 2018-08-25
Proper Snap builds


tyler-smith:
Currently we have Snap builds but they don't build correctly: https://launchpad.net/~openbazaar/+snaps

It would be great to support Snaps for both the server and the bundle. But none of us know much about Snap AFAIK and doing it right will take some time and research.

We don't want Snap (and other) builds taking up test time. The ideal end goal for ob-go is that launchpad would correctly build valid Snaps on every master commit. If LP can't do that, then we should do it only on new tags unless we get paid CI concurrency.

PR #1130, opened 2018-08-25
CI: Remove Snap building until we can make the builds work correctly.

tyler-smith:
Snaps would be great to support, but they're not working right now and until we fix them they're just wasting build/test time. They take more than 160s to complete.

Issue for proper Snaps is here: #1129


placer14:
Do we want to add these back later and what does that take? Were we using them for anything?

tyler-smith:
@placer14 We're not using them, but I would like to add them back at some point. I would like to have easy packaging available in native formats of our supported operating systems. It would require learning how to build a Snap package and setting up all the plumbing.

I made issue #1129 to track that.

PR #1131, opened 2018-08-25
Remove unused code, goimports, simplify some code

tyler-smith:
Another round of linting changes from TS_travis_linting.


schema/path.go
```diff
@@ -47,5 +47,5 @@ func directoryName(isTestnet bool) (directoryName string) {
 	if isTestnet {
 		directoryName += "-testnet"
 	}
-	return
+	return directoryName
```
placer14: Same as above.

PR #1132, opened 2018-08-25
Cleanup makefile

tyler-smith:
Remove unused Make targets.


cpacia:
Is this not used when building new tags?

tyler-smith:
@cpacia No, that is in deploy.sh in the current master. PR #1133 makes deploy.sh use the existing build.sh script. In either case, the Makefile targets are not used. I'm trying to remove the Docker ones as well and the Makefile will just be protos.

PR #1147, opened 2018-08-28
Linting: Remove unnecessary type conversions and do some goimports.

tyler-smith:
This PR touches a lot of lines but is only removing unnecessary type conversions and goimport'ing the touched files so should be pretty quick to go through.


tyler-smith:
Closing until we can resolve issues around different architectures having different underlying types for the same names in some cases.

PR #1148, opened 2018-08-28
Linting: Fix unused assignments

tyler-smith:
Fix a bunch of places where we assign values that are never used. Many are unchecked errors or other bugs, some are just unnecessary and had no use so they get assigned to `_`.

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -1371,7 +1371,7 @@ func (i *jsonAPIHandler) GETListing(w http.ResponseWriter, r *http.Request) {
 		OrigName:     false,
 	}
 	if peerId == "" || strings.ToLower(peerId) == "listing" || peerId == i.node.IPFSIdentityString() {
-		sl := new(pb.SignedListing)
+		var sl *pb.SignedListing
 		_, err := cid.Decode(listingId)
 		if err == nil {
 			sl, err = i.node.GetListingFromHash(listingId)
@@ -1726,7 +1726,6 @@ func (i *jsonAPIHandler) GETModerators(w http.ResponseWriter, r *http.Request) {
 					j++
 				}
 			}
-			xs = (xs)[:j]
 		}
 		peerInfoList, err := ipfs.FindPointers(i.node.IpfsNode.Routing.(*routing.IpfsDHT), ctx, core.ModeratorPointerID, 64)
 		if err != nil {
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -1716,7 +1716,7 @@ func (i *jsonAPIHandler) GETModerators(w http.ResponseWriter, r *http.Request) {
 
 	ctx := context.Background()
 	if !async {
-		removeDuplicates := func(xs []string) {
+		removeDuplicates := func(xs []string) []string {
 			found := make(map[string]bool)
 			j := 0
 			for i, x := range xs {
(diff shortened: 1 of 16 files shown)


tyler-smith:
Travis is succeeding, but Coveralls is marking the error logging/propogation as a decrease in code coverage however we were always missing coverage there, it just wasn't being counted before.

api/jsonapi.go
```diff
@@ -1726,7 +1726,6 @@ func (i *jsonAPIHandler) GETModerators(w http.ResponseWriter, r *http.Request) {
 					j++
 				}
 			}
-			xs = (xs)[:j]
```
cpacia: Why is this not needed?

core/import.go
```diff
@@ -45,7 +45,6 @@ func (n *OpenBazaarNode) ImportListings(r io.ReadCloser) error {
 	indexLock := new(sync.Mutex)
 	wg := new(sync.WaitGroup)
 
-	stringPool := sync.Pool{New: func() interface{} { var s string; return s }}
```
cpacia: Also not sure why this isn't needed as well

core/order.go
```diff
@@ -775,11 +784,14 @@ func (n *OpenBazaarNode) CalculateOrderTotal(contract *pb.RicardianContract) (ui
 			return 0, err
 		}
 		itemTotal += satoshis
-		selectedSku, err := GetSelectedSku(l, item.Options)
```
cpacia: This is returning the selected sku, why is being ignored?

mobile/node.go
```diff
@@ -444,9 +444,6 @@ func newHTTPGateway(node *core.OpenBazaarNode, ctx commands.Context, authCookie
 		return nil, fmt.Errorf("newHTTPGateway: manet.Listen(%s) failed: %s", gatewayMaddr, err)
 	}
 
-	// We might have listened to /tcp/0 - let's see what we are listing on
-	gatewayMaddr = gwLis.Multiaddr()
```
cpacia: why is this going away?

api/jsonapi.go
tyler-smith: This spot is actually a bug, which I missed initially. I'll fix it instead of just removing the assignment.

core/import.go
tyler-smith: I don't know why it's not needed, but it's not used. `stringPool` is defined and never used. `stringSlicePool` is used though.

mobile/node.go
tyler-smith: `gatewayMaddr` is never used after that assignment.

core/import.go
tyler-smith: Okay, so these pools are being used ineffectively. It's using fairly simple/low memory types and then is reassigning over them pool'ed versions anyways. I've removed them to make the code clearer.

core/import.go
```diff
@@ -193,18 +189,14 @@ listingLoop:
 			pos, ok = fields["image_urls"]
 			if ok {
 				listing.Item.Images = []*pb.Listing_Item_Image{}
-				imageUrls := stringSlicePool.Get().([]string)
-				imageUrls = []string{}
-				imageUrls = strings.Split(record[pos], ",")
```
placer14: I have no idea why or what is going on here with this triple-assignment attempt. What purpose does this even serve?

core/import.go
placer14: This sync.Pool usage is a pattern I've never encountered. It seems to be used to just generate zero strings values? But... it's not clear why the Pool is required instead of just inlining the var decl.

core/import.go
tyler-smith: That's a great question. My guess is the thought was the first one gets the memory location from the pool, the second is clearing it out, and the 3rd is setting it to desired value?

Regardless, I think the new code is more correct.

PR #1150, opened 2018-08-29
BUGFIX: Fix build on Windows

tyler-smith:
This reverts commit f8640a39cc754a21c94b0ea77a406fbd508b2b67.

It was discovered that syscall.Stdin on Windows is not an int
like on other systems, so it must be converted here.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -241,7 +241,7 @@ func (x *Start) Execute(args []string) error {
 	if sqliteDB.Config().IsEncrypted() {
 		sqliteDB.Close()
 		fmt.Print("Database is encrypted, enter your password: ")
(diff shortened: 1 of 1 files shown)


cmd/start.go
```diff
@@ -241,7 +241,7 @@ func (x *Start) Execute(args []string) error {
 	if sqliteDB.Config().IsEncrypted() {
 		sqliteDB.Close()
 		fmt.Print("Database is encrypted, enter your password: ")
-		bytePassword, _ := terminal.ReadPassword(syscall.Stdin)
+		bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
```
placer14: Should we handle the err?

PR #1165, opened 2018-09-06
CI: Add goconst check to linter

tyler-smith:
Check for string values that are used too many times (currently configured to 10).

Refactor 2 recurring strings to encapsulate the values in functions.

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -229,6 +229,10 @@ func SanitizedResponseM(w http.ResponseWriter, response string, m proto.Message)
 	fmt.Fprint(w, string(out))
 }
 
+func isNullJSON(jsonBytes []byte) bool {
+	return string(jsonBytes) == "null"
+}
+
 func (i *jsonAPIHandler) POSTProfile(w http.ResponseWriter, r *http.Request) {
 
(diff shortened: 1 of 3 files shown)


core/import.go
```diff
@@ -160,7 +160,7 @@ listingLoop:
 				errChan <- fmt.Errorf("Error in record %d: %s", i, "price is a mandatory field")
 				return
 			}
-			if NormalizeCurrencyCode(listing.Metadata.PricingCurrency) != "BTC" {
```
cpacia: This command is going to need to be fixed at some point. It was never updated since we allowed multiple currencies.

api/jsonapi.go
```diff
@@ -229,6 +229,10 @@ func SanitizedResponseM(w http.ResponseWriter, response string, m proto.Message)
 	fmt.Fprint(w, string(out))
 }
 
+func isNullJSON(jsonBytes []byte) bool {
+	return string(jsonBytes) == "null"
+}
```
placer14: 👏

core/import.go
placer14: @cpacia Can you describe what change needs to be made in a new issue so we don't lose it?

PR #1170, opened 2018-09-06
CLEANUP: Remove some unused code and an unused assignment

tyler-smith:
Also combines a string declaration with its assignment.

### DIFF
--- a/repo/migrations/Migration006_test.go
+++ b/repo/migrations/Migration006_test.go
@@ -109,15 +109,10 @@ func TestMigration006(t *testing.T) {
 	}
 
 	for _, listing := range migratedListings {
-		actualModeratorIDs := make([]string, 0)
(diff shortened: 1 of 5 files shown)


repo/migrations/Migration006_test.go
```diff
@@ -109,15 +109,10 @@ func TestMigration006(t *testing.T) {
 	}
 
 	for _, listing := range migratedListings {
-		actualModeratorIDs := make([]string, 0)
-		for _, peerID := range listing.ModeratorIDs {
-			actualModeratorIDs = append(actualModeratorIDs, peerID)
-		}
-
-		if !reflect.DeepEqual(expectedStoreModerators, actualModeratorIDs) {
```
placer14: I wrote this, but I'm not sure what I was thinking when I was doing it. :|

PR #1174, opened 2018-09-12
Add ethereum config

amangale:


repo/init.go
```diff
@@ -206,6 +206,20 @@ func addConfigExtensions(repoRoot string) error {
 				MediumFeeDefault: 10,
 				LowFeeDefault:    5,
 			},
+			ETH: &schema.CoinConfig{
```
cpacia: We will probably need a migration for this since it wont be in existing configs. We can probably hold off on that though and bundle it with the other migration we need to do related to the insight servers.

placer14: @cpacia I don't know what that scope is and would like it to be tracked. Can you please make an issue for the required migrations so we don't lose it?

placer14:
Reopening as merging multiwallet (which this PR was targetted at) was resolved without this PR.

wallet/builder.go
```diff
@@ -127,8 +125,24 @@ func NewMultiWallet(cfg *WalletConfig) (multiwallet.MultiWallet, error) {
...
+				coin.MaxFee = uint64(cfg.ConfigFile.ETH.MaxFee)
+				if !testnet {
+					coin.ClientAPIs = []string{cfg.ConfigFile.ETH.API}
+				} else {
+					coin.ClientAPIs = []string{cfg.ConfigFile.ETH.APITestnet}
+				}
+				coin.Options = schema.EthereumDefaultOptions()
+			}
 		}
+		coinCfgs = append(coinCfgs, coin)
```
placer14: @amangale This looks like it was missing in both the LTC and ETH cases. Was this intentional?

wallet/builder.go
```diff
@@ -44,14 +44,15 @@ func NewMultiWallet(cfg *WalletConfig) (multiwallet.MultiWallet, error) {
...
-	coinsToUse[wallet.Bitcoin] = true
-	coinsToUse[wallet.BitcoinCash] = true
-	coinsToUse[wallet.Zcash] = true
-	coinsToUse[wallet.Litecoin] = true
+	walletEnableOpt := make(map[wallet.CoinType]bool)
+	walletEnableOpt[wallet.Bitcoin] = true
+	walletEnableOpt[wallet.BitcoinCash] = true
+	walletEnableOpt[wallet.Zcash] = true
+	walletEnableOpt[wallet.Litecoin] = true
+	walletEnableOpt[wallet.Ethereum] = false
```
placer14: Minor detail, but explicitness around having ETH off doesn't hurt.

PR #1175, opened 2018-09-14
Update country codes

jjeffryes:
This adds the United States Minor Outlying Islands, and updates Swaziland to Eswatini.

It should be merged in conjunction with the matching client PR. 

https://github.com/OpenBazaar/openbazaar-desktop/pull/1562


core/listings.go
```diff
@@ -1219,7 +1219,7 @@ func validatePhysicalListing(listing *pb.Listing) error {
 		for _, region := range shippingOption.Regions {
 			if int(region) == 0 {
 				return errors.New("Shipping region cannot be NA")
-			} else if int(region) > 246 && int(region) != 500 {
+			} else if int(region) > 247 && int(region) != 500 {
```
placer14: This is painful but okay for now. Tracking https://github.com/OpenBazaar/openbazaar-go/issues/1177 for later.

pb/countrycodes.pb.go
```diff
@@ -233,7 +233,7 @@ const (
 	CountryCode_SUDAN                          CountryCode = 209
 	CountryCode_SURINAME                       CountryCode = 210
 	CountryCode_SVALBARD                       CountryCode = 211
-	CountryCode_SWAZILAND                      CountryCode = 212
```
placer14: I checked through the server codebase and we don't appear to reference by the `pb.CountryCode_SWAZILAND`, `pb.SWAZILAND` or indirectly through any accessor. Hopefully this won't bite us later.

cpacia: We need to recompile. Just changing the file will likely produce errors.

Issue #1177, opened 2018-09-14
Remove numeric references for pb.CountryCodes


placer14:
https://github.com/OpenBazaar/openbazaar-go/blob/b183b30a1832930e0043b2298679684bacbb5f82/core/listings.go#L1220-L1222

These should be abstracted away from their hardcoded values.

PR #1187, opened 2018-09-23
Removes numeric reference to pb.CountryCode

allen-munsch:
Removes numeric reference to pb.CountryCode / Region using 
`proto.EnumValueMap("CountryCode")[region.String()]`.

#1177

### DIFF
--- a/core/listings.go
+++ b/core/listings.go
@@ -1220,9 +1220,7 @@ func validatePhysicalListing(listing *pb.Listing) error {
 			if int(region) == 0 {
 				return errors.New("Shipping region cannot be NA")
 			} else {
-				if val, ok := proto.EnumValueMap("CountryCode")[region.String()]; ok {
-					log.Debugf("Region %s:%d in pb.CountryCode", region, val)
-				} else {
+				if val, ok := proto.EnumValueMap("CountryCode")[region.String()]; !ok {
 					return fmt.Errorf("Invalid shipping region [ %s ]: ", region)
 				}
 			}
--- a/core/listings.go
+++ b/core/listings.go
@@ -1219,10 +1219,13 @@ func validatePhysicalListing(listing *pb.Listing) error {
 		for _, region := range shippingOption.Regions {
 			if int(region) == 0 {
 				return errors.New("Shipping region cannot be NA")
-			} else if int(region) > 247 && int(region) != 500 {
-				return errors.New("Invalid shipping region")
+			} else {
+				if val, ok := proto.EnumValueMap("CountryCode")[region.String()]; ok {
+					log.Debugf("Region %s:%d in pb.CountryCode", region, val)
+				} else {
+					return fmt.Errorf("Invalid shipping region [ %s ]: ", region)
+				}
 			}
-
 		}
 		if len(shippingOption.Regions) > MaxCountryCodes {
 			return fmt.Errorf("Number of shipping regions is greater than the max of %d", MaxCountryCodes)
--- a/core/listings.go
+++ b/core/listings.go
@@ -104,6 +104,12 @@ type ListingData struct {
 	CoinType           string    `json:"coinType"`
 }
 
+var (
+	ErrShippingRegionMustBeSet          = errors.New("shipping region must be set")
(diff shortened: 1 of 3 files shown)


cpacia:
That's for the PR! We'll review.

allen-munsch:
@tyler-smith thanks! honestly wasn't sure how to do !ok. I wasn't able to see how core was being tested, other than coverage analysis `go test -coverprofile=core.cover.out ./core`, did I miss something?

core/listings.go
```diff
@@ -1219,10 +1219,11 @@ func validatePhysicalListing(listing *pb.Listing) error {
 		for _, region := range shippingOption.Regions {
 			if int(region) == 0 {
 				return errors.New("Shipping region cannot be NA")
-			} else if int(region) > 247 && int(region) != 500 {
-				return errors.New("Invalid shipping region")
+			} else {
```
placer14: The logic before and after has changed. The logic before seems to error on country codes between 501-508 inclusive, whereas now those values would be ok. I would prefer to maintain existing logic within this PR. How about something like:

Some pseudo:
`if ok := validShippingRegion(region); !ok { ...`
```
func validShippingRegion(region) bool {
  return EnumValueMapLookupIsOk(region) && RegionIsNotAContinent(region)
}
```

placer14:
@allen-munsch The `core` package has very little unit test coverage. Would love it if you unit tested the `isValidShippingRegion` method that you extract.

core/listings.go
placer14: I wouldn't even be upset if you pulled in the `int(region) == 0` check in there, too. Maybe return a `(bool, error)` instead?

allen-munsch:
#1207 @placer14  are there some extra tests I can write to check for this? or if you'd like me to pull on 1207 instead, let me know. 

It's been fun learning go. Thanks for the reviews!

core/listings.go
```diff
@@ -1175,6 +1175,27 @@ func validateListing(listing *pb.Listing, testnet bool) (err error) {
...
+		}
+		if ok {
+			// region exists but it is a continent
+			if int32(region) > 500 {
+				return 3
+			}
+		}
+	}
+	return 0
+}
```
placer14: Please return a descriptive `error` instead of an error number? If you like, you can enumerate them like so:
```
var (
  ErrShippingRegionMustBeSet = errors.New("shipping region must be set")
  ErrShippingRegionUndefined = errors.New("undefined shipping region")
  ErrShippingRegionMustNotBeContinent = errors.New("cannot specify continent as shipping region")
)

func ValidShippingRegion(option) error {
  ...
  if int32(region) > 500 {
    return ErrShippingRegionMustNotBeContinent
  }
  ...
  return nil
}

if err := ValidShippingRegion(shippingOption); err != nil {
  return fmt.Errorf("invalid shipping option (%s): %s", shippingOption.String(), err.Error())
}
```

core/listings_test.go
```diff
@@ -12,3 +14,77 @@ func TestFactoryCryptoListingCoinDivisibilityMatchesConst(t *testing.T) {
...
+		507: true,
+		508: true,
+		// !exist
+		509: true,
+		510: true,
+		511: true,
+		// some random numbers
+		5678:   true,
+		123456: true,
+	}
```
placer14: Instead of specifying a bool, each example could point at the expected error message it should receive. Valid options can be tested here as well by pointing to a `nil` error. Appreciate the thorough test on both positive and negative values.

placer14:
@allen-munsch If you're feeling up to it, I'd welcome a PR on #1207. Please make sure you re-read the ticket and understand it before digging in.

placer14:
@allen-munsch We've had to revert the change that this PR and #1207 is dependent upon in master.  I haven't heard from you about #1207 but I'd like to get it resolved before we merge this PR. Could you let me know if you think you can handle #1207 in the next few days? Otherwise, I'm planning on getting started on it so this PR doesn't get stale. (Wouldn't want the effort to go to waste.) If you could let me know ASAP, that would be ideal. 🍻

placer14:
Closing in preference of merging #1225

PR #1193, opened 2018-09-26
Fix bugs saving and loading IPNS cache

cpacia:


placer14:
@cpacia How are we tracking the changes made in the vendor/ipfs-go package?

cpacia:
We do have a fork of go-ipfs as OpenBazaar/go-ipfs. I will push these changes upstream.

Issue #1205, opened 2018-10-03
Add unconvert linter rules


placer14:
In order to improve the overall quality of the repo, we are enabling the linter within travis but required our code to be made compliant with the lint rules before enabling each ruleset.

See https://github.com/OpenBazaar/openbazaar-go/issues/1169 for steps to enable rules and run linter.

PR #1209, opened 2018-10-04
Enable errcheck linter rules

frodeaa:
fix all errors reported by the linter

fixes: https://github.com/OpenBazaar/openbazaar-go/issues/1203

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,6 +31,7 @@ linters-settings:
 
 linters:
   enable:
+    - errcheck
     - gofmt
     - ineffassign
     - staticcheck
@@ -47,7 +48,6 @@ linters:
     - goconst
     - govet
   disable:
-    - errcheck
     - golint
     - megacheck
     - goimports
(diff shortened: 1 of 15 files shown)


tyler-smith:
This fixes many of the issues, but not all of them so the linter will fail with `errcheck` added. We should probably merge the fixes in this PR but not add the `errcheck` linter yet.

One place it doesn't fix that we need to decide how to handle is calling functions with error returns as go routines. I think the only way to handle these is to wrap them and log errors. e.g.

```
func mightReturnAnError() error {
  return errors.New("test")
}

// Instead of this:
go mightReturnAnError()

// Do this:
go func(){
  err := mightReturnAnError()
  if err != nil {
    log.Debug(err)
  }
}()
```

@placer14 @cpacia thoughts?

frodeaa:
just ignore this PR, was working on https://github.com/OpenBazaar/openbazaar-go/issues/1205, fixing unconvert, not errcheck  :sleeping:

frodeaa:
@tyler-smith linters can be disabled with `//nolint` https://github.com/golangci/golangci-lint#false-positives

PR #1210, opened 2018-10-04
Enable unconvert linter rules

frodeaa:
fix all errors reported by the linter

fixes: https://github.com/OpenBazaar/openbazaar-go/issues/1205

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,6 +31,7 @@ linters-settings:
 
 linters:
   enable:
+    - unconvert
     - gofmt
     - ineffassign
     - staticcheck
@@ -51,5 +52,4 @@ linters:
     - golint
     - megacheck
     - goimports
-    - unconvert
     - prealloc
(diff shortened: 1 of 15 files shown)


tyler-smith:
The `int(syscall.Stdin)` to `syscall.Stdin` changes break builds on Windows because build tags are used to defined `syscall.Stdin` as a different type than on other platforms.

We'll have to `// nolint` those lines.

tyler-smith:
I don't think that there is runtime risk introduced here but if somebody else knows otherwise let me know. Because it only deals with type conversions, at worse it should only break compilation (like the `syscall.Stdin` issue does).

It still needs to be tested well in various environments, but I don't think runtime bugs are possible here.

placer14:
Tested compilation successfully on the following OS versions:
- MacOS 10.12, go 1.10.3
- Ubuntu 18.04
- Awaiting windows confirmation

PR #1217, opened 2018-10-09
Partial cleanup for adding golint

ianbibby:
Fix golint errors in api/jsonapi.go.


placer14:
Hey @ianbibby. Thanks for the submission... this a good progress. I think you'll realize there are some fields which will fail this lint rule. I would suggest staying away from the protobuf and other struct fields as some of those are exported out of the node. Function names and variables etc should not be a problem. I would welcome more of these small PRs toward getting `golint`'s rule added to CI.

placer14:
Closing in favor of #1250

Issue #1188, opened 2018-09-24
Additional fields in the notification response for orders


drwasho:
A typical order related notification looks like this:

```JSON
{
  "notifications": [
    {
      "notification": {
        "buyerHandle": "das_test",
        "buyerId": "QmVLRA8hzADMwYDxXXFgHf2i9bbmbcm9WQHVojeAensNcR",
        "notificationId": "QmUh2aupaEoudW5feEFeyGmturjSDWxpfdDnzKeuZhx4rR",
        "orderId": "QmZwQbg4XPCRqFqWhDw4Z1694jpD7GjdxuxCebkXpwP2t7",
        "slug": "vintage-dress-physical-w-options2",
        "thumbnail": {
          "small": "QmVsoT9iabv6GZhxhvtjSpQMJA6QyMivGTs6MmHJr6TBm9",
          "tiny": "QmbjyAxYee4y3443kAMLcmRVwggZsRDKiyXnXus1qdJJWz"
        },
        "title": "Vintage dress (physical; w/ options)",
        "type": "order"
      },
      "read": false,
      "timestamp": "2018-06-08T00:36:16Z",
      "type": "order"
    }
  ],
  "total": 123,
  "unread": 122
}
```

There are two issues:

1. Currently, we have to make a separate API call (`GET /ob/order/{orderId}`) to fetch the *price* of the item from the order notification. This is a problem for mobile as we have this design for displaying order-related notifications:

![image](https://user-images.githubusercontent.com/3887694/45936390-1ed77a00-bff9-11e8-91c2-02d2d40f7d98.png)

Having to make an GET /ob/order API call just to fetch the price is wasteful and adds unnecessary overhead.

2. It is impossible to tell from this notification alone whether or not the node is the buyer or seller. We have to add additional logic in a client to check against the node's own peerId against the `buyerId` to determine whether or not they are the seller.

---

I propose that we:

- [x] Add a `price` object, with the `value` and `currencyCode` to the order notification
- [ ] Add a field called `vendor`, which is a boolean to denote whether the node is the seller or not. (We are breaking this out into another issue since the scope is not clear yet. See #1230)

placer14:
@drwasho Can you show an example of what these fields would look like when implemented? Curious about the diff between value and price. Additionally, how about putting the vendorID into the payload and the user can compare their own VendorID to either the Vendor or the Buyer ID to see which party they are to this transaction?

drwasho:
[...]

PR #1223, opened 2018-10-10
Add price and listingType fields to order websocket notification

placer14:
Completes #1188. New fields in `order` websocket notification.


placer14:
Final example of order websocket notification:
```
{
  "notifications": [
    {
      "notification": {
        "buyerHandle": "das_test",
        "listingType": "SERVICE",
        "buyerId": "QmVLRA8hzADMwYDxXXFgHf2i9bbmbcm9WQHVojeAensNcR",
        "notificationId": "QmUh2aupaEoudW5feEFeyGmturjSDWxpfdDnzKeuZhx4rR",
        "orderId": "QmZwQbg4XPCRqFqWhDw4Z1694jpD7GjdxuxCebkXpwP2t7",
        "price": {
          "amount": 1000,
          "currencyCode": "eth",
          "priceModifier": 0.0,
          "coinDivisibility": 100000000
        },
        "slug": "vintage-dress-physical-w-options2",
        "thumbnail": {
          "small": "QmVsoT9iabv6GZhxhvtjSpQMJA6QyMivGTs6MmHJr6TBm9",
          "tiny": "QmbjyAxYee4y3443kAMLcmRVwggZsRDKiyXnXus1qdJJWz"
        },
        "title": "Vintage dress (physical; w/ options)",
        "type": "order"
      },
      "read": false,
      "timestamp": "2018-06-08T00:36:16Z",
      "type": "order"
    }
  ],
  "total": 123,
  "unread": 122
}
```

core/coin.go
```diff
@@ -0,0 +1,44 @@
...
+		Currency("zec"),
+		Currency("tbtc"),
+		Currency("tbch"),
+		Currency("teth"),
+		Currency("tltc"),
+		Currency("tzec"),
+	}
+)
+
+// Currency describes
```
tyler-smith: This godoc comment should be completed, and all the other exported identifiers should probably have them as well (when we have golint in CI it'll be required for all changes).

core/coin.go
```diff
@@ -0,0 +1,44 @@
...
+	"errors"
+	"strings"
+)
+
+// DefaultCurrencyDivisibility - decimals for price
+const DefaultCurrencyDivisibility uint32 = 1e8
+
+var (
+	ErrUnsupportedCurrencyCode = errors.New("unsupported currency code")
+	SupportedCurrencies        = []Currency{
```
cpacia: I'm not understanding why we are hardcoding a default list of SupportedCurrencies rather than loading the list of actual currencies from the multiwallet. 

This seems like a particularly bad way of doing it as everything else in the app is designed around not hardcoding values related to wallets. 

Same with the `DefaultCurrencyDivisibility`. The currency divisibility is available from the wallet interface. I don't know why we would hardcode it in openbazaar-go.

placer14: Because the wallet isn't guaranteed to be running any particular currency and the server would still need to provide divisibility for the currency even if the wallet doesn't know about it. (As in the case where a server is using SPV and doesn't have divisibility for ETH available there.) I don't think we should expect the currency definition to be coupled to the wallet since the wallet will not necessarily know about all of the currencies on the network.

core/coin.go
placer14: I think I started down this path and got distracted elsewhere. Will do.

core/coin.go
cpacia: Neither of that sounds correct. You use it in:

1) creating the notification in the listener. You'll never being calling that function with a currency the wallet doesn't support

2) in crypto listings. Which isn't even an appropriate place to use it because the divisibility is just defaulted to the same divisibility for every coin. It doesn't vary by coin.

placer14: Besides we were already doing this, the only reason this was extracted was because we're doing it in two places now. The only thing that's new is that I have a constructor that creates a `Currency` and decorated it with behavior `Divisibility()` instead of passing through a constant for any string input. I could relax the error case, but modularizing this behavior is the right thing to do. And if you help me understand why my case above isn't a problem, I could make it request the data from the wallet-interface instead (if you'll let me know where to retrieve it).

core/currency.go
```diff
@@ -0,0 +1,5 @@
+package core
+
+// DefaultCurrencyDivisibility is the Divisibility of the Currency if not
+// defined otherwise
+const DefaultCurrencyDivisibility uint32 = 1e8
```
placer14: I left this here. Hope no one minds.

core/listings.go
```diff
@@ -1286,7 +1279,14 @@ func validateCryptocurrencyListing(listing *pb.Listing) error {
 		return ErrCryptocurrencyListingCoinTypeRequired
 	}
 
-	if listing.Metadata.CoinDivisibility != coinDivisibilityForType(listing.Metadata.CoinType) {
+	var expectedDivisibility uint32
+	if wallet, err := n.Multiwallet.WalletForCurrencyCode(listing.Metadata.CoinType); err != nil {
+		expectedDivisibility = DefaultCurrencyDivisibility
+	} else {
+		expectedDivisibility = uint32(wallet.ExchangeRates().UnitsPerCoin())
```
placer14: Would be nice if UnitsPerCoin matched types with the listing data. (https://github.com/OpenBazaar/wallet-interface/issues/9)

wallet/listeners/transaction_listener.go
```diff
@@ -212,6 +219,18 @@ func (l *TransactionListener) processSalePayment(txid string, output wallet.Tran
...
+func currencyDivisibilityFromContract(mw multiwallet.MultiWallet, contract *pb.RicardianContract, orderID string) uint32 {
+	var currencyDivisibility = contract.VendorListings[0].Metadata.CoinDivisibility
+	if currencyDivisibility != 0 {
+		return currencyDivisibility
+	}
+	wallet, err := mw.WalletForCurrencyCode(contract.BuyerOrder.Payment.Coin)
+	if err == nil {
+		return uint32(wallet.ExchangeRates().UnitsPerCoin())
+	}
+	return core.DefaultCurrencyDivisibility
```
placer14: Prefer divisibility from listing, then check wallet's preference, then fail to default as a last resort.

cpacia:
LGTM

Issue #1207, opened 2018-10-04
Previous CountryCode protobuf change breaks backward compatibility


placer14:
Existing listings which ship to all regions will have `SWAZILAND` enumerated in the listing itself. It will probably exist in prior orders and should remain compatible. (Bug created in https://github.com/OpenBazaar/openbazaar-go/pull/1175).

Resolution requires:
- [ ] Enable `allow_alias = true` feature of Protobuffers can Unmarshal `212` to `ESWATINI` while still accepting `SWAZILAND` as a value being Marshaled. This should be proven in a unit test. (https://developers.google.com/protocol-buffers/docs/proto#enum)

 - or -

- [ ] Bumping the listing version number so all future generated/updated listings have the new value.
- [ ] Creating migration to transform existing listings which include `SWAZILAND` in the CountryCodes portion and replace it with `ESWATINI` and increase the version number to match the recently bumped value. Migration should be unit tested as well.

In addition...

- [ ] the commit which causes this bug was reverted in `master` until we get this resolved. We need to reapply https://github.com/OpenBazaar/openbazaar-go/commit/d24c39044df4ab48e8b9e03c6ac38290dd744c09 as part of this fix as well.

Calling `POST /ob/estimatetotal` against [this listing](https://gateway.ob1.io/ipfs/zb2rhXdiuRc5GhfmnvP35DTQNur3cGRUBfL3WNNSF3ZKHEkJd) should produce the error as shown in the screengrab.
![image 2](https://user-images.githubusercontent.com/305332/46451565-e5bfa800-c764-11e8-966f-483fe566dc82.png)

Note: A development environment is provided via Docker and Docker Compose to allow easy compilation of protobufs. After docker is installed, you can run `docker-compose run --entrypoint "make protos" dev` and it will build the dev environment and compile the protobufs in place. `docker-compose down --rmi all -v` will remove all of the built up docker infrastructure on your computer if you want to get rid of it.

placer14:
[...]

PR #1225, opened 2018-10-11
#1177 #1207 Testing pb.CountryCodes, allow_alias SWAZILAND, ESWATINI

allen-munsch:
Attempt at merging/rebasing fixes for #1177, #1207, as well as keeping history for https://github.com/OpenBazaar/openbazaar-go/commit/c2fe645351bc5623f0573f9bb7039646ebae8275


placer14:
@allen-munsch Sorry for the slow reply. Appreciate you keeping the commits the same and obvious. This appears to cover https://github.com/OpenBazaar/openbazaar-go/pull/1187 as well so I assume we can close that and merge this instead. Reviewing this now.

core/listings_test.go
```diff
@@ -104,35 +80,7 @@ func TestListingProtobufAlias(t *testing.T) {
...
-			t.Fatal(err)
-		}
-		for _, region := range unmarshalled2Listing.ShippingOptions[0].Regions {
-			if region != pb.CountryCode(pb.CountryCode_SWAZILAND) && region != pb.CountryCode(pb.CountryCode_ESWATINI) {
-				if id == 0 || id == 1 {
-					t.Logf("( %v ) : should fail : SWAZILAND/ESWATINI proto allow_alias ", unmarshalled2Listing.ShippingOptions[0].Regions)
-				}
-				if id != 0 && id != 1 {
-					t.Fatalf("( %v ) : failed : SWAZILAND/ESWATINI proto allow_alias", unmarshalled2Listing.ShippingOptions[0].Regions)
-				}
```
placer14: Sorry @allen-munsch, your cleverness has out done you. I really didn't understand what was being tested here and felt okay with removing it. Please take a look at the other changes I applied in this commit with regard to the error messages (which still indicate failure but also document how the test is expected to pass).

test/factory/listing.go
```diff
@@ -109,32 +109,13 @@ func NewCryptoListing(slug string) *pb.Listing {
...
-	return listing
-}
-
-func NewShippingRegionsProtoBufAlias(slug string, countrycodes []pb.CountryCode) *pb.Listing {
-	listing := NewListing(slug)
-	listing.ShippingOptions = []*pb.Listing_ShippingOption{
-		{
-			Name:    "usps",
-			Type:    pb.Listing_ShippingOption_FIXED_PRICE,
-			Regions: countrycodes,
```
placer14: See the commit message for the rationale here. The two factories were quite similar and I was able to reduce these two functions to just one extra line after the factory's result in the unit test itself.

placer14:
I'd appreciate a second look from @tyler-smith or @cpacia. 🍻

cpacia:
This LGTM. Thanks @allen-munsch !

Issue #1043, opened 2018-07-23
Protect new Message Types from being omitted in net/retriever.MessageProcessingOrder


placer14:
Background: https://github.com/OpenBazaar/openbazaar-go/pull/1040#discussion_r204418698

As a developer, I expect a new message type to raise a failure in the unit test suite when not included in either the `MessageProcessingOrder` or a blacklist indicating it should not be included so I will not forget to include it in the future.

ianbibby:
@placer14 for comments on PR #1229 .

PR #1229, opened 2018-10-15
Test for MessageProcessingOrder omissions

ianbibby:
This PR addresses #1043.

It adds a test that makes use of the pb.Message_MessageType_value map comparing its values against the contents of net/retriever.MessageProcessingOrder, in addition to a BlackList map for deliberate omissions.

When new message types are added and not included in MessageProcessingOrder (or the BlackList), the unit test suite will fail with:
```
--- FAIL: TestEnsureNoOmissionsInMessageProcessingOrder (0.00s)
Found 1 omission(s): map[FOO:123]
```

### DIFF
--- a/net/retriever/retriever_test.go
+++ b/net/retriever/retriever_test.go
@@ -1,0 +1,39 @@
+package net
+
+import (
+	"testing"
+
+	"github.com/OpenBazaar/openbazaar-go/pb"
+)
+
+// TestEnsureNoOmissionsInMessageProcessingOrder ensures that
+// new message types are either included in MessageProcessingOrder
+// or added to the BlackList, indicating it should not be added.
+func TestEnsureNoOmissionsInMessageProcessingOrder(t *testing.T) {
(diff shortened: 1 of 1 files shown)


net/retriever/retriever_test.go
```diff
@@ -0,0 +1,39 @@
...
+	for _, msgType := range MessageProcessingOrder {
+		delete(messages, msgType.String())
+	}
+	// BlackList check
+	for msgType := range BlackList {
+		delete(messages, msgType.String())
+	}
+
+	if l := len(messages); l > 0 {
+		t.Errorf("Found %d omission(s): %v", l, messages)
```
placer14: I like for unit test errors to describe the expected outcome:
`"found %d unexpected message types which are not considered in MessageProcessOrder: %v"`

(Note the error message is not capitalized... this is also idiomatic in go error messages generally.)

cpacia:
LGTM

PR #1240, opened 2018-10-19
improved README

StrikerRUS:
Fixed and added links, added missed dots and removed reference to nonexistent `Todo` section.


placer14:
@StrikerRUS, would you mind reaching out to us in slack or via email? We'd like to send a small token of appreciation but aren't sure where to send it. You can find my [email here](https://gist.github.com/placer14/b7bb0c3c0510af1cfda3c474cc923a3a).

Issue #949, opened 2018-05-17
Provide unit test for gencerts subcommand


placer14:
As a maintainer, I want to be able to ensure the `gencerts` subcommand performs the following behavior:
- produces valid and usable certificates
- generates the certificates in the expected location
- exercises interesting use cases for arguments and their side effects

Note: Follow-up work from https://github.com/OpenBazaar/openbazaar-go/issues/868

placer14:
This is resolved as part of https://github.com/OpenBazaar/openbazaar-go/pull/1602

PR #1241, opened 2018-10-20
[WIP] Start of unit test for gencerts subcommand:

allen-munsch:
#949 

 - WIP for testing: produces valid and usable certificates
 - generates the certificates in the expected location

However I was confused by `args` : https://github.com/OpenBazaar/openbazaar-go/blob/3df2e661e2d5e13064f6838c11f5611d4b04eeb4/cmd/gencerts.go#L27

I didn't understand how args was being used, sort of looked like it wasn't being used.

As far as interesting use cases, i'm not sure how interesting the current settings are.

It'd be awesome if i could get pointed to some use cases that I could test for, if they exist somewhere.


cpacia:
Thanks for this. I think the number one use case is for getting the electron UI client to make a valid ssl connection to the server. 

The trouble we had was programatically generating a cert that electron would accept. I believe the current code does that but not 100% sure. If we could even some how import it into electron rather than the OS level that would be OK too. 

But remote connecting client to server is the use case.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		pub, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			t.Fatal("failed to parse DER encoded public key: " + err.Error())
+		}
+		if pub.Issuer.String() != "O=OpenBazaar" {
+			t.Fatalf("%s != OpenBazaar", pub.Issuer.String())
+		}
+		// t.Log(pub.IssuingCertificateURL)
```
placer14: If you're using x509 here, you might consider adding `ValidHostname` to check the host here.
You can also use fields `NotBefore` and `NotAfter` to test duration. That's all I think I'd want in this unit test. You might also manually test the cert works in the use case @cpacia describes above and comment in the PR whether it worked.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"testing"
+
+	"github.com/OpenBazaar/openbazaar-go/cmd"
+)
+
+func TestGenCertsGenericDefaults(t *testing.T) {
+
+	if dataDir, err := ioutil.TempDir("", "gencerts_test"); err != nil {
+		t.Fatal(err)
+	} else {
```
placer14: You don't need the `else` case here. `t.Fatal` will exit the function as soon as it finishes.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"os"
+	"path"
+	"testing"
+
+	"github.com/OpenBazaar/openbazaar-go/cmd"
+)
+
+func TestGenCertsGenericDefaults(t *testing.T) {
+
+	if dataDir, err := ioutil.TempDir("", "gencerts_test"); err != nil {
```
placer14: We have a schema manager which helps with this stuff already. Take a look at https://github.com/OpenBazaar/openbazaar-go/blob/master/repo/db/chat_test.go#L13-L20 and replicate the pattern there.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,89 @@
+package cmd_test
```
placer14: I like that you scoped it to `_test`.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,89 @@
...
+		if block == nil {
+			t.Fatal("failed to parse PEM block containing the public key")
+		}
+
+		pub, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			t.Fatal("failed to parse DER encoded public key: " + err.Error())
+		}
+		if pub.Issuer.String() != "O=OpenBazaar" {
+			t.Fatalf("%s != OpenBazaar", pub.Issuer.String())
```
placer14: You should use `t.Errorf` when test conditions are not actually fatal to the rest of the unit test. Error and Log will allow the test to continue running so you can capture more than one failure per run instead of solving them one-at-a-time.

allen-munsch:
@placer14 thank you for the feedback I think I made the suggested changes.

As far as certificate import this is what i've found so far, however i'm not sure if the nssdb method is outdated, or if its cross platform.

  **win32/64**
```shell
  certutil.exe -addstore -user Root openbazaard.crt
```
- https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/certutil#BKMK_Store

**darwin**
```shell
  security add-trusted-cert -p ssl openbazaard.crt
```
  **linux**
```shell
  certutil -d sql:$HOME/.pki/nssdb -A -t "C,," -n 127.0.0.1 -i openbazaard.pem
```

A rough sketch of ssl configuration through electron, but only with linux support.
 - https://github.com/electron/electron/blob/master/docs/api/app.md#appimportcertificateoptions-callback-linux
```javascript
  if (process.platform === 'linux') {
    let sslcert = {
                   certificate: `${config.sslcert.path}`,
                   password: `${config.sslcert.password}`
                  }
    app.importCertificate(sslcert, function(result) {
      if (result != 0) {
        // https://cs.chromium.org/chromium/src/net/base/net_error_list.h
        dialog.showErrorBox('Error importing ssl certificate',
          `Error number: ${}\nCertificate path: ${config.sslcert.path}`)
      }
    })
  })
```

I'm still looking into how openbazaar-go and openbazaar-desktop do ssl. To better understand the test case.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	certSchema := schema.MustNewCustomSchemaManager(ctx)
+	if err := certSchema.BuildSchemaDirectories(); err != nil {
+		return nil, nil, err
+	}
+	return &ctx.DataPath, certSchema.DestroySchemaDirectories, nil
+}
+
+func TestGenCertsGenericDefaults(t *testing.T) {
+	var dataPath, Destroy, schema_err = buildCertDirectory()
```
placer14: Typically, we use capitalized names to indicate which variables are exported publicly from the package. Variables cannot be exported out of the function scope so we try to keep all function-scoped variable lowercase to not confuse readers. Idomatic go also prefers camelcase over snakecase.

Please correct:
`var datapath, destroy, schemaErr = buildCertDirectory()`

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	config := cmd.GenerateCertificates{
+		DataDir:  dataDir,
+		Testnet:  true,
+		Host:     "127.0.0.1",
+		ValidFor: 1, // 1ns ... 1*1e9 == 1s
+	}
+
+	args := []string{"args?"}
```
placer14: FYI, Execute satisfies the Commander interface defined here: https://github.com/jessevdk/go-flags/blob/c0bdd90e3f6f784046c4b384a3e78770aaa31b07/command.go#L41

It would have the specific args passed in from the command line.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+	return &ctx.DataPath, certSchema.DestroySchemaDirectories, nil
+}
+
+func TestGenCertsGenericDefaults(t *testing.T) {
+	var dataPath, Destroy, schema_err = buildCertDirectory()
+	dataDir := string([]byte(*dataPath))
+	if schema_err != nil {
+		Destroy()
+		t.Fatal(schema_err)
+	}
```
placer14: There's a way to make sure destroy() is included in every exit from this function. It's called `defer`. Once you know destroy is safe to call, you can include ` defer destroy()` and it will be executed just before the function exits.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+	} else {
+		if fileInfo.Mode().Perm() != 0755 {
+			Destroy()
+			t.Fatal("ssl directory does not have 0755 permissions")
+		}
+		if !fileInfo.IsDir() {
+			Destroy()
+			t.Fatalf("Expecting a directory: %s", dataDir)
+		}
+	}
```
placer14: Remember that `Fatal` stops the test. Because the if would stop the test as part of the call to `Fatal` we don't need the else here as it's implied that those two checks will be performed if `err != nil` is false in the check above. I realize the `else` extends the scope of the `fileInfo` and `err` variables, but it's preferred to have logic that is "not nested". Rewriting that would look like:
```
fileInfo, err := os.Stat(sslPath)
if err != nil {
	t.Fatal(err)
}
if fileInfo.Mode().Perm() != 0755 {
	t.Fatal("ssl directory does not have 0755 permissions")
}
if !fileInfo.IsDir() {
	t.Fatalf("Expecting a directory: %s", dataDir)
}
```

(Note: the Destroy() is gone because we used `defer Destroy()` as described above.)

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+	} else {
+		if fileInfo.Mode().Perm() != 0644 {
+			Destroy()
+			t.Fatal("cert.pem does not have 0644 permissions")
+		}
+		if !fileInfo.Mode().IsRegular() {
+			Destroy()
+			t.Fatalf("Expecting a file: %s", certPemPath)
+		}
+	}
```
placer14: Flatten please and the same with the following tests.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+		t.Fatal("failed to parse PEM block containing the public key")
+	}
+
+	pub, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		Destroy()
+		t.Fatal("failed to parse DER encoded public key: " + err.Error())
+	}
+	if pub.Issuer.String() != "O=OpenBazaar" {
+		t.Errorf("%s != OpenBazaar", pub.Issuer.String())
```
placer14: Descriptive test failures are important. `unexpected issuer found on certificate: %s != OpenBazaar`

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	// t.Log(pub.IssuingCertificateURL)
+	if err := pub.VerifyHostname("127.0.0.1"); err != nil {
+		t.Errorf("%s", err)
```
placer14: Descriptive test failure: `certificate could not verify local hostname: %s`

cmd/gencerts_test.go
```diff
@@ -0,0 +1,122 @@
...
+		if !fileInfo.Mode().IsRegular() {
+			Destroy()
+			t.Fatalf("Expecting a file: %s", keyPemPath)
+		}
+	}
+
+	certPemBytes, err := ioutil.ReadFile(certPemPath)
+	if err != nil {
+		Destroy()
+		t.Fatal(err)
```
placer14: Descriptive test failure: `unable to read generated certificate`

allen-munsch:
@placer14 thank you very much for the review. I've been super busy, but i'm hoping to get to these changes soon. Just a personal factor, but i'm on a low bandwidth connection, like 20Gb allotment, and it get's eaten up pretty quick using openbazaar. recently switched machines and dependency downloads with 70kb connection are difficult, so i've been waiting. lol.

placer14:
@allen-munsch I'm really sorry to hear that. Unfortunately, our dependency management is still using `godep` and is not kind on the bandwidth.

We've had a few people complaining about SSL not working properly. Seeing that you're already familiar with this area of the code, maybe you'd like to take a look at it? You could use your current versions to test/explore as well. https://github.com/OpenBazaar/openbazaar-go/issues/1307

cpacia:
Reopening as #1602 in an ob-go branch

PR #1246, opened 2018-10-20
CLEANUP: Fix misspellings.

tyler-smith:
Fix the last of the misspellings.


net/service/messagesender.go
```diff
@@ -104,7 +104,7 @@ func (ms *messageSender) prep() error {
 
 // streamReuseTries is the number of times we will try to reuse a stream to a
 // given peer before giving up and reverting to the old one-message-per-stream
-// behaviour.
```
placer14: I guess we're not using the Queen Mum's English? :P

PR #1258, opened 2018-10-25
Remove unused Python modules in qa/chat.py

gy741:
Hello,

In my opinion, No time, OrderedDict Python module are required.

Because it is not used.

What do you think of it?

Thanks.


cpacia:
I think you're correct. They are used in the subclasses but they are imported there as well. Thanks!

Issue #1220, opened 2018-10-10
Duplicate currencies when saving via /ob/profile


placer14:
When saving the profile, additional currencies are appended to the profile. The action should not duplicate the items and should instead be idempotent.

Example request:
`curl 'http://0.0.0.0:4002/ob/profile/QmVVdnd8vyUfrmWwFqytr9PYQyxasRQAoG9Azp1B728ug5' -X PUT -H 'Content-Type: application/json' --data-binary '{"about":"","handle":"","location":"","moderator":false,"name":"A change #2","nsfw":false,"shortDescription":"","vendor":true,"colors":{"primary":"#FFFFFF","secondary":"#ECEEF2","text":"#252525","highlight":"#2BAD23","highlightText":"#252525"},"contactInfo":{"website":"","email":"","phoneNumber":"","social":[]},"peerID":"QmVVdnd8vyUfrmWwFqytr9PYQyxasRQAoG9Azp1B728ug5","bitcoinPubkey":"0239e7c045191195324c4c96909c3c15aac92571c6110cc9a1546a9ff017414bd2","currencies":["BTC","BCH","ZEC","LTC"]}' --compressed`

Example response:
```
{
    "peerID": "QmVVdnd8vyUfrmWwFqytr9PYQyxasRQAoG9Azp1B728ug5",
    "handle": "",
    "name": "A change #2",
    "location": "",
    "about": "",
    "shortDescription": "",
    "nsfw": false,
    "vendor": true,
    "moderator": false,
    "contactInfo": {
        "website": "",
        "email": "",
        "phoneNumber": "",
        "social": [
        ]
    },
    "colors": {
        "primary": "#FFFFFF",
        "secondary": "#ECEEF2",
        "text": "#252525",
        "highlight": "#2BAD23",
        "highlightText": "#252525"
    },
    "bitcoinPubkey": "0239e7c045191195324c4c96909c3c15aac92571c6110cc9a1546a9ff017414bd2",
    "lastModified": "2018-10-10T13:26:38.536429800Z",
    "currencies": [
        "BTC",
        "BCH",
        "ZEC",
        "LTC",
        "ZEC",
        "LTC",
        "BTC",
        "BCH"
    ]
}
```

jjeffryes:
The currency codes are also duplicated in profile.moderatorInfo. For example, my test node returns this in it's profile when it's a moderator:

```
"moderatorInfo": {
                "acceptedCurrencies": [
                    "BTC",
[...]

PR #1259, opened 2018-10-25
[#1220] Currencies in profile no longer duplicate on update

placer14:
`core.UpdateProfile` was causing currency lists in the profile to become duplicated. Manually verified this fix works for the case illustrated in #1220.


cpacia:
lgtm

PR #1264, opened 2018-10-27
Add nil check to addresses in transaction listener

### LINKED ISSUE #1112: transaction_listener.go should should check for nil linkedAddress in input
The `input.linkedAddress` may be nil if the send sent from an exotic script and the wallet was unable to parse it. There should be a nil check on the input address which `continue`s at the start of the loop to avoid a panic.

aslafy-z:
Fixes #1112

Let me know if it needs more, I followed instructions from the issue @cpacia.


cpacia:
LGTM Thanks!

Issue #1004, opened 2018-07-05
Isolate `repo/db` tests


placer14:
Our current tests mutate a single pre-existing DB which make debugging core DB tests difficult. Ensuring that each test is isolated from another ensures all dependencies of the test are the only things present and under test.

This is an issue to track progress of walking through the `repo/db` unit tests and ensure they are all using their own instantiated objects and moving the test to the `db_test` package.

- [ ] cases_test
- [ ] chat_test
- [ ] coupons_test
- [ ] db_test
- [ ] followers_test
- [ ] following_test
- [ ] inventory_test
- [ ] keys_test
- [ ] moderatedstores_test
- [ ] notifications_test
- [ ] offlinemessages_test
- [ ] pointers_test
- [ ] purchases_test
- [ ] query_test
- [ ] sales_test
- [ ] settings_test
- [ ] stxo_test
- [ ] txmetadata_test
- [ ] txns_test
- [ ] utxos_test
- [ ] watched_scripts_test

PR #1265, opened 2018-10-28
[#1004] Isolate dependencies in repo/db/utxos_test.go

frodeaa:


gubatron:
can't this be refactored into a function? so much copy/paste.
```
	var uxdb, teardown, err = buildNewUnspentTransactionOutputStore()
	if err != nil {
		t.Fatal(err)
	}
	defer teardown()
	utxo := mustNewUtxo()
	err = uxdb.Put(utxo)
```

gubatron:
sweet! tnx for the refactor

repo/db/utxos_test.go
```diff
@@ -32,11 +26,39 @@ func init() {
 	}
 }
```
placer14: This wants to be a factory that lives in `test/factory/utxo.go`. Example: https://github.com/OpenBazaar/openbazaar-go/blob/master/test/factory/sale_record.go

placer14: Completed in https://github.com/OpenBazaar/openbazaar-go/pull/1265/commits/d2a28fdc497decc195647cc38031e38b381116d1

PR #1273, opened 2018-11-02
Use orderID in transaction listener

amangale:
@cpacia - This is a follow-up PR to the wallet interface changes.

In the transaction listener callback, I have made changes to the fetching of the order/contract details.
When orderID is available in the TransactionInput/TransactionOutput, it will be used to fetch the details else the existing logic of using the output.Address/input.LinkedAddress will be used.


cpacia:
LGTM thanks @amangale

PR #1277, opened 2018-11-05
Retry failed STORE against push nodes

### LINKED ISSUE #1276: Errors storing blocks with push nodes abort initial publishing
Found that when ob-go starts, if there are any errors publishing blocks to the push nodes, they will fail and not be re-attempted. Due to the push nodes not successfully connecting, this also degrades the experience slightly when attempting to connect to nodes which cannot be reached directly because push nodes (which cache this content for others) is not added to the local peer table and p2p queries end up taking longer than usual.

Suspect it also was a root cause of [this issue](https://github.com/OpenBazaar/openbazaar-go/issues/862).

placer14:
Fixes #1276. Retry storing initial update with push node via exponential backoff and a little bit of logging.


hoffmabc:
How much testing have you done on this? Should I help out?

PR #1290, opened 2018-11-08
TWEAK: Update the default config Insight servers for BTC and LTC.

tyler-smith:
Make config use openbazaar.org Insight hosts by default.


placer14:
@tyler-smith I assume we should wait until the Blockbook research shakes out before merging this?

tyler-smith:
Yeah. I made this PR right before we started investigating BB. I'll just reopen it later.

PR #1299, opened 2018-11-16
WIP: Refactor start command

hoffmabc:


placer14:
Reopening as merging `multiwallet` (which this PR was targeted at) was resolved without this PR.

cmd/start.go
```diff
@@ -160,12 +160,7 @@ func (x *Start) Execute(args []string) error {
 	}
 
 	// Create user-agent file
-	userAgentBytes := []byte(core.USERAGENT + x.UserAgent)
-	err = ioutil.WriteFile(path.Join(repoPath, "root", "user_agent"), userAgentBytes, os.ModePerm)
-	if err != nil {
-		log.Error("write user_agent:", err)
-		return err
-	}
+	createUserAgentFile(repoPath, x.UserAgent)
```
placer14: This needs to check the returned err and bubble it up stream.

cmd/start.go
```diff
@@ -110,6 +110,18 @@ type Start struct {
 	ZCash                string   `long:"zcash" description:"use a ZCash wallet in a dedicated data directory. To use this you must pass in the location of the zcashd binary."`
 }
 
+var (
+	mainConfig			[]byte
```
placer14: `mainConfig` is actually data and unusable as a config in this state. Want to keep var names as accurate as possible with something like `configData` or similar.

cmd/start.go
```diff
@@ -737,13 +698,80 @@ func (x *Start) Execute(args []string) error {
...
+		if err != nil && err != repo.ErrRepoExists {
+			return err
+		}
+		if db.Config().IsEncrypted() {
+			log.Error("Invalid password")
+			os.Exit(3)
+		}
+	}
+	return nil
+}
```
placer14: It would be good for `ensureDatabaseAccess` checked when the db is not encrypted as well. Otherwise, this is just `ensureEncryptedDatabaseAccess`.

cmd/start.go
```diff
@@ -737,13 +698,80 @@ func (x *Start) Execute(args []string) error {
...
+		log.Error("scan republish interval config:", err)
+		return err
+	}
+	walletsConfig, err = schema.GetWalletsConfig(mainConfig)
+	if err != nil {
+		log.Error("scan wallets config:", err)
+		return err
+	}
+	return nil
+}
```
placer14: This is a nice step. Next would be to formalize the config struct, remove the globals, and make `getConfigFiles` return an instantiated config. This removes mutation and global vars from the config handling as well as makes unit testing this possible.

cmd/start.go
```diff
@@ -698,9 +561,228 @@ func (x *Start) Execute(args []string) error {
...
+		nodeConfig.ExtraOpts["ipnsps"] = true
+	}
+
+	if onionTransport != nil {
+		nodeConfig.Host = defaultHostOption
+	}
+	return nodeConfig
+}
+
+func defaultHostOption(ctx context.Context, id peer.ID, ps pstore.Peerstore, bwr metrics.Reporter, fs []*net.IPNet, tpt smux.Transport, protec ipnet.Protector, opts *ipfscore.ConstructPeerHostOpts) (p2phost.Host, error) {
```
placer14: This long method signature seems ready for args to be flattened into a config struct.

cmd/start.go
```diff
@@ -698,9 +561,228 @@ func (x *Start) Execute(args []string) error {
...
+	if password != "" {
+		torPw = password
+	}
+	onionTransport, err = oniontp.NewOnionTransport("tcp4", torControl, torPw, nil, repoPath, (usingTor && usingClearnet))
+	if err != nil {
+		log.Error("setup tor transport:", err)
+		return err
+	}
+	return nil
+}
```
placer14: This method mutates the node and would better to pass in the mutated objects as args instead of pulling them out of the global vars. Limits mutation to this function and will make it testable later.

cmd/start.go
```diff
@@ -698,9 +561,228 @@ func (x *Start) Execute(args []string) error {
...
+
+	cfg, err := nodeRepo.Config()
+	cfg.Bootstrap = testnetBootstrapAddrs
+	dht.ProtocolDHT = "/openbazaar/kad/testnet/1.0.0"
+	bitswap.ProtocolBitswap = "/openbazaar/bitswap/testnet/1.1.0"
+	service.ProtocolOpenBazaar = "/openbazaar/app/testnet/1.0.0"
+
+	dataSharingConfig.PushTo = []string{}
+	return nil
+}
```
placer14: Why not make a function which handles setting up peers on both mainnet and testnet?

Also, that this mutates the configs suggest they should be passed in as a param instead of grabbing from the global vars.

PR #1300, opened 2018-11-19
Add referenceID to spend

amangale:
@cpacia - Changed the Spend calls in the jsonapi and refunds.
In jsonapi, the send struct will require memo to contain the referenceID/orderID to be passed from the invoker.


placer14:
Reopening as merging `multiwallet` (which this PR was targeted at) was resolved without this PR.

PR #1317, opened 2018-11-29
Update socketio dep

cpacia:


cpacia:
This was tyler's change which I have reviewed so I'm going to merge.

PR #1322, opened 2018-11-30
Update Docs

Quaternioneer:
I got a math.round error from 1.9. I skipped 1.10 and went to 1.11, but a post I read said that math.round was added in 1.10 not 1.11.  Presumably 1.10 is sufficient.

This isn't quite atomic. I am quite new to git.  I do not know how I would PR for each file individually.


placer14:
Had some additional updates. Closing in favor of https://github.com/OpenBazaar/openbazaar-go/pull/1364.

Issue #1319, opened 2018-11-29
Add POST /ob/orderspend


placer14:
In order to differentiate the intent of a client wanting to send money from the wallet in payment of an order (as part of completing checkout) or simply out of the wallet for other means, an API endpoint will be available for each case.

- `/ob/orderspend` will be created and require the same arguments + `orderID string` 
- Requests to `/ob/orderspend` missing `orderID` will respond with a `400 Bad Request`
- Valid requests can flow through to the existing behavior

```
POST /ob/spend
{
	"wallet": "BTC",
	"address": "14M5mXu2Z4tiTnQrRL7xR7PjBXbm4eBTN1",
	"amount": 100000,
	"feeLevel": "NORMAL",
	"memo": "Yo"
}
```

```
POST /ob/orderspend
{
	"wallet": "BTC",
	"address": "14M5mXu2Z4tiTnQrRL7xR7PjBXbm4eBTN1",
	"amount": 100000,
	"feeLevel": "NORMAL",
	"memo": "Yo",
	"orderID": "QmYepkN7bAgqLiSa1Nu3cnVUX6bQQNwgHyc7vCf",
}
```

placer14:
> Adjust POST /ob/spend API argument to orderID instead of memo
@rmisio Not sure where this came from. I said: 
> Instead of overloading the usage of this argument, we should create a new argument to this endpoint called orderID.

This would be the latter suggestion you're asking for.

placer14:
After discussion, we will not overload the existing API endpoint and create order-specific endpoint. Updated the description of the issue to reflect the new plans.

rmisio:
What about the use case of a user copying the payment address and going into the wallet and making a payment for the order?

Before, with the unified spend call, that would work. Now, I guess it will still work, but there won't be any orderId coming over.

placer14:
We would probably have to remove the ability for someone to copy the payment address to be used outside the wallet in the Ethereum case because it would not complete. Otherwise, external wallet payments would work as normal.

PR #1325, opened 2018-11-30
[#1319] Separate spend and spendorder behavior

placer14:
This change adds a new possible error to `POST /ob/spendorder`. Our error handling is a little all over the place so it simply returns a string "ERROR_ORDER_NOT_FOUND" if the param is empty or if the order cannot be retrieved. Hopefully we can round back on to the error handling at some point.

### DIFF
--- a/core/currency.go
+++ b/core/currency.go
@@ -1,5 +1,141 @@
 package core
 
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/OpenBazaar/openbazaar-go/pb"
+	"github.com/OpenBazaar/openbazaar-go/repo"
+	wallet "github.com/OpenBazaar/wallet-interface"
+)
+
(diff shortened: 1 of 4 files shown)


core/currency.go
```diff
@@ -1,5 +1,141 @@
...
+}
+
+func (n *OpenBazaarNode) getOrderContractBySpendRequest(args *SpendRequest) (*pb.RicardianContract, error) {
+	if args.OrderID != "" {
+		contract, _, _, _, _, err := n.Datastore.Purchases().GetByOrderId(args.OrderID)
+		if err == nil && contract != nil {
+			return contract, nil
+		}
+	}
+	contract, _, _, _, _, err := n.Datastore.Purchases().GetByOrderId(args.Address)
```
amangale: GetOrderById should be GetByPaymentAddress

placer14: Thanks, meant to check both ways and didn't include. Will fix.

PR #1345, opened 2018-12-14
Sign Verify Content with IPFS Key

hoffmabc:
Per @drwasho's request I whipped up a quick PR to kick this request off, but please be aware this needs conversation and I am not positive it even meets the requirements. 

**/ob/signmessage** -
Input: `{ "content": "<string_content>" }`
Output: `{ "signature": "<hex_encoded_sig>, "pubkey": "<hex_encoded_pubkey>", "peerId": "<peer_id>" }`

**/ob/verifymessage** -
Input: `{ "content": "<string_content>", "signature": "<hex_encoded_sig>", "pubkey": "<hex_encoded_pubkey>", "peerId": "<peerId>" }`
Output: `{ "error": "<empty, PEER_ID_PUBKEY_MISMATCH, or VERIFICATION_FAILED>", "peerId": "<submitted_peerId>" }`

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -8,6 +8,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"gx/ipfs/QmPvyPwuCgJ7pDmrKDxRtsScJgBaM5h4EpRL2qQJsmXf4n/go-libp2p-crypto"
 
 	"gx/ipfs/QmPSQnBKM9g7BaUcZCvswUJVscQ1ipjmwxN5PXCjkp9EQ7/go-cid"
 	mh "gx/ipfs/QmPnFwZ2JXKnXgMw8CdBPxn7FWh6LLdjUjxV1fKHuJnkr8/go-multihash"
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -2261,6 +2262,27 @@ func (i *jsonAPIHandler) POSTReleaseEscrow(w http.ResponseWriter, r *http.Reques
 	}
 
 	SanitizedResponse(w, `{}`)
+}
+
+func (i *jsonAPIHandler) POSTSignMessage(w http.ResponseWriter, r *http.Request) {
+	type ipfsmessage struct {
+		Content string `json:"content"`
+	}
+	var msg ipfsmessage
+	decoder := json.NewDecoder(r.Body)
+	err := decoder.Decode(&msg)
+	if err != nil {
+		ErrorResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	sig, err := i.node.IpfsNode.PrivateKey.Sign([]byte(msg.Content))
+	if err != nil {
+		ErrorResponse(w, http.StatusBadRequest, "there was an error signing the submitted content")
+		return
+	}
+
+	SanitizedResponse(w, fmt.Sprintf(`{"signature": "%s"}`, base64.URLEncoding.EncodeToString(sig)))
 }
 
 func (i *jsonAPIHandler) POSTChat(w http.ResponseWriter, r *http.Request) {
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/sha256"
-	"encoding/base64"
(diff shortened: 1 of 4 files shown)


drwasho:
@hoffmabc with `POST /ob/verifymessage`, how does the user know that a pubkey corresponds to a specific peerId?

hoffmabc:
A Peer ID can be generated from the pubkey.

drwasho:
I guess my point is that there's no way for a typical user to determine that; can the response in `verifymessage` include a peerId generated from the pubkey?

tyler-smith:
@drwasho it's technically possible but the verifier calculating the peer id from the pub key them-self is a critical part of the security. If they don't then they don't know that the message was really signed by that identity.

drwasho:
Most users won't know how to do that. It would be better if the `POST /ob/verifymessage` endpoint takes the `peerId` as an input and verifies:

1. **The `pubkey` matches the `peerId`.** This will actively prevent an attacker trying to attribute a correctly signed message to another `peerId`, rather than relying on the user to verify this themselves somehow.
2. **The signed message matches the `pubkey`**. That the message is cryptographically valid.

tyler-smith:
I deleted my comment because I realized I misunderstood you. I thought I did it fast enough you wouldn't see it.

Your suggestion to have verifymessage return the peerID would work. The user's ob-go server is acting as the verifier so it should be calculating the peer id. I do like the idea of making the peerID an input as well and then the response just needs to be a yes or no.

drwasho:
No problem. The cherry on top (sorry @hoffmabc) would be to have error responses for each failure scenario:

1. Pubkey failed to match peerId
2. Bad signature

drwasho:
LGTM

placer14:
@hoffmabc I was trying to test if they were inverse functions of each other and they appear to not be. Am I using this correctly? Or is there something wrong here?

```
$ curl user:password@0.0.0.0:14002/ob/signmessage -d '{"content":"something"}'

{
    "peerId": "QmZL2xN6N7DbWkiMp1kQS4g4AVXnH8Cwwbx5TmnxgTdqPU",
    "pubkey": "08011260b87a9435fdec44c9adfe3fada61d211b981aedc2fb76ac05e2cc35b390cb14cad8427e9cf86b5d47e2d19fe7f010e966980b41d2a36596d758706c8adff54cd5d8427e9cf86b5d47e2d19fe7f010e966980b41d2a36596d758706c8adff54cd5",
    "signature": "5bf5ae946b1a03cb6f931d12a28f4ced4641f9fed1c50c1f84e2416fe65e28230437f26d10adadb2d4c377a9f037775da6a6e0e06089b04f1f203f6acbd3860f"
}

$ curl user:password@0.0.0.0:14002/ob/verifymessage -d '{"content":"something","signature":"5bf5ae946b1a03cb6f931d12a28f4ced4641f9fed1c50c1f84e2416fe65e28230437f26d10adadb2d4c377a9f037775da6a6e0e06089b04f1f203f6acbd3860f", "pubkey":"08011260b87a9435fdec44c9adfe3fada61d211b981aedc2fb76ac05e2cc35b390cb14cad8427e9cf86b5d47e2d19fe7f010e966980b41d2a36596d758706c8adff54cd5d8427e9cf86b5d47e2d19fe7f010e966980b41d2a36596d758706c8adff54cd5", "peerId":"QmZL2xN6N7DbWkiMp1kQS4g4AVXnH8Cwwbx5TmnxgTdqPU"}'

{
    "error": "PEER_ID_PUBKEY_MISMATCH"
}
```

placer14:
Also, if there's a unit test we can add which proves the endpoints are actually inverse behaviors of each other, we should add that. Something which takes the output of `/ob/signmessage` and verifies it with `/ob/verifymessage`.

api/jsonapi.go
```diff
@@ -2223,6 +2224,92 @@ func (i *jsonAPIHandler) POSTReleaseEscrow(w http.ResponseWriter, r *http.Reques
...
+
+	keyBytes, err := i.node.IpfsNode.PrivateKey.Bytes()
+	if err != nil {
+		ErrorResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	SanitizedResponse(w, fmt.Sprintf(`{"signature": "%s","pubkey":"%s","peerId":"%s"}`,
+		hex.EncodeToString(sig),
+		hex.EncodeToString(keyBytes),
```
tyler-smith: This is returning the private key and not the public key.

hoffmabc: Well it seems fix but unfortunately it's not converting the privkey to pubkey properly.

hoffmabc:
This is ready for review @placer14 @cpacia

hoffmabc:
@placer14 @cpacia are we looking good to merge?

Issue #1183, opened 2018-09-19
Track API endpoint for each coin in CoinConfig


placer14:
Each coin should have a list of insight-compatible endpoints it should communicate with.

Requirements:
- The list will later be used for roundrobin failure so order should be significant.
- The endpoints should be captured as well-formed strings which are validated on startup by `url.Parse`
- The endpoint should be persisted in the current flatfile config for user editing purposes

To-Do:
- [x] Convert default config to produce a list of API endpoints for both mainnet and testnet for each coin (BTC, BCH, LTC, ZEC, and ETH)
- [x] Migrate existing configurations to use the provided lists of endpoints for each coin

API endpoint lists:

*BTC*
https://btc.api.openbazaar.org/api
*BCH*
https://bch.api.openbazaar.org/api
*LTC*
https://ltc.api.openbazaar.org/api
*ZEC*
https://zec.api.openbazaar.org/api
*ETH*
https://rinkeby.infura.io

*TBTC*
https://tbtc.api.openbazaar.org/api
*TBCH*
https://tbch.api.openbazaar.org/api
*TLTC*
https://tltc.api.openbazaar.org/api
*TZEC*
https://tzec.api.openbazaar.org/api
*TETH*
https://rinkeby.infura.io

placer14:
Blockbook endpoints to date:
https://btc.blockbook.api.openbazaar.org/
https://ltc.blockbook.api.openbazaar.org/
https://zec.blockbook.api.openbazaar.org/
https://bch.blockbook.api.openbazaar.org/

Insight endpoints to date:
https://btc.insight.api.openbazaar.org/
https://ltc.insight.api.openbazaar.org/
https://zec.insight.api.openbazaar.org/
https://bch.insight.api.openbazaar.org/

Can you confirm that these are the tentative endpoints for each type, @tyler-smith ?

cpacia:
I believe this is done.

PR #1347, opened 2018-12-14
[#1183] Migrate multiwallet config; Refactor multiwallet config handling

placer14:
- Reconcile schema.DefaultWalletsConfig()
- Migrate legacy "Wallet" config key to "LegacyWallet" which is ignored
by config handling
- Migrate default "Wallets" config key to contain default multiwallet
settings. Ignores prior fee defaults, but legacy config is intact should
the user decide to restore their prior settings.
- A failure to create a single wallet will no longer impact other
wallets
- Removal of the coin key under the "Wallets" key will disable that
wallet. Alternatively, changing the Type to a value other than "API" and
"SPV" will also disable that coin.

### DIFF
--- a/repo/migrations/Migration015.go
+++ b/repo/migrations/Migration015.go
@@ -53,7 +53,7 @@ func migration015DefaultWalletConfig() *migration015WalletsConfig {
 			Type:             "API",
 			APIPool:          []string{"https://bch.api.openbazaar.org/api"},
 			APITestnetPool:   []string{"https://tbch.api.openbazaar.org/api"},
-			FeeAPI:           feeAPI,
+			FeeAPI:           "", // intentionally blank
 			LowFeeDefault:    1,
 			MediumFeeDefault: 5,
 			HighFeeDefault:   10,
@@ -64,7 +64,7 @@ func migration015DefaultWalletConfig() *migration015WalletsConfig {
 			Type:             "API",
 			APIPool:          []string{"https://ltc.api.openbazaar.org/api"},
 			APITestnetPool:   []string{"https://tltc.api.openbazaar.org/api"},
-			FeeAPI:           feeAPI,
+			FeeAPI:           "", // intentionally blank
 			LowFeeDefault:    5,
 			MediumFeeDefault: 10,
 			HighFeeDefault:   20,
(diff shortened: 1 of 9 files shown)


repo/migrations/Migration015.go
```diff
@@ -0,0 +1,77 @@
...
+		return fmt.Errorf("reading config: %s", err.Error())
+	}
+
+	if err = json.Unmarshal(configBytes, &configMap); err != nil {
+		return fmt.Errorf("unmarshal config: %s", err.Error())
+	}
+
+	configMap["LegacyWallet"] = configMap["Wallet"]
+	delete(configMap, "Wallet")
+	configMap["Wallets"] = schema.DefaultWalletsConfig()
```
placer14: I need to remove this reference and use a literal instead.

wallet/builder.go
```diff
@@ -2,247 +2,364 @@ package wallet
...
+		}
+	case wallet.BitcoinCash:
+		if notMainnet {
+			actualCoin = wallet.TestnetBitcoinCash
 		} else {
-			mw[wallet.BitcoinCash] = bitcoinCashSPVWallet
+			actualCoin = wallet.BitcoinCash
 		}
 	}
+	newSPVWallet, err := spvwallet.NewSPVWallet(coinConfig)
```
cpacia: This will need to use `bitcoincash.NewSPVWallet()` if it's a bitcoin cash wallet.

schema/configuration.go
```diff
@@ -57,6 +57,67 @@ type DataSharing struct {
...
+			MediumFeeDefault: 10,
+			HighFeeDefault:   50,
+			MaxFee:           200,
+			WalletOptions:    nil,
+		},
+		BCH: &CoinConfig{
+			Type:             WalletTypeAPI,
+			APIPool:          CoinPoolBCH,
+			APITestnetPool:   CoinPoolTBCH,
+			FeeAPI:           feeAPI,
```
cpacia: I think `feeAPI` for everything other than BTC should bet set to "". If it isn't the wallet may try to use that API.

repo/migrations/Migration015.go
```diff
@@ -0,0 +1,167 @@
...
+			MediumFeeDefault: 10,
+			HighFeeDefault:   50,
+			MaxFee:           200,
+			WalletOptions:    nil,
+		},
+		BCH: &migration015CoinConfig{
+			Type:             "API",
+			APIPool:          []string{"https://bch.api.openbazaar.org/api"},
+			APITestnetPool:   []string{"https://tbch.api.openbazaar.org/api"},
+			FeeAPI:           feeAPI,
```
cpacia: Same comment about the feeAPI as below

cpacia:
This LGTM

PR #1365, opened 2019-01-02
Update multiwallet dep with timestamp bug fix

cpacia:


cpacia:
Going to merge if no objections

PR #1378, opened 2019-01-07
Fix panic on exchange rate endpoint

cpacia:


placer14:
What's the difference between coinType and currencyCode? Any way you can make the reason for this change more clear? Like with a unit test or make the API return an error indicating the problem?

cpacia:
The currency code is fiat and the cointype is the wallet coin type. With this fix in place if they omit the cointype then it will return an "unknown cointype" error the same way as if they just entered garbage.

cpacia:
Updated with tests for this panic. Also added issue #1381 to figure out how to get mock data from exchange rate implementations so we can run more tests on this endpoint.

PR #1379, opened 2019-01-08
Fix issue: cannot create and save listing with full Chinese name

fengzie:
Cannot save listing if the name was full Chinese when creating listing.


cpacia:
Commits moved to #1464 and rebased

mf-00:
Thank you Chris. 
The following commits might not be required, it is just for incorrect commit trial. You could remove it.
screen shot 2019-02-26 at 8 49 11 am

Regards

fengzie:
Thank you Chris.
The following commits might not be required, it is just for incorrect commit trial. You could remove it.
screen shot 2019-02-26 at 8 49 11 am

Regards

PR #1380, opened 2019-01-08
Multiwallet lock fix

cpacia:


cpacia:
Tested this fix and confirmed it solves the blocking close issue. Also can confirm the socket remains functional as expected.

Issue #1388, opened 2019-01-10
Crash on re-broadcasting transaction


cpacia:
When a new block comes in it will try to re-broadcast transactions which have zero confirmations. This sometimes results in a panic when the broadcast API call returns an error.

```
goroutine 113579 [running]:
github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client.(*healthState).nextAvailable(0x0, 0x227c8f5013f, 0x27f09c0, 0xc0040c9ba0)
	/home/chris/workspace/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client/rotation_manager.go:40 +0x22
github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client.(*healthState).markUnhealthy(0x0)
	/home/chris/workspace/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client/rotation_manager.go:20 +0x51
github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client.(*rotationManager).FailCurrent(0xc0011b5440)
	/home/chris/workspace/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client/rotation_manager.go:146 +0x9a
github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client.(*ClientPool).FailAndCloseCurrentClient(0xc0004de780)
	/home/chris/workspace/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client/pool.go:127 +0x38
github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client.(*ClientPool).executeRequest(0xc0004de780, 0xc001b23b00, 0x20, 0xc0016976c0)
	/home/chris/workspace/go/src/github.com/OpenBazaar/openbazaar-go/vendor/github.com/OpenBazaar/multiwallet/client/pool.go:161 +0x99
[...]

PR #1389, opened 2019-01-10
Update multiwallet dep with healthstate nil check

cpacia:


placer14:
Expected to fix #1388

PR #1402, opened 2019-01-15
Provide currency to GETOrder

placer14:
This temporarily includes the `BuyerOrder.Payment.Coin` when missing until a proper contract migration can be executed.


placer14:
This is ready for review @cpacia

PR #1405, opened 2019-01-15
Revert order of operations change in disputes.go

cpacia:


jjeffryes:
@cpacia can you do this part so we don't forget why this was done?

>>> Please open an issue tracking the re-application of this revert. Please describe why the original ordering was harmful as well as why the new ordering didn't work so we have context to decide on how to repair this while it's still fresh.

PR #1414, opened 2019-01-19
Revert data directory name to preserve identity and all other data after upgrade to v13

pinheadmz:
After upgrading to server v13 I found I could not connect from my client any more, and even authenticated `curl` requests were failing, from both remote and local. After checking the server again I realized OB was not using any of my existing data! My onion key, API credentials, all were replaced with fresh new arguments in a brand new folder `~/.openbazaar`

This change was made here https://github.com/OpenBazaar/openbazaar-go/commit/0c9e934bac634a546efe3e7d0ef4f8d0e58cd328

Title of that PR is **"Change default data directory path for new node instances"** which makes sense, but in my case the migration to v13 never took place because the old data was not parsed.

I'm not sure how the v13 upgrade works for the bundled application, or what the motivation was to rename that data directory so I understand if this PR doesn't get merged, but just in case anyone else is running with separate server / client setups, they may run in to this problem after upgrade to v13 as well.

ping @placer14 thanks!

### DIFF
--- a/schema/path.go
+++ b/schema/path.go
@@ -39,9 +39,9 @@ func OpenbazaarPathTransform(basePath string, testModeEnabled bool) (path string
 }
 func directoryName(isTestnet bool) (directoryName string) {
 	if runtime.GOOS == "linux" {
-		directoryName = ".openbazaar"
+		directoryName = ".openbazaar2.0"
 	} else {
-		directoryName = "openbazaar"
+		directoryName = "openbazaar2.0"
 	}
 
 	if isTestnet {


placer14:
@pinheadmz The expectation was that bundled users which _were_ using the coin flags for the server to derive their installation path are now getting the `-d` flag passed explicitly with their old paths being kept intact. We also expected that stand-alone server setups would always be using `-d` and would allow the migration to run in-place. Stand-along server setups without using the `-d` flag would definitely have this problem, but we don't know if the `~/.openbazaar` location was using old v1 data (which is the path we used then) or if it's v2. The best solution that affected the least number of users was to not move your existing directory as part of a migration and allow legacy paths to continue to be used with the `-d` flag. I'm not sure if you can think of a better way to handle this, but I'm open to suggestions.

In your case, if you want to use the default path, I would just move your data directory to the new default and continue running the server without `-d`. Perhaps we should also create a support article to help users understand why their standalone server started with a new user and all of their old history being lost?

PR #1424, opened 2019-01-23
[WIP] Bulk update coupons [Relies on PR #1423]

hoffmabc:
This relies on PR #1423 getting merged first.

Here is the Bulk Update Coupons.

It accepts an array of coupon codes and replaces all coupon codes on all listings.

*POST* http://127.0.0.1:4002/ob/bulkupdatecoupons

Example Payload:
```
{"coupons": 
  [ 
    { "title": "test", "code": "test2",  "percentDiscount": 99 },
    { "title": "test", "code": "test23",  "percentDiscount": 32 } 
  ] 
}
```

### DIFF
--- a/core/moderation.go
+++ b/core/moderation.go
@@ -178,6 +178,74 @@ func (n *OpenBazaarNode) GetModeratorFee(transactionTotal uint64, paymentCoin, c
 	}
 }
 
+// SetCurrencyOnListings - set currencies accepted for a listing
+func (n *OpenBazaarNode) SetCurrencyOnListings(currencies []string) error {
+	absPath, err := filepath.Abs(path.Join(n.RepoPath, "root", "listings"))
+	if err != nil {
+		return err
+	}
+	hashes := make(map[string]string)
+	walkpath := func(p string, f os.FileInfo, err error) error {
+		if !f.IsDir() {
+			file, err := ioutil.ReadFile(p)
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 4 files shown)


core/moderation.go
```diff
@@ -178,6 +178,76 @@ func (n *OpenBazaarNode) GetModeratorFee(transactionTotal uint64, paymentCoin, c
...
+			if err != nil {
+				return err
+			}
+			sl := new(pb.SignedListing)
+			err = jsonpb.UnmarshalString(string(file), sl)
+			if err != nil {
+				return err
+			}
+
+			sl.Listing.Coupons = coupons
```
placer14: Should we remove the prior coupons when we apply the new ones? Merge them?

cpacia: I think it's propbably OK to replace the existing coupons as the intent is to set all the listings to the same thing. Or so it seems to me.

hoffmabc: That's what I was thinking, but either way we have to be clear about it and make it obvious to users what's going on.

cpacia:
Since this combines the bulk currency PR as well maybe change the title so as to reflect that this PR contains both.

hoffmabc:
I think that's why we had the moderators setting functionality set up in this way, but @placer14 asked to re-use the UpdateListing call. I'll defer to your best judgment on this.

cpacia:
Closing for now as this is a backlog item. Can revisit later

PR #1450, opened 2019-02-18
Override travis default install

placer14:
Travis changed it's default `install` behavior to automatically attempt restoring the dependencies defined by `Godep.json`. Since we vendor our dependencies to not require this step, we are replacing the default behavior with a no-op.


hoffmabc:
LGTM

PR #1455, opened 2019-02-20
Add withTimeout to Chat and Offline Messages 

### LINKED ISSUE #1224: API calls where messages are sent to offline stores exceeding intended timeouts
@placer14 [recently committed some code](https://github.com/OpenBazaar/openbazaar-go/commit/37a31e2e1b8654f55f087bfe7ca217bd3cfec026) to reduce the timeout period, for mobile servers, where the node will attempt to send an online message to a peer to 5 seconds. If the timeout is hit, an offline message will be sent via a background goroutine, and any API should return an immediate `200` response while the message is sent in the background.

I was testing this out by trying to send a `POST /ob/follow` API call to `QmTmCkNLUcPGvf3mSYDme4UQudgn9oCVqE13GHnrF6sjLj` (who I confirmed was offline). The call took `64833 ms` to return a response. This seems to indicate that the timeout isn't working as intended in mobile and perhaps desktop (where the timeout should be 30 seconds).

hoffmabc:
Helps fix #1224

### DIFF
--- a/core/net.go
+++ b/core/net.go
@@ -467,7 +467,7 @@ func (n *OpenBazaarNode) SendChat(peerID string, chatMessage *pb.Chat) error {
 	if err != nil {
 		return err
 	}
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), n.OfflineMessageFailoverTimeout)
 	defer cancel()
 	err = n.Service.SendMessage(ctx, p, &m)
 	if err != nil && chatMessage.Flag != pb.Chat_TYPING {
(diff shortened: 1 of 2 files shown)


hoffmabc:
cc: @drwasho 

@placer14 @cpacia please take a closer look at this as I added some more timeouts to offline messaging. The reason for this is that mobile is also having trouble with offline timeouts related to purchases. I've tested this locally on desktop and it seems to be performing as intended, but I'm not sure if you guys will find side effects this may impact.

Below are two purchase calls that are hitting the 5s limit (modified the start for desktop to use 5s) and falling through.

image

image

core/net.go
```diff
@@ -96,7 +96,7 @@ func (n *OpenBazaarNode) SendOfflineMessage(p peer.ID, k *libp2p.PubKey, m *pb.M
 	log.Debugf("Sending offline message to: %s, Message Type: %s, PointerID: %s, Location: %s", p.Pretty(), m.MessageType.String(), pointer.Cid.String(), pointer.Value.Addrs[0].String())
 	OfflineMessageWaitGroup.Add(2)
 	go func() {
-		ctx, cancel := context.WithCancel(context.Background())
+		ctx, cancel := context.WithTimeout(context.Background(), n.OfflineMessageFailoverTimeout)
```
cpacia: Since this is in a separate goroutine I don't think we need a timeout here.

hoffmabc: Are we good on the other ones?

cpacia: I think the other ones are OK. If you remove this one we can merge.

hoffmabc:
GTG @cpacia

Issue #1437, opened 2019-02-07
ETH precision: openbazaar-go updates


drwasho:
## `openbazaar-go` updates

- [ ] `repo.Listing` abstraction (for handling contract versioning properly)
- [ ] `(*repo.Listing) Quantity() big.Int` respects versions properly (w/ unit tests)
- [ ] `MarshalJSON` and `UnmarshalJSON`
- [ ] `protobuf` changes (updates to proto usage should use `repo.Listing` getters for accessing changed schema data)
- [ ] JSON API updates
- [ ] `openbazaar-go` datastore migrations

placer14:
Changed the spec to remove the `magnitude` key in favor of a simpler "one value" reprentation.
```
{
    "data": "value",
    "amount": { // this key would match the replaced key in the BEFORE case
        "value": "12345678",
      //  "magnitude": -8,    # omit this key
        "currency": {
            "code": "BTC", // this key would always be `code`
            "divisibility": 8
        }
    },
    "otherdata": "othervalue"
}
```

placer14:
@amangale Discussion has raised the following requirement:
> NOTE: We need to ensure values lower than `0` in the `value` field when used in the JSON API are rejected with an error explaining that the value must be a positive integer. Recommend a `400 Bad Request` error response.

jjeffryes:
@amangale you asked for some details of calls that use the amount field, here's what I turned up with a quick search that I know the desktop client uses. This may not include everything:

- ob/purchase
- wallet/spend
- wallet/orderspend
- wallet/balance
- ob/listing/[id]/[slug]
- ob/listing/ipfs/[hash]
- ob/listings
- ob/case/[id]
- ob/order/[id]
- ob/profile/[id] (in the moderator fee info)

rmisio:
@amangale 

Ashwan, just following up on my comment in the slack thread:

Keep in mind that a few of those endpoints include the full listing which does already have the coinDivisibility in it in the metadata object. It’s a required field when creating the listing.

So, for those api’s no change is needed.

I think these are the ones with the full listing:

- ob/listing/[id]/[slug]
- ob/listing/ipfs/[hash]
- ob/order/[id]

rmisio:
@amangale 

Another thing is the metadata.coinDivisibility in the listing is in the old format (e.g. 100000000 instead of 8. Can you please make sure it uses the new format?

[...]

PR #1465, opened 2019-02-26
Improve Exchange Rates

placer14:
Improve exchange rate handling to be more generic and self-contained. This work begins #1437 by adding structs which will implement the CurrencyValue and CurrencyDefinition interfaces. This uses those same structs to perform the currency conversion behavior and validate many currency usages.

Future work: 
- Move exchange rate handler out of (multiple) wallet implementations and into the CurrencyDictionary. Replace exchange rate interface in wallets with native currency rates as it relates to a set of reserve currencies (which will allow us to support beyond BTC as a reserve currency, if BTC is unavailable at some point later).


api/jsonapi.go
```diff
@@ -1517,8 +1517,10 @@ func (i *jsonAPIHandler) POSTOrderConfirmation(w http.ResponseWriter, r *http.Re
 	}
 
 	// TODO: Remove once broken contracts are migrated
-	if _, err := repo.NewCurrencyCode(contract.BuyerOrder.Payment.Coin); err != nil {
-		log.Warningf("missing contract BuyerOrder.Payment.Coin on order (%s)", conf.OrderID)
+	lookupCoin := contract.BuyerOrder.Payment.Coin
+	_, err = repo.LoadCurrencyDefinitions().Lookup(lookupCoin)
```
placer14: `NewCurrencyCode` had validation to ensure this value was not empty. Now, instead of only checking that it's valid, we check if it's a known coin by finding out if it's defined by the daemon.

core/order.go
```diff
@@ -1016,27 +1017,71 @@ func quantityForItem(version uint32, item *pb.Order_Item) uint64 {
...
+	}
+	resultValue, err := reserveValue.ConvertTo(paymentCurrencyDef, reserveIntoResultRate)
+	if err != nil {
+		return 0, fmt.Errorf("converting from reserve: %s", err.Error())
+	}
+	result, err := resultValue.AmountUint64()
+	if err != nil {
+		return 0, fmt.Errorf("unable to represent (%s) as uint64: %s", resultValue.String(), err.Error())
+	}
+	return result, nil
```
placer14: The new behavior for `getPriceInSatoshi(paymentCurrency, listedCurrency, amount)`:
1. Assume reserve currency is `BTC` (this can be flexible later)
2. Get reserve currency's rates vs. all other currencies.
3. Get rates of payment and listed as it relates to the reserve currency.
4. Take amount in `listingCurrency` and `ConvertTo` reserve currency. Then again to `paymentCurrency`. (this consider divisibility within the CurrencyValue as part of the conversion)
5. Attempt to cast it back into int64 (this would still fail for very large values, but it would be protected)

cpacia: I'd like to see a test for this function as it's pretty critical though it requires a good deal of mocking. Maybe this could be broken out into a separate task.

placer14: I attempted to protect the conversion of the value as much as possible without requiring a mocked API response or exchange rate provider. I think we should replace the exchange rate interface to provide something like:
```
type ExchangeRater interface{
  // returns the rate of native wallet currency to the reserve currencies
  // indentified by normalized currency code string as the key
  RatesByReserveCurrency() map[string]float64
}
```

When the rates are requested by ob-go, it can get the rates for all CurrencyDefinitions in the mainnet dictionary from the local ExchangeRater (which looks much like the currency ExchangeRate provider in the BTC SPV wallet) merged with the results from any wallets which also implement the ExchangeRater interface. Not sure how native vs wallet values get overwrite priority, but this would give us a local exchange rate impl (which would be tested) and the integration test would be easier. Thoughts?

placer14: The exchange rater would also need to manage multiple pairs. As different wallets might offer different reserve currencies. I imagine the exchange rater cache would look something like `map[string]map[string]float64` or maybe `map[repo.CurrencyCode]repo.ReserveRatesCache`
where the key is the reserve currency (which would be largely populated with BTC values) and the `repo.ReserveRatesCache` is a model which knows the rates for the `repo.CurrencyCode` key requested. Once we have this mapping, we can ignore the reserve currency completely and could just ask for a ConvertTo from LTC->USD and it would know how to handle the rates and find a matching reserve to convert through. 

Concern: Nothing would stop the wallet from injecting bad rates into the ExchangeRater. Would be nice to find some way to prove the rates are legit. (Granted wallets already require a degree of trust, this design would allow a wallet to inject its rates into any transaction to or from this wallet's native currency(s). Before, the wallet could only impact its own blockchain state on behalf of the user. Here, the wallet might impact vendors which have not accepted the buyer's wallet's rate. Or, worse, could make counterparty verification of the order value inconsistent depending on which rates are consumed on either side of the trade.)

PR #1466, opened 2019-02-26
Revert mobile logging to file

placer14:
iOS does not permit filesystem access, need to find another way to support this case on iOS. I was unable to build this with `gomobile` and assume I have some toolchain problems to resolve before I can again.

Would appreciate it if someone could try building the frameworks before accepting this.

### DIFF
--- a/mobile/config.go
+++ b/mobile/config.go
@@ -1,15 +1,12 @@
 package mobile
 
 import (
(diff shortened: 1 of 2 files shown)


placer14:
FYI, the PR build will pass as soon as the fix in https://github.com/OpenBazaar/openbazaar-go/pull/1465 is merged.

PR #1469, opened 2019-02-27
Extract SignPayload and VerifyPayload w unit tests

placer14:
Improvement intended for #1345

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -8,14 +8,13 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


api/jsonapi.go
```diff
@@ -2266,7 +2265,7 @@ func (i *jsonAPIHandler) POSTReleaseEscrow(w http.ResponseWriter, r *http.Reques
 
 func (i *jsonAPIHandler) POSTSignMessage(w http.ResponseWriter, r *http.Request) {
 	type plaintext struct {
-		Content string `json:"content"`
+		Content []byte `json:"content"`
```
cpacia: Can a client easily post a byte array?

placer14: The client won't need to change anything, it just marshals the json value into the field as a []byte instead of a string.

cpacia: oh ok

PR #1484, opened 2019-03-06
Fix json api keyname in setapicreds

dagurval:
Currently `setapicreds` is broken as it just adds a section `JSON_API` to the configuration file that isn't read by the server. 

The server is supposed to read it back here:
https://github.com/OpenBazaar/openbazaar-go/blob/master/schema/configuration.go#L133


cpacia:
Ah good catch!

PR #1485, opened 2019-03-06
Improve errors on config file issues

dagurval:
Give more useful feedback to user. For example, this will output

> config file is malformed: Error parsing section 'JSON-API'
> exit status 1

instead of just

> config file is malformed
> exit status 1

### DIFF
--- a/schema/configuration.go
+++ b/schema/configuration.go
@@ -2,7 +2,7 @@ package schema
 
 import (
 	"encoding/json"
-	"errors"
+	"fmt"
 	"time"
 )
 
@@ -55,7 +55,15 @@ type DataSharing struct {
 	PushTo              []string
 }
 
-var MalformedConfigError = errors.New("config file is malformed")
+type MalformedConfigError struct {
+	What string
+}
+
+var _ error = MalformedConfigError{}
+
+func (err MalformedConfigError) Error() string {
+	return fmt.Sprintf("config file is malformed: %s", err.What)
+}
 
 func DefaultWalletsConfig() *WalletsConfig {
 	var feeAPI = "https://btc.fees.openbazaar.org"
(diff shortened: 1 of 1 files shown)


cpacia:
Thanks dagur

schema/configuration.go
```diff
@@ -419,48 +442,49 @@ func GetDataSharing(cfgBytes []byte) (*DataSharing, error) {
...
 	dataSharing := new(DataSharing)
 
 	cfg, ok := cfgIface.(map[string]interface{})
 	if !ok {
-		return dataSharing, MalformedConfigError
+		return dataSharing, MalformedConfigError{What: "Config file is empty"}
 	}
 
-	dscfg, ok := cfg["DataSharing"]
+	const DATASHARING string = "DataSharing"
```
placer14: We usually just make constants with capital first letters instead of all caps, FYI. Might also be a good idea to group the constants at the top of the function:
```
const (
   JSON-API = "JSON-API"
   DataSharing = "DataSharing"
)
```

schema/configuration.go
```diff
@@ -55,7 +55,15 @@ type DataSharing struct {
 	PushTo              []string
 }
 
-var MalformedConfigError = errors.New("config file is malformed")
+type MalformedConfigError struct {
+	What string
```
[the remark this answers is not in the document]
placer14: Better yet, I see you make MalformedConfigErrors in four different ways.

- MalformedConfigError{What: "JSON error"} (general error case)
- MalformedConfigError{What: sectionError(JSONAPI)} (specific section case)
- MalformedConfigError{What: keyError(JSONAPI, "HTTPHeaders")} (key within section case)
- MalformedConfigError{What: keyError("", "Dropbox-api-token")} (key outside section case?)

Instead of making it the responsibility of the dev using `MalformedConfigError` to understand how to use each of the cases, you should make helpers which produce the right version of `MalformedConfigError`. Something like:

- malformedConfigError("invalid raw JSON") // we already know it's a error, don't repeat yourself...
- malformedConfigError("invalid section JSONAPI")
- malformedConfigError("JSONAPI", "HTTPHeaders") // section-specific helper function
- malformedConfigError("Dropbox-api-token") // key-specific helper function outside of a section

Instead consider variadic parameters:
```
func malformedConfigKey(pathArgs ...string) malformedConfigError {
  return malformedConfigKey{path: pathArgs...}
}

func (err malformedConfigKey) Error() string {
  return fmt.Sprintf("malformed config: %s", strings.Join(err.path..., "."))
}
```

This makes `malformedConfigError` have a "variadic" parameter which is treated like a slice. Then you can pass in as many values as you like to describe how "deep" the malformed data is in the JSON.

Ex: 
```
malformedConfigError("JSON is invalid") // "malformed config: JSON is invalid"
malformedConfigError("JSON-API","config") // "malformed config: JSON-API.config"
malformedConfigError("JSON-API","space1","value1") // "malformed config: JSON-API.space1.value1"
```

schema/configuration.go
```diff
@@ -55,7 +55,15 @@ type DataSharing struct {
 	PushTo              []string
 }
 
-var MalformedConfigError = errors.New("config file is malformed")
+type MalformedConfigError struct {
```
placer14: This does not need to be a public type. Make the first letter lowercase to make private. `malformedConfigError`

cpacia:
Closing since it's been open a while. We can revisit later if need be.

dagurval:
Thanks for the review @placer14. I've pushed an update where I think I've addressed all yours points.

@cpacia I forgot about this PR, can you re-open again?

PR #1488, opened 2019-03-07
Create migration019

cpacia:
- Migration019 to update the listing index hashes to match the hashes created by IPFS 0.4.18.
- Rolling migration to update image hashes within listings at the time they are updated.


core/images.go
```diff
@@ -189,3 +190,69 @@ func (n *OpenBazaarNode) GetBase64Image(url string) (base64ImageData, filename s
...
+		if originalID.Version() > 0 {
+			hash, err := ipfs.AddFile(n.IpfsNode, path.Join(n.RepoPath, "root", "images", "original", image.Filename))
+			if err != nil {
+				return err
+			}
+			image.Original = hash
+		}
+		listing.Item.Images[i] = image
+	}
+	return nil
```
placer14: Change this behavior so it's not repeating every time. Extract this behavior into a helper or a local anon function:
```
originalID, err := cid.Decode(image.Original)
		if err != nil {
			return err
		}
		if originalID.Version() > 0 {
			hash, err := ipfs.AddFile(n.IpfsNode, path.Join(n.RepoPath, "root", "images", "original", image.Filename))
			if err != nil {
				return err
			}
			image.Original = hash
		}
```

core/images_test.go
```diff
@@ -27,3 +37,105 @@ var (
...
+	if err = testRepo.BuildSchemaDirectories(); err != nil {
+		t.Fatal(err)
+	}
+	defer testRepo.DestroySchemaDirectories()
+
+	node := OpenBazaarNode{IpfsNode: ipfsNode, RepoPath: testRepo.DataPath()}
+
+	listing := factory.NewListing("testListing")
+	for _, image := range listing.Item.Images {
+		err := ioutil.WriteFile(path.Join(testRepo.DataPath(), "root", "images", "large", image.Filename), []byte("image"), os.ModePerm)
```
placer14: There's a helper function: `testRepo.DataPathJoin(...string)`

core/images_test.go
```diff
@@ -27,3 +37,105 @@ var (
...
+		err = ioutil.WriteFile(path.Join(testRepo.DataPath(), "root", "images", "original", image.Filename), []byte("image"), os.ModePerm)
+		if err != nil {
+			t.Error(err)
+		}
+	}
+
+	checkImage := func(image *pb.Listing_Item_Image) {
+		largeID, err := cid.Decode(image.Large)
+		if err != nil {
+			t.Error(err)
```
placer14: Decodes should be `t.Fatal` so the rest of the test stops.

core/images_test.go
```diff
@@ -27,3 +37,105 @@ var (
...
+		}
+		if largeID.Version() > 0 {
+			t.Error("Large image failed to migrate to v0")
```
placer14: These can remain as `Error` since it doesn't affect the success of the rest of the test.

PR #1501, opened 2019-03-15
Fix NotifyModerators panic if node is not yet bootstrapped

### LINKED ISSUE #1451: Can't Make Offline Purchase
When making a purchase from an offline 2.3 node, a node using the IPFS refactor branch returns a status of 200 OK from the ob/purchase call, but returns no data.

### LINKED ISSUE #1502: Mobile access on POST /ob/settings causes panic
When POST /ob/settings is called, moderators are notified. In mobile, this occurs before the MessageService is ready and causes a panic.

cpacia:
Refactors the notify moderators function. It removes a race condition which could result in the wrong modes being sent notifications. Also waits for the service to start before sending to avoid a panic. 

Also adds blockingStartupMiddleware to purchase post which avoids a panic for trigger happy users. 

Fixes #1502 
Fixes #1451

### DIFF
--- a/api/utils_test.go
+++ b/api/utils_test.go
@@ -10,16 +10,16 @@ func Test_extractModeratorChanges(t *testing.T) {
 
 	toAdd, toDelete := extractModeratorChanges(newModList, &currentModList)
 	if len(toAdd) != 1 {
-		t.Errorf("Test_extractModeratorChanges returned incorrect number of additions: expected %d got %d", 1, len(toAdd))
+		t.Errorf("Returned incorrect number of additions: expected %d got %d", 1, len(toAdd))
 	}
 	if len(toDelete) != 1 {
-		t.Errorf("Test_extractModeratorChanges returned incorrect number of deletions: expected %d got %d", 1, len(toDelete))
+		t.Errorf("Returned incorrect number of deletions: expected %d got %d", 1, len(toDelete))
 	}
 
 	if toAdd[0] != "x" {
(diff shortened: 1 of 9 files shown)


api/utils_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"testing"
+)
+
+func Test_extractModeratorChanges(t *testing.T) {
+	currentModList := []string{"a", "b", "c"}
+	newModList := []string{"a", "x", "c"}
+
+	toAdd, toDelete := extractModeratorChanges(newModList, &currentModList)
+	if len(toAdd) != 1 {
+		t.Errorf("Test_extractModeratorChanges returned incorrect number of additions: expected %d got %d", 1, len(toAdd))
```
placer14: The error output will have the test name included. Lead with a lowercase verb.

core/moderation.go
```diff
@@ -262,32 +262,12 @@ func (n *OpenBazaarNode) SetModeratorsOnListings(moderators []string) error {
...
-
-	removedMods := currentMods
-
+func (n *OpenBazaarNode) NotifyModerators(addedMods, removedMods []string) error {
+	n.Service.WaitForReady()
 	for _, mod := range addedMods {
 		go n.SendModeratorAdd(mod)
 	}
-	for mod := range removedMods {
+	for _, mod := range removedMods {
```
placer14: Nice catch.

placer14: Nice catch.

PR #1515, opened 2019-03-22
[WIP] Native Embedded Tor for OpenBazaar

hoffmabc:
This uses bine and static tor libs to allow everyone to use Tor and OpenBazaar without running anything outside of the app. 

You will need to do the following:

```
go get -u github.com/cretz/tor-static
cd $GOPATH/github.com/cretz/tor-static 
go run build.go build-all
```

Once that's built you can run OpenBazaar with the nativetor flag set to use it:

```
go run openbazaard.go start --nativetor 
```

We'll need to figure out how to bundle it.


placer14:
I believe there was a window build concern that we couldn't get around for the time being. Based on our prior conversation, @hoffmabc, I believe we agreed to close this and revisit for later? Can we close?

hoffmabc:
I’m going to close this and revisit

PR #1530, opened 2019-04-01
#1527 Added Method check for HEAD to validat undefined endpoint

### LINKED ISSUE #1527: Requesting an endpoint which is not defined on JSON API should return 404
Currently, an undefined endpoint responds with a 200 and a response body of
```
{
    "success": false,
    "reason": "Not Found"
}
```

An example fo this can be seen with `curl  localhost:4002/ob/blah -I` 
```
$ curl  user:password@0.0.0.0:4002/ob/blah -I
HTTP/1.1 200 OK
Content-Type: application/json
Date: Thu, 28 Mar 2019 02:33:26 GMT
```

Instead, the HTTP response code should be 404 Not Found.

jwfrizzell:
When making the request curl localhost:4002/ob/blah -I the method being passed was HEAD. Added this to the switch case and passed as a get request.

Fixes #1527

### DIFF
--- a/api/jsonapi.go
+++ b/api/jsonapi.go
@@ -176,6 +176,8 @@ func (i *jsonAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		deleter(i, u.String(), w, r)
 	case "PATCH":
 		patch(i, u.String(), w, r)
+	case "HEAD":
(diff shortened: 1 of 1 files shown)


placer14:
Updated API documentation to reflect support for HEAD requests.

PR #1532, opened 2019-04-05
Create Migration021

### LINKED ISSUE #1531: Migrate config to include option for AutoRelay defaulted true
IPFS 0.4.19 rebase should include new option in config for `AutoRelay` to enable relaying usage when behind a NAT-like device.

cpacia:
This migration migrates the config file to include the EnableAutoRelay option
in the Swarm config. This will allow the node to automatically detect if it's
behind a restricted NAT, automatically find relay nodes, and set their own
swarm addresses to use the relays.

closes #1531

### DIFF
--- a/repo/init.go
+++ b/repo/init.go
@@ -17,7 +17,7 @@ import (
 	"github.com/tyler-smith/go-bip39"
 )
 
-const RepoVersion = "21"
+const RepoVersion = "22"
 
 var log = logging.MustGetLogger("repo")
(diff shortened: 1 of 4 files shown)


cpacia:
This config object is added by IPFS when the node is created. It's not in our configuration file.

placer14:
Ah, I didn't know that, thanks. And also for adding the AutoNAT default to the config as well.

PR #1533, opened 2019-04-05
Use ETH on mainnet or testnet

amangale:
This PR will be used for 
1) Keeping ethereum-master in sync with master
2) Changes for the testnet (rinkeby) and mainnet modes for ethereum


repo/init.go
```diff
@@ -35,7 +35,7 @@ func DoInit(repoRoot string, nBitsForKeypair int, testnet bool, password string,
 	if err != nil {
 		return err
 	}
-	if nodeSchema.BuildSchemaDirectories(); err != nil {
+	if err := nodeSchema.BuildSchemaDirectories(); err != nil {
```
placer14: Wow. Thanks for running the linter through this code.

placer14:
There are also some travis failures that need to be resolved as well.

amangale:
> There are also some travis failures that need to be resolved as well.
> https://travis-ci.org/OpenBazaar/openbazaar-go/builds/518869366

The latest build is also failing: https://travis-ci.org/OpenBazaar/openbazaar-go/builds/521773356
The issue reported here cannot be reproduced locally using the golangci-lint as well as gofmt.

placer14:
It appears that this travis failure is due to the merge and not because of your branch. We can handle that travis failure in a separate PR. A few other changes requested for the Migration023 tests (validating all the changed values in the test Up and Down) and it should be ready to merge.

PR #1538, opened 2019-04-11
Add a blurb about more commands

pRizz:
This would have saved me a decent amount of time. I was looking for something like the `restore` command but didn't see it mentioned in the docs anywhere.


placer14:
Thank you for the suggestion. We'll add better links to the README and include the subcommand docs  to our docs site that we've been improving: https://docs.openbazaar.org

PR #1539, opened 2019-04-11
Add logging for mobile nodes

hoffmabc:
This is to support PR https://github.com/OB1Company/native-bazaar/pull/830
on the Haven repo.


placer14:
I tried to test this but was unable to get my build to complete (it failed while it was syncing gradle?). Can we get a confirmation on whether this works or not before we merge? /cc @hoffmabc @drwasho

PR #1540, opened 2019-04-12
Add spend all functionality to all wallets

### LINKED ISSUE #929: Add Sweep Wallet API
Currently is it very difficult to move all funds from the internal wallet to another address. To facilitate this, there should be an API to move all funds.

Suggestion:

Add a wallet/spendall endpoint that works similarly to the spend endpoint, but moves all funds, minus a fee, to the supplied address.

Data Sent:
```
{
  "feeLevel": "NORMAL",
  "memo": "move to personal wallet",
  "address": "tb1q2gzlqfs27al4087h56gg9pgza2l7fp0v7tl0llgd9agfhx6us8qs5ht6fw",
}
```

Data Received
```
{
  "amount": 27106,
  "confirmedBalance": 0,
  "memo": "move to personal wallet",
  "timestamp": "2018-04-19T15:16:18-04:00",
  "txid": "ee4f0bb4b405f17eeeb3e84aaf6bb02a13763cf87a71375f7d59ca5fe916a47d",
  "unconfirmedBalance": 0
}
```

The above is the same as the spend endpoint, but does not send an amount (since the amount is always all available funds).

cpacia:
This commit updates the wallet-interface, multiwallet, spvwallet, and
bch spvwallet to add a spendAll bool to the spend interface method that will
allow us to sweep all funds from the wallet. Currently to remove all funds you
need to try to guess what fee the wallet will use and substract that from the
amount you're trying to spend. The spendAll option will subtract the fee from
the amount you're sending rather than adding it. Improving the UX in this regard.

A new bool is added to the spend POST JSON body.

closes #929

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -33,91 +33,87 @@
 		},
 		{
 			"ImportPath": "github.com/OpenBazaar/multiwallet",
-			"Rev": "00a64ca560adcc87cc0ffee0102f3c43f4a701bb"
+			"Rev": "5d687b8c67d127035479739cbf2258f7b275f3e7"
 		},
 		{
 			"ImportPath": "github.com/OpenBazaar/multiwallet/bitcoin",
-			"Rev": "00a64ca560adcc87cc0ffee0102f3c43f4a701bb"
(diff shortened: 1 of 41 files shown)


placer14:
Please merge when ready (pref after the docs are updated).

Issue #1535, opened 2019-04-08
Include daemon version in log output


placer14:
In order to improve debugging, we should include the version in log output. Currently, the version is only outputted to stdout. Instead, current version should be included on startup.

Issue #1524, opened 2019-03-26
Can't Edit Crypto Listings With Zero Inventory


jjeffryes:
I have a crypto listing that sold out, reducing it's inventory to zero.

It can't be edited as long as the inventory remains zero. This prevents updating the listing, or using the `ob/bulkupdatecurrency` call.

The error returned is `cryptocurrency listing quantity must be a non-negative integer`.

The error appears to be happening here:
```
func validateListingSkus(listing *pb.Listing) error {
    if listing.Metadata.ContractType == pb.Listing_Metadata_CRYPTOCURRENCY {
        for _, sku := range listing.Item.Skus {
            if sku.Quantity < 1 {
                return ErrCryptocurrencySkuQuantityInvalid
            }
        }
    }
    return nil
}
```

PR #1541, opened 2019-04-15
#1524 CRYPTO LISTING UPDATE #1535 SERVER DAEMON LOG

jwfrizzell:
Allow for the editing of crypto listing when inventory is 0. Currently when the crypto listing is zero the store owner can't update any of the crypto listing info.


jwfrizzell:
Should I close this request or is it still in consideration?

Issue #1542, opened 2019-04-15
Listings/Posts with Emoji Characters Only Fails


hoffmabc:
![image](https://user-images.githubusercontent.com/45482/56136390-b46c3400-5f60-11e9-9b46-28bc486e34da.png)

placer14:
Resolved as part of #1543

PR #1543, opened 2019-04-15
Use go-emoji to convert to html chars

hoffmabc:
Potential fix for #1542

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -2,6 +2,9 @@ package core
 
 import (
 	"errors"
+	"fmt"
+	"regexp"
+	"unicode/utf8"
 
 	"gx/ipfs/QmSY3nkMNLzh9GdbFKK5tT7YMfLpf52iUZ8ZRkr29MJaa5/go-libp2p-kad-dht"
 	libp2p "gx/ipfs/QmTW4SdgBWq9GjsBsHeUx8WuGxzhgzAf88UMH2w62PC8yK/go-libp2p-crypto"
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -2,14 +2,19 @@ package core
...
 	"gx/ipfs/QmSY3nkMNLzh9GdbFKK5tT7YMfLpf52iUZ8ZRkr29MJaa5/go-libp2p-kad-dht"
 	libp2p "gx/ipfs/QmTW4SdgBWq9GjsBsHeUx8WuGxzhgzAf88UMH2w62PC8yK/go-libp2p-crypto"
 	ma "gx/ipfs/QmTZBfrPJmjWsCvHEtX5FE6KimVJhsJg5sBbqEFYf4UZtL/go-multiaddr"
-	cid "gx/ipfs/QmTbxNB1NwDesLmKTscr4udL2tVP7MaxvXnD1D9yX7g3PN/go-cid"
+	"gx/ipfs/QmTbxNB1NwDesLmKTscr4udL2tVP7MaxvXnD1D9yX7g3PN/go-cid"
 	"gx/ipfs/QmUadX5EcvrBmxAV9sE7wUWtWSqxns5K84qKJBixmcT1w9/go-datastore"
-	peer "gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer"
-	routing "gx/ipfs/QmYxUdYY9S6yg5tSPVin5GFTvtfsLauVcr7reHDD3dM8xf/go-libp2p-routing"
+	"gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer"
+	"gx/ipfs/QmYxUdYY9S6yg5tSPVin5GFTvtfsLauVcr7reHDD3dM8xf/go-libp2p-routing"
```
placer14: Can you undo the unrelated import changes?

core/core.go
```diff
@@ -23,9 +28,8 @@ import (
 	"github.com/OpenBazaar/openbazaar-go/repo"
 	sto "github.com/OpenBazaar/openbazaar-go/storage"
 	"github.com/btcsuite/btcutil/hdkeychain"
-	"github.com/gosimple/slug"
 	"github.com/ipfs/go-ipfs/core"
-	logging "github.com/op/go-logging"
+	"github.com/op/go-logging"
```
placer14: Here, too.

core/core.go
```diff
@@ -2,14 +2,19 @@ package core
 
 import (
 	"errors"
+	"fmt"
+	"regexp"
+	"unicode/utf8"
+
+	"github.com/gosimple/slug"
```
placer14: If the linter will let you, move this back where it was.

hoffmabc: this is new though.

placer14:
I removed the dependency on the `assert` library. One less thing to vendor. Let's see if travis passes and then we'll merge.

cpacia:
For this one maybe we should add a test to make sure non-emoji strings do not get mutated?

PR #1544, opened 2019-04-15
Add API router functionality

cpacia:
Requirements

- [x] Create API routing implementation that queries a server for IPNS values (but not IPFS providers). 

- [x] Create a wrapper around API router and the DHT that tires the DHT first where possible then the API.

- [x] Whenever querying for values, returned values should be pushed back to the API to make sure the API has them available.

- [x] Publishes should go out to all routers.

- [x] Remove the obsolete code which queried the gateway for IPNS values when the DHT failed.

- [x] Migrate the config file to include the new API router URI

- [x] Make sure API router can handle sending queries over Tor.

- [x] API router not used when in regtest mode.


ipfs/api_router.go
```diff
@@ -41,7 +41,7 @@ func NewAPIRouter(uri string) APIRouter {
 	return APIRouter{uri: uri}
 }
 
-func (r APIRouter) Start(proxyDialer proxy.Dialer) {
+func (r *APIRouter) Start(proxyDialer proxy.Dialer) {
```
placer14: Good catch.

ipfs/api_router.go
```diff
@@ -57,23 +57,28 @@ func (r APIRouter) Bootstrap(_ context.Context) error {
...
 func (r APIRouter) PutValue(ctx context.Context, key string, value []byte, opts ...ropts.Option) error {
 	<-r.started
-	req, err := http.NewRequest("PUT", r.pathForKey(key), bytes.NewBuffer(value))
+	path := r.pathForKey(key)
+	req, err := http.NewRequest("PUT", path, bytes.NewBuffer(value))
 	if err != nil {
 		return err
 	}
 
+	log.Debugf("Write value to %s", path)
```
placer14: Nitpick: No captials in the logs.

placer14:
Thanks, LGTM

PR #1546, opened 2019-04-16
FEATURE: Add endpoint to resolve peer IDs into IPNS entries.

drwasho:


api/endpoints.go
```diff
@@ -195,6 +195,8 @@ func get(i *jsonAPIHandler, path string, w http.ResponseWriter, r *http.Request)
 		i.GETWalletStatus(w, r)
 	case strings.HasPrefix(path, "/ob/ipns"):
 		i.GETIPNS(w, r)
+	case strings.HasPrefix(path, "/ob/resolveipns"):
+		i.ResolveIPNS(w, r)
```
placer14: Core function `ResolveIPNS` does not follow pattern of `HTTPVERBFunctionName`. Nitpick: These two function calls are a little ambiguous to a newcomer. Any way to clarify?

api/jsonapi.go
```diff
@@ -3803,6 +3803,27 @@ func (i *jsonAPIHandler) GETIPNS(w http.ResponseWriter, r *http.Request) {
...
+func (i *jsonAPIHandler) ResolveIPNS(w http.ResponseWriter, r *http.Request) {
+	_, peerID := path.Split(r.URL.Path)
+
+	pid, err := peer.IDB58Decode(peerID)
+	if err != nil {
+		ErrorResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*180)
```
placer14: The timeout for this call is 2 minutes. Is this intended? /cc @drwasho

api/endpoints.go
tyler-smith: You're right, I totally overlooked the http verb prefix thing. I'll fix that.

tyler-smith: I'm not sure how to clarify without renaming /ob/ipns, which would not be ideal.

One option would be making it a flag on /ob/ipns, e.g. /ob/ipns?resolve=true

api/jsonapi.go
tyler-smith: Yeah I need it to be at least 2 minutes but we could make it configurable if somebody only wanted entries that resolve quickly.

placer14: This is no problem for me, I just want to make sure this was intended.

api/endpoints.go
placer14: I can't think of any suggestions to offer either. We can punt if we can't think of anything better. I think the GET param overloads the endpoint with two different behaviors. And really, I would want to change the `ipns` endpoint to be more specific, but that has repercussions which are annoying. We can leave it as-is for now.

cpacia:
Looks good.

PR #1549, opened 2019-04-18
Increase chat timestamp resolution

### LINKED ISSUE #1545: Non-deterministic chat messages
When sending a series of closely timed chat messages to a peer, they can easily be presented out-of-order in the UI. We apply timestamps to the messages to ensure they can be deterministically ordered, however, the timestamps only have resolution to the second and it's not unreasonable for multiple chat messages to potentially arrive within the same second.

So we'd like to increase this resolution and understand what the limitation of the new resolution are to handle them appropriately in the future.

placer14:
Fixes #1545

### DIFF
--- a/repo/db/chat.go
+++ b/repo/db/chat.go
@@ -23,7 +23,7 @@ func (c *ChatDB) Put(messageId string, peerId string, subject string, message st
 
 	// timestamp.UnixNano() is undefined when time has a zero value
 	if timestamp.IsZero() {
-		log.Errorf("putting chat message (%s): recieved zero timestamp, using current time", messageId)
+		log.Warningf("putting chat message (%s): recieved zero timestamp, using current time", messageId)
 		timestamp = time.Now()
 	}
 
--- a/repo/db/chat.go
+++ b/repo/db/chat.go
@@ -47,7 +47,7 @@ func (c *ChatDB) Put(messageId string, peerId string, subject string, message st
 		subject,
 		message,
 		readInt,
-		int(timestamp.Unix()),
+		int(timestamp.UnixNano()),
 		outgoingInt,
 	)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


repo/db/chat.go
```diff
@@ -21,6 +21,12 @@ func (c *ChatDB) Put(messageId string, peerId string, subject string, message st
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	// timestamp.UnixNano() is undefined when time has a zero value
+	if timestamp.IsZero() {
+		log.Errorf("putting chat message (%s): recieved zero timestamp, using current time", messageId)
```
cpacia: Maybe a warn here instead of an error?

repo/migrations/Migration023_test.go
```diff
@@ -0,0 +1,158 @@
...
+	if err = appSchema.BuildSchemaDirectories(); err != nil {
+		t.Fatal(err)
+	}
+	defer appSchema.DestroySchemaDirectories()
+
+	var (
+		databasePath = appSchema.DatabasePath()
+		schemaPath   = appSchema.DataPathJoin("repover")
+
+		//schemaSQL          = "pragma key = 'foobarbaz';"
```
cpacia: Maybe remove commented out var

repo/migrations/Migration023_test.go
```diff
@@ -0,0 +1,158 @@
...
+		CreateTableChatSQL = "create table chat (messageID text primary key not null, peerID text, subject text, message text, read integer, timestamp integer, outgoing integer);"
+		insertChatSQL      = "insert into chat(messageID, peerID, subject, message, read, timestamp, outgoing) values(?,?,?,?,?,?,?);"
+		selectChatSQL      = "select messageID, timestamp from chat;"
+		setupSQL           = strings.Join([]string{
```
cpacia: and update here

cpacia:
Looks really good

placer14:
We're going to move forward with this fix seeing that we have confirmed it properly resolves #1545. Need to update timestamps in the JSON API to ensure the full resolution can be shown in the string representation. Recommendation is to use RFC3339 which looks like `2006-01-02T15:04:05.999999999Z07:00`. Will create a new issue to track the changes needed. (FYI: @drwasho @jjeffryes)

PR #1553, opened 2019-04-23
Fix IPFS protocol identifier handling

hoffmabc:
Update: This PR also manages the other protocol strings the daemon uses (kademlia, dht legacy, app, and bitswap protocols).

Original Motivation:
So currently we are not setting our DHT routing options correctly for testnet nodes (AFAICT). This means that when publishing and resolving, our testnet nodes just use mainnet DHT and mainnet nodes to store data. This code actually changes that and makes it so we only use testnet nodes. The only problem right now is that there are actually not any nodes that speak this protocol at the moment so it seems like all our testnet nodes will need to upgrade to this code before we see it acting properly (including seed nodes). 

**This needs validation though**

### DIFF
--- a/cmd/restore.go
+++ b/cmd/restore.go
@@ -5,13 +5,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"gx/ipfs/QmQmhotPUzVrMEWNK3x1R5jQ5ZHWyL7tVUrmRPjrBrvyCb/go-ipfs-files"
+	files "gx/ipfs/QmQmhotPUzVrMEWNK3x1R5jQ5ZHWyL7tVUrmRPjrBrvyCb/go-ipfs-files"
 
 	"gx/ipfs/QmRxk6AUaGaKCfzS1xSNRojiAPd7h2ih8GuCdjJBF3Y6GK/go-libp2p"
-	"gx/ipfs/QmSY3nkMNLzh9GdbFKK5tT7YMfLpf52iUZ8ZRkr29MJaa5/go-libp2p-kad-dht/opts"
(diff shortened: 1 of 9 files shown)


cmd/restore.go
```diff
@@ -26,7 +26,7 @@ import (
 	"syscall"
 	"time"
 
-	"gx/ipfs/QmXLwxifxwfc2bAwq6rdjbYqAsGzWsDE9RM5TWMGtykyj6/interface-go-ipfs-core"
+	iface "gx/ipfs/QmXLwxifxwfc2bAwq6rdjbYqAsGzWsDE9RM5TWMGtykyj6/interface-go-ipfs-core"
```
hoffmabc: why do we need aliases on all these?

placer14: They are there for convenience and added by `goimports` on saving the file. When the package has a conflicting name with another package or doesn't match the final token of the import path, putting the name helps the dev easily identify which package provides `iface` in the code below.

hoffmabc: I know what it does ;) I'm just saying previously you've been hesitant to add those changes when not absolutely necessary like on my last PR. So was just curious why this time was different.

placer14: I'm not sure if this was the case on your PR where I commented, but if it's just the imports being changed and nothing else in the file, I'd rather the change be omitted as it's functionally a no-op change. Of course, it's also possible I'm just being inconsistent. We generally have a lot noise in our repo and that's something that I grapple back and forth with being pedantic about and just accepting. (Of course, a lot of this must be accepted due to golang's opinionated formatting.)

tyler-smith: I agree with Brian but not enough to hold up this PR I guess.

placer14: I don't mind removing them. We previously agreed to remove the goimports requirement from the linter to revert back to gofmt and my editor is still configured to run goimports on save. I've removed that settings to use gofmt again so this doesn't keep happening.

Issue #1574, opened 2019-05-15
[WIP] Provide NoOp Migration pattern


placer14:
When there are long-lived feature branches being developed in parallel, it's not uncommon for migration numbers to collide between branches (since, for the time being, we use numerically sequenced migrations). In order to create placeholder migrations for numbers which are already being used in other branches (since the migrator doesn't know how to skip numbers), this pattern would allow the branch which doesn't implement the actual migration to use a `MigrationNoOp` to hold the place while still allowing later migrations to added on.... allowing the conflict to be easily resolved between the two branches are merged together.

This is one approach to the problem. Drawbacks include not being able to retroactively apply actual migrations once the NoOp migration has been run on some schema. (If a branch runs a no-op migration X on some data, it cannot be rolledback to apply the proper/actual migration later.)

placer14:
Another interesting point which was raised in discussion is that upstream branches which have NoOp migrations in anticipation of a downstream branches merge would cause the upstream branch to be held until the merge is completed (as NoOp migrations necessarily cannot be included in anything but a testing release and not intended for public use). There are also implications here with users that build from `master`. So more strictly than not including a NoOp migration in a public release, we probably can't include it in `master` as well. With these drawbacks in mind, I'm strongly leaning away from this solution.

placer14:
We decided we weren't going to use this in the short term until we design a better approach.

PR #1575, opened 2019-05-15
Create no-op migration pattern

placer14:
Implementation attempt on #1574


placer14:
We decided that we will manually reconcile migration conflicts in the short term and will carve out an opportunity later for a proper migration improvement. Closing.

PR #1576, opened 2019-05-15
Configure ipfs global vars in mobile package

### LINKED ISSUE #1570: Mobile: Fails to resolve old-style IPFS hashes
Using the `master` branch of version `13.3`, I was able to resolve by IPFS content with old-style hashes (e.g. `zb2rhj2TpNPsjPsPNFXoaiaeBtKny1KxcFgFWFiEkeKkRJqJ8`, `zb2rhiG4zJJaASeuxfdtxJhbQY1MpwvW2H1VqDA9FALBw3Kcq`). However, when I switched run the mobile version (`go run mobile/cmd/main.go -d mobileTest`), these hashes failed to resolve.

cpacia:
closes #1570

### DIFF
--- a/mobile/node.go
+++ b/mobile/node.go
@@ -89,6 +89,8 @@ func NewNodeWithConfig(config *NodeConfig, password string, mnemonic string) (*N
 	repoLockFile := filepath.Join(config.RepoPath, fsrepo.LockFile)
 	os.Remove(repoLockFile)
 
+	ipfs.UpdateIPFSGlobalProtocolVars(config.Testnet)
+
 	// Logging
 	backendStdout := logging.NewLogBackend(os.Stdout, "", 0)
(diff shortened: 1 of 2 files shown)


placer14:
@cpacia Could you investigate how we can include the tests in https://github.com/OpenBazaar/openbazaar-go/pull/1578 as part of this PR and let me know what you find?

cpacia:
I can change it but in general I'm not a fan of different packages for tests. This one doesn't need to test a private field "yet", but it may in the future.

placer14:
General rule when testing: if you need to reach into the internals of the object to validate something, that's probably an indication that the object wasn't designed very well and is highly dependent on other objects. Making our unit testing easier is a lot about keeping dependency chains short.

PR #1584, opened 2019-05-17
Manually send order messages and trigger offline message scan

### LINKED ISSUE #1459: ACKs for Some or All Direct Messages
We should consider whether there are some messages that need an ACK even when they are sent directly to another online node.

If there are, they need to be carefully examined for how this affects the user experience and state management, so user's don't end up in unexpected states if messages are resent.

Related to #1442

amangale:
This PR adds order_messages table which is used to store order related messages for a node. It also adds two api end points : GET `/ob/scanofflinemessages` and POST `/ob/sendordermessage` to manually trigger offline message scan and send any previously sent order message of type - Message_ORDER, Message_ORDER_REJECT, Message_ORDER_CANCEL, Message_ORDER_CONFIRMATION, Message_ORDER_FULFILLMENT, Message_ORDER_COMPLETION.

Fixes #1459
Fixes #1584


placer14:
Initial conversation around this improvement suggests we will move forward with this PR, pending final peer review of the implementation.

api/jsonapi.go
```diff
@@ -4150,3 +4153,62 @@ func (i *jsonAPIHandler) GETPost(w http.ResponseWriter, r *http.Request) {
 	}
 	SanitizedResponseM(w, out, new(pb.SignedPost))
 }
+
+// POSTSendOrderMessage - used to manually send an order message
+func (i *jsonAPIHandler) POSTSendOrderMessage(w http.ResponseWriter, r *http.Request) {
+	type sendRequest struct {
+		OrderID     string `json:"orderID"`
+		MessageType int32  `json:"messageType"`
```
placer14: MessageType should be a string instead of the internal enum we use to identify a message type. (The internal enum is an implementation detail of the protocol that we shouldn't expose if we don't have to.) Verbatim strings for each message type would be fine for now.

api/endpoints.go
```diff
@@ -113,6 +113,8 @@ func post(i *jsonAPIHandler, path string, w http.ResponseWriter, r *http.Request
 		i.POSTPost(w, r)
 	case strings.HasPrefix(path, "/ob/bulkupdatecurrency"):
 		i.POSTBulkUpdateCurrency(w, r)
+	case strings.HasPrefix(path, "/ob/sendordermessage"):
+		i.POSTSendOrderMessage(w, r)
```
placer14: The semantics around this endpoint should be clear that this is for resending. Maybe `/ob/resendordermessage` and `POSTResendOrderMessage` instead?

api/jsonapi.go
```diff
@@ -63,11 +63,14 @@ type jsonAPIHandler struct {
...
 }
 
+var lastManualScan time.Time
+
 func newJSONAPIHandler(node *core.OpenBazaarNode, authCookie http.Cookie, config schema.APIConfig) *jsonAPIHandler {
 	allowedIPs := make(map[string]bool)
 	for _, ip := range config.AllowedIPs {
 		allowedIPs[ip] = true
 	}
+	lastManualScan = time.Now().Add(time.Duration(-10) * time.Minute)
```
placer14: You don't need to initialize `lastManaualScan`. The zero value would make the check below pass: (https://play.golang.org/p/yHj9vObnYS-) `time.Since(time.Time{}).Minutes() > 10 // true`

repo/datastore.go
```diff
@@ -28,6 +28,7 @@ type Datastore interface {
 	Coupons() CouponStore
 	TxMetadata() TransactionMetadataStore
 	ModeratedStores() ModeratedStore
+	OrderMessages() OrderMessageStore
```
placer14: Can we make this more generic for all messages?

repo/db/order_messages.go
```diff
@@ -0,0 +1,92 @@
...
+	var (
+		msg0   []byte
+		peerID string
+	)
+
+	stmt, err := o.db.Prepare("select message, peerID from order_messages where messageID=?")
+	if err != nil {
+		return nil, "", err
+	}
+	err = stmt.QueryRow(fmt.Sprintf("%s-%d", orderID, mType)).Scan(&msg0, &peerID)
```
placer14: Let's find a better index which isn't type dependent...considering that we'd like to use this pattern for other message types. Maybe `messageID` could be a hash of the `other peerID`+`timestamp` concatenated together?

repo/db/order_messages.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return nil, "", err
+	}
+
+	msg := new(pb.Message)
```
placer14: We should not be passing around unwrapped pb.Message items so we can safely support multiple versions of Message (similar the problems we're facing with Listing and Contracts). If anything, create a thin `repo.Message` struct to handle the JSON Marshaling.

schema/constants.go
```diff
@@ -36,6 +36,7 @@ const (
 	CreateTableCouponsSQL                   = "create table coupons (slug text, code text, hash text);"
 	CreateIndexCouponsSQL                   = "create index index_coupons on coupons (slug);"
 	CreateTableModeratedStoresSQL           = "create table moderatedstores (peerID text primary key not null);"
+	CreateOrderMessagesSQL                  = "create table order_messages (messageID text primary key not null, orderID text, message_type integer, message blob, peerID text, url text, acknowledged bool, tries integer, created_at integer, updated_at integer);"
```
placer14: What is `url` for?

schema/constants.go
```diff
@@ -36,6 +36,7 @@ const (
...
 	// End SQL Statements
```
placer14: We should consider adding indices for the `Messages` table and using normalized data for the lookup. (ex: `where peerID = ? and orderType = ?`.

schema/constants.go
amangale: It is for IPFS URL. We might use the IPFS url (in future) or message blob to resend a message.

repo/migrations/Migration024.go
```diff
@@ -0,0 +1,92 @@
...
+	// AM06CreateIndexMessagesSQL2 - the messages composite index on orderID and messageType create sql
+	AM06CreateIndexMessagesSQL2 = "create index index_messages2 on messages (orderID, message_type);"
+	// AM06UpVer - set the repo Up version
+	AM06UpVer = "25"
+	// AM06DownVer - set the repo Down version
+	AM06DownVer = "24"
+)
+
+// AM06 - local migration struct
+type AM06 struct{}
```
placer14: I don't mind this abstraction away from `Migration024` but can we give it a more meaningful name? Maybe something self-descriptive like `MigrationCreateMessagesTable`?

repo/migrations/Migration024.go
```diff
@@ -0,0 +1,92 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+)
+
+const (
+	// AM06MessagesCreateSQL - the messages create sql
+	AM06MessagesCreateSQL = "create table messages (messageID text primary key not null, orderID text, message_type integer, message blob, peerID text, url text, acknowledged bool, tries integer, created_at integer, updated_at integer);"
+	// AM06CreateIndexMessagesSQL1 - the messages index on messageID sql
+	AM06CreateIndexMessagesSQL1 = "create index index_messages1 on messages (messageID);"
```
placer14: Please make these variable and index names meaningful. `index_messages_message_id`?

repo/migrations/Migration024.go
```diff
@@ -0,0 +1,92 @@
...
+	// AM06CreateIndexMessagesSQL2 - the messages composite index on orderID and messageType create sql
+	AM06CreateIndexMessagesSQL2 = "create index index_messages2 on messages (orderID, message_type);"
```
placer14: `index_messages_order_id_type`?

repo/migrations/Migration024.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Up - the migration Up code
+func (AM06) Up(repoPath, databasePassword string, testnetEnabled bool) error {
+	return createMessages(repoPath, databasePassword, AM06UpVer, testnetEnabled)
+}
+
+// Down - the migration Down code
+func (AM06) Down(repoPath, databasePassword string, testnetEnabled bool) error {
+	return createMessages(repoPath, databasePassword, AM06DownVer, testnetEnabled)
```
placer14: I was expecting to see the table being removed rather than created again.

repo/migrations/Migration024_test.go
```diff
@@ -0,0 +1 @@
+package migrations_test
```
placer14: I'd like to see a test that proves the table was created by adding and retrieving a full row with `Up()` and a that the table was removed on `Down()`.

repo/migrations/Migration024.go
placer14: And if we're removing the MessageByMessageID function, let's remove this index at the same time.

schema/constants.go
amangale: Three indices have been added: MessageID, <OrderID, MessageType> and <PeerID, MessageType>

repo/datastore.go
```diff
@@ -431,3 +432,17 @@ type WatchedScriptStore interface {
...
+	Queryable
+
+	// Save a new message
+	Put(messageID, orderID string, mType pb.Message_MessageType, peerID string, msg Message) error
+
+	// GetByOrderIDType returns the message for specified order and type
+	GetByOrderIDType(orderID string, mType pb.Message_MessageType) (*Message, string, error)
+
+	// GetByMessageID returns the message for specified message id
+	GetByMessageID(messageID string) (*Message, string, error)
```
placer14: This does not appear to be used at all? If not, let's remove this and the associated indices.

schema/constants.go
```diff
@@ -36,6 +36,10 @@ const (
 	CreateTableCouponsSQL                   = "create table coupons (slug text, code text, hash text);"
 	CreateIndexCouponsSQL                   = "create index index_coupons on coupons (slug);"
 	CreateTableModeratedStoresSQL           = "create table moderatedstores (peerID text primary key not null);"
+	CreateMessagesSQL                       = "create table messages (messageID text primary key not null, orderID text, message_type integer, message blob, peerID text, url text, acknowledged bool, tries integer, created_at integer, updated_at integer);"
+	CreateIndexMessagesSQL1                 = "create index index_messages1 on messages (messageID);"
+	CreateIndexMessagesSQL2                 = "create index index_messages2 on messages (orderID, message_type);"
+	CreateIndexMessagesSQL3                 = "create index index_messages2 on messages (peerID, message_type);"
```
placer14: I don't see a lookup which uses `CreateIndexMessagesSQL1` or `CreateIndexMessagesSQL3`. Remove?

placer14: If `CreateIndexMessagesSQL3` is not removed, please update the index name so it's not repeating the same from `CreateIndexMessagesSQL2`

repo/db/messages.go
```diff
@@ -91,8 +91,8 @@ func (o *MessagesDB) GetByOrderIDType(orderID string, mType pb.Message_MessageTy
 	return msg, peerID, nil
 }
 
-// GetByMessageIDType returns the message for the specified message id
-func (o *MessagesDB) GetByMessageIDType(messageID string) (*repo.Message, string, error) {
+// GetByMessageID returns the message for the specified message id
+func (o *MessagesDB) GetByMessageID(messageID string) (*repo.Message, string, error) {
```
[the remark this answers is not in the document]
amangale: This is for future use where only the messageID is used for fetching the message. One of the things being discussed is a logical grouping of messages and this method will be useful there.

repo/migrations/Migration024_test.go
amangale: The test of adding a row, finding it is in repo/db/messages_test. The presence of the table after Up and its absence after Down have been tested here.

repo/datastore.go
amangale: This method has been removed. The indices' future use has been discussed. It is beeter to have an index now when the table is being created rather than adding an index later which may cause issues.

repo/db/messages.go
```diff
@@ -0,0 +1,122 @@
...
+	modelStore
+}
+
+// NewMessageStore - return new MessagesDB
+func NewMessageStore(db *sql.DB, lock *sync.Mutex) repo.MessageStore {
+	return &MessagesDB{modelStore{db, lock}}
+}
+
+// Put - insert record into the messages
+func (o *MessagesDB) Put(messageID, orderID string, mType pb.Message_MessageType, peerID string, msg repo.Message) error {
```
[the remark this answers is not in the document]
placer14: What are your thoughts here, @amangale?

repo/datastore.go
placer14: I'm not sure what issues might arise from creating an index later (aside from the extra work of handling the migration, which I agree is annoying). Thanks for the explanation. Let's leave the index in place.

core/net.go
```diff
@@ -387,11 +434,23 @@ func (n *OpenBazaarNode) SendOrderCompletion(peerID string, k *libp2p.PubKey, co
...
-	return n.sendMessage(peerID, k, pb.Message{
+	}  
+	m := pb.Message{
 		MessageType: pb.Message_ORDER_COMPLETION,
 		Payload:     a,
-	})
+	}  
+	orderID0, err := n.CalcOrderID(completionMessage.BuyerOrder)
+	if err != nil {
+		log.Errorf("failed calculating order id")
```
placer14: @amangale Doing some testing around this code, I hit this error and the message never saved: https://github.com/OpenBazaar/openbazaar-go/pull/1584/files#diff-8c006b1217c14444730603927b6f7515R444

Please add the error output to the log: `log.Errorf("failed calculating order id: %s", err)`.

core/net.go
```diff
@@ -302,6 +315,17 @@ func (n *OpenBazaarNode) SendOrderConfirmation(peerID string, contract *pb.Ricar
 	if err != nil {
 		return err
 	}
+	orderID0, err := n.CalcOrderID(contract.BuyerOrder)
+	if err != nil {
+		log.Errorf("failed calculating order id")
```
placer14: Add err to output here, as well.

core/net.go
```diff
@@ -276,6 +277,18 @@ func (n *OpenBazaarNode) SendOrder(peerID string, contract *pb.RicardianContract
 		Payload:     pbAny,
 	}
 
+	orderID0, err := n.CalcOrderID(contract.BuyerOrder)
+	if err != nil {
+		log.Errorf("failed calculating order id")
```
placer14: Add err to output here, as well.

core/net.go
```diff
@@ -379,6 +415,17 @@ func (n *OpenBazaarNode) SendOrderFulfillment(peerID string, k *libp2p.PubKey, f
 		MessageType: pb.Message_ORDER_FULFILLMENT,
 		Payload:     a,
 	}
+	orderID0, err := n.CalcOrderID(fulfillmentMessage.BuyerOrder)
+	if err != nil {
+		log.Errorf("failed calculating order id")
```
placer14: Add err to output here, as well.

core/net.go
placer14: Looking at the call into this function, the `completionMessage.BuyerOrder` isn't populated which is causing this to fail. You may want to populate the BuyerOrder on the `completionMessage` in the calling function to fix this error. The `completionMessage` is created at `core/completion.go:244.

amangale: In place of `CalcOrderID`, I have used `orderID0 := fulfillmentMessage.VendorOrderFulfillment[0].OrderId`.

repo/db/messages.go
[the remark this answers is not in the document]
amangale: As of now the messageID generation is based on the orderID and messageType. This may not hold true forever. Hence, it is better to explicitly specify the messageID with the generation logic of messageID independent of Put.

api/jsonapi.go
```diff
@@ -1480,7 +1482,7 @@ func (i *jsonAPIHandler) GETProfile(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		if profile.PeerID != peerID {
-			ErrorResponse(w, http.StatusNotFound, "invalid profile: peer id mismatch on found profile")
+      ErrorResponse(w, http.StatusNotFound, "invalid profile: peer id mismatch on found profile")
```
placer14: Indent here is not consistent.

api/jsonapi.go
```diff
@@ -4220,3 +4222,76 @@ func (i *jsonAPIHandler) GETPost(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err = i.node.Service.SendMessage(ctx, p, &msg.Msg)
+	if err != nil {
+		// If send message failed, try sending offline message
+		log.Errorf("resending message failed: %v", err)
```
placer14: This is not an error (yet). Maybe `Warningf` instead.

api/jsonapi.go
```diff
@@ -4220,3 +4222,76 @@ func (i *jsonAPIHandler) GETPost(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+	SanitizedResponse(w, `{}`)
+}
+
+// GETScanOfflineMessages - used to manually trigger offline message scan
+func (i *jsonAPIHandler) GETScanOfflineMessages(w http.ResponseWriter, r *http.Request) {
+	t := time.Since(lastManualScan).Minutes()
+	if t < 100000000 {
```
placer14: It looks like `t > 10_000_000` only occurs when `lastManualScan` is zero? Why ignore this case?
If you're set on ignoring the zero value, we should use `time.IsZero` like so?
``` 
if lastManualScan.IsZero() {
  lastManualScan = time.Now()
} else {
  if time.Since(lastManualScan) >= 1 * time.Minute { // 1 * time.Minute should be a constant
    i.node.MessageRetriever.RunOnce()
  }
}
SanitizedResponse(w, `{}`)
```

PR #1602, opened 2019-05-24
Unit test for gencerts subcommand

cpacia:
Continuation of  #1241 in an ob-go branch and rebased with master.

### DIFF
--- a/cmd/gencerts_test.go
+++ b/cmd/gencerts_test.go
@@ -1,0 +1,108 @@
+package cmd_test
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/OpenBazaar/openbazaar-go/cmd"
+	"github.com/OpenBazaar/openbazaar-go/schema"
+)
+
+func buildCertDirectory() (string, func(), error) {
+	ctx := schema.SchemaContext{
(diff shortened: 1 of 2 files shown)


cmd/gencerts_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	config := cmd.GenerateCertificates{
+		DataDir:  dataDir,
+		Testnet:  true,
+		Host:     "127.0.0.1",
+		ValidFor: 1, // 1ns ... 1*1e9 == 1s
+	}
+
+	args := []string{"args"} // "Commander" interface args
```
placer14: Args aren't used for this command. We shouldn't suggest that they are meaningful for these tests by including this red-herring value.

cmd/gencerts_test.go
```diff
@@ -0,0 +1,108 @@
...
+	certSchema := schema.MustNewCustomSchemaManager(ctx)
+	if err := certSchema.BuildSchemaDirectories(); err != nil {
+		return "", nil, err
+	}
+	return ctx.DataPath, certSchema.DestroySchemaDirectories, nil
+}
+
+func TestGenCertsGenericDefaults(t *testing.T) {
+	dataDir, destroy, schemaErr := buildCertDirectory()
+	//defer destroy()
```
placer14: Don't forget to destroy on exit.

PR #1613, opened 2019-05-29
Use separate namespace for IPNS persistent cache

### LINKED ISSUE #1612: Use separate namespace for IPNS cache
Our ipfs package adds a persistent cache layer on top of IPNS. The data is stored in the ipfs db. However, we are currently having a namespace collision with the DHT package which uses the datastore to store some of it's records. 

We need to create a separate namespace for our cache.

cpacia:
Our IPNS package implements a separate persistent cache for IPNS records. The
format of the value stored in the database is different from the format used
by the DHT package. This commit changes the database key used by the persistent
cache to make sure it does not collide with the key used by the DHT package
which might otherwise cause records to be overriden with a different format.

closes #1612

### DIFF
--- a/ipfs/resolve_test.go
+++ b/ipfs/resolve_test.go
@@ -1,0 +1,79 @@
+package ipfs
+
+import (
+	"github.com/ipfs/go-ipfs/core/mock"
+	"github.com/ipfs/go-ipfs/namesys"
+	ipath "gx/ipfs/QmQAgv6Gaoe2tQpcabqwKXKChp2MZ7i3UXv9DqTTaxCaTR/go-path"
+	ipnspb "gx/ipfs/QmUwMnKKjH3JwGKNVZ3TcP37W93xzqNA4ECFFiMo6sXkkc/go-ipns/pb"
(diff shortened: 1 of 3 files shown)


ipfs/resolve_test.go
```diff
@@ -0,0 +1,79 @@
...
+	ipnspb "gx/ipfs/QmUwMnKKjH3JwGKNVZ3TcP37W93xzqNA4ECFFiMo6sXkkc/go-ipns/pb"
+	"gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer"
+	"gx/ipfs/QmddjPSGZb3ieihSseFeCfVRpZzcqczPNsD2DvarSwnjJB/gogo-protobuf/proto"
+	"testing"
+)
+
+// TestDatastoreCaching tests that the data can be inserted and retrieved from the
+// database using the IPNS namespace as well as our persistent cache namespace and
+// that the two do not conflict with each other.
+func TestDatastoreCaching(t *testing.T) {
```
placer14: I'm not in the habit of putting docs on our tests... if the test well-written, these comments will be redundant.

PR #1629, opened 2019-06-10
GET /ob/resolveipns/:peerid can perform self-lookups

### LINKED ISSUE #1601: GET /ob/resolveipns/:peerID should use own peerID if an argument is not passed
When calling `GET /ob/resolveipns/:peerID`, a user should expect the endpoint to use the local peerID in the lookup when one is not provided "as-is our pattern for other API endpoints".

placer14:
Closes #1601


cpacia:
Do you want to try it with the `SanitizedResponse`?

placer14:
@drwasho observes that the request doesn't work when you omit the peerID as an argument in the URL. Fixing to allow an empty request to assume you want your own IPNS record...

PR #1635, opened 2019-06-13
Fix bug starting node in regtest mode

cpacia:
Regtest routing does not use the API router so we need to extract the DHT from the tired router directly from the tiered router rather than the API router during startup. This comment makes the change to allow it to run in regtest mode.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -18,9 +18,7 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


placer14:
@cpacia, don't forget to update the description of the PR with the details of the problem, if this isn't going to be associated with an issue.

placer14:
This changeset has become stale. Closing until we can get back to it properly.

PR #1637, opened 2019-06-13
Merge fixes from v0.13.4

### LINKED ISSUE #1564: IPNS name publish failed
Regular startup of the daemon will produce the following log output after the OpenBazaar service starts:
```
13:08:05.066 [ERROR] [core/publish] name publish failed
```

### LINKED ISSUE #1593: Test generalized fix for "name publish failed" issues
Test general fix to delete cached private key from Namesys Datastore on-demand as a user-triggerable action. Will probably deliver this as a separate subcommand, but simple feasibility will be a modified build tailored specifically for a test user which always deletes the cached private key on startup.

This is a potential fix for #1564

### LINKED ISSUE #1612: Use separate namespace for IPNS cache

### LINKED ISSUE #1626: IPFS secio handshake patch likely not forward compatible
In the code segment below when establishing a new outgoing connection `s.remotePeer` is set to the peerID of the node we're trying to connect to. Ultimately this will either be an old style (hashed) peerID or a new style inline key depending on the actual ID that is used when we try to make a connection.

So far so good. Also `peer.IDFromPublicKey` is currently programmed to return an _old_ style (hashed) peerID since we set `AdvancedEnableInlining` to false.

OK . Now take a look at the secio handshake... this is their raw code without any modifications:

```
// get peer id
    actualRemotePeer, err := peer.IDFromPublicKey(s.remote.permanentPubKey)
    if err != nil {
        return err
    }
    switch s.remotePeer {
    case actualRemotePeer:
        // All good.
    case "":
        // No peer set. We're accepting a remote connection.
        s.remotePeer = actualRemotePeer
    default:
        // Peer mismatch. Bail.
        s.insecure.Close()
        log.Debugf("expected peer %s, got peer %s", s.remotePeer, actualRemotePeer)
        return ErrWrongPeer
    }
```

If we ever enable inline keys, old nodes which have not upgraded, when trying to connect to a new style peerID, will pass in a new style ID into `s.remotePeer` but `actualRemotePeer` will still be an old style key. Hence `s.remotePeer` != `actualRemotePeer` and the handshake will fail.

If an upgraded node tries to connect to a non-upgraded peerID then `peer.IDFromPublicKey` will return a _new_ style peerID while `s.remotePeer` is an _old_ style key. Hence it will fail to connect.

So to make our current release _forward_ compatible with a future release using inline keys I've done:

```
// get peer id
    actualRemotePeer, err := peer.IDFromPublicKey(s.remote.permanentPubKey)
    if err != nil {
        return err
    }
    switch s.remotePeer {
    case actualRemotePeer:
        // All good.
    case "":
        // No peer set. We're accepting a remote connection.
        s.remotePeer = actualRemotePeer
    default:
        pubkeyBytes, err := s.remote.permanentPubKey.Bytes()
        if err != nil {
            return err
        }
        oldMultihash, err := mh.Sum(pubkeyBytes, mh.SHA2_256, 32)
        if err != nil {
            return err
        }
        oldStylePeer, err := peer.IDB58Decode(oldMultihash.B58String())
        if err != nil {
            return err
        }
        if s.remotePeer != oldStylePeer {
            // Peer mismatch. Bail.
            s.insecure.Close()
            log.Debugf("expected peer %s, got peer %s", s.remotePeer, actualRemotePeer)
            return ErrWrongPeer
        }
    }
```

But looking at it, `s.remotePeer` would be an inline peerID and `oldStylePeer` would be an old style key, so I think this is screwed up and would prevent old nodes from connecting to new nodes. I think our current release should be a `newStylePeer` and compare it to `s.remotePeer` rather than an `oldStylePeer`.

And then a subsequent release with inline keys should use the code snippet above. (edited)

### LINKED ISSUE #1634: Investigate libp2p secio 0.0.3 for 0.13.4 release
From https://blog.ipfs.io/93-go-ipfs-0.4.21:

> ‼️ALL USERS MUST UPGRADE.‼️ We intended to introduce a feature this release that, unfortunately, reliably triggered this bug. To avoid partitioning the network, we’ve decided to postpone this feature for a release or two.

> Specifically, we’re going to provide a minimum one month upgrade period. After that, we’ll start testing the impact of deploying the proposed changes.

>:point_right: If you’re running the mainline go-ipfs, please upgrade ASAP. If you’re building a separate app or working on a forked go-ipfs, make sure to upgrade github.com/libp2p/go-libp2p-secio to at least v0.0.3.

### LINKED ISSUE #1645: Squelch MODERATOR_ADD and MODERATOR_REMOVE notifications
In order to removed the noise of notifications from stores adding/removing moderators from their store, the go daemon should prevent moderator node from seeing notifications from stores when adding and removing the node as a moderator for their store.

placer14:
This is a placeholder PR to ensure that changes from the final v0.13.4 release are merged back into master.

Fixes #1564
Fixes #1593 
Fixes #1612
Fixes #1626 
Fixes #1634
Fixes #1645

### DIFF
--- a/vendor/gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer/peer.go
+++ b/vendor/gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer/peer.go
@@ -169,8 +169,21 @@ func IDFromPublicKey(pk ic.PubKey) (ID, error) {
 	return ID(hash), nil
 }
 
-// OpenBazaar: temporary helper function to remain forward compatible with
-// inline keys
+// HashedIDFromPublicKey will always return the SHA256 hash of
+// the pubkey bytes. OpenBazaar: temporary helper to isolate the
+// hash-producing ID behavior.
+func HashedIDFromPublicKey(pk ic.PubKey) (ID, error) {
+	b, err := pk.Bytes()
+	if err != nil {
+		return "", err
+	}
+	hash, _ := mh.Sum(b, mh.SHA2_256, -1)
+	return ID(hash), nil
+}
+
+// InlineIDFromPublicKey will always return the new inline ID format
+// of the pubkey bytes.  OpenBazaar: temporary helper function to
+// remain forward compatible with inline keys
 func InlineIDFromPublicKey(pk ic.PubKey) (ID, error) {
 	b, err := pk.Bytes()
 	if err != nil {
@@ -178,6 +191,16 @@ func InlineIDFromPublicKey(pk ic.PubKey) (ID, error) {
 	}
 	hash, _ := mh.Sum(b, mh.ID, -1)
 	return ID(hash), nil
+}
+
+// AlternativeIDFromPublicKey returns SHA256 hash ID when AdvancedEnableInlining
+// is true, and returns new InlineID otherwise. This allows legacy IDs to be compared
+// after they are no longer available by the default IDFromPublicKey function.
+func AlternativeIDFromPublicKey(pubkey ic.PubKey) (ID, error) {
+	if AdvancedEnableInlining {
+		return HashedIDFromPublicKey(pubkey)
+	}
+	return InlineIDFromPublicKey(pubkey)
 }
 
 // IDFromPrivateKey returns the Peer ID corresponding to sk
--- a/vendor/gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer/peer.go
+++ b/vendor/gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer/peer.go
@@ -165,37 +165,51 @@ func IDFromPublicKey(pk ic.PubKey) (ID, error) {
 	if AdvancedEnableInlining && len(b) <= maxInlineKeyLength {
 		alg = mh.ID
 	}
-	hash, _ := mh.Sum(b, alg, -1)
+	// OpenBazaar: handle the Sum error
+	hash, err := mh.Sum(b, alg, -1)
+	if err != nil {
+		return "", err
+	}
 	return ID(hash), nil
 }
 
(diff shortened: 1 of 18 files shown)


vendor/gx/ipfs/QmYVXrKrKHDC9FobgmcmshCDyWwdrfwfanNQN4oxJ9Fk3h/go-libp2p-peer/peer.go
```diff
@@ -169,8 +169,21 @@ func IDFromPublicKey(pk ic.PubKey) (ID, error) {
...
-// inline keys
+// HashedIDFromPublicKey will always return the SHA256 hash of
+// the pubkey bytes. OpenBazaar: temporary helper to isolate the
+// hash-producing ID behavior.
+func HashedIDFromPublicKey(pk ic.PubKey) (ID, error) {
+	b, err := pk.Bytes()
+	if err != nil {
+		return "", err
+	}
+	hash, _ := mh.Sum(b, mh.SHA2_256, -1)
```
tyler-smith: This error shouldn't be swallowed

cpacia: Only thing I would prefer in these three patches to peer is for the comment to start with `// OpenBazaar:` so that we can find them later with a search.

placer14: @cpacia The line includes `OpenBazaar:`. Is that sufficient here?

placer14: Fixed the error handling.

cpacia: Probably sufficient. I would maybe also add it to the functions that were added but at least the change is noted.

PR #1642, opened 2019-06-14
Animated GIF Support

hoffmabc:
Some exploration of animated GIF support within our existing image handling.
I know this may step on some toes @placer14 since you're looking at close code.

### DIFF
--- a/core/images.go
+++ b/core/images.go
@@ -4,13 +4,12 @@ import (
 	"bytes"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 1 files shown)


hoffmabc:
Currently we treat all image types as outputting to JPEG. This will detect if it's a GIF and try to manage the resizing directly. If it's an animated GIF it will loop through each frame and resize it and re-assemble the GIF and encode back to GIF format. 

It also will not try to upscale the GIF so it will do bounds checking to make sure it's not trying to upscale.

placer14:
I need to review the `lilliput` dep more thoroughly before I look at resolving the conflicts above.

PR #1643, opened 2019-06-14
Allow EXIF data to be considered when uploading images

### LINKED ISSUE #1630: Processing input via POST /ob/images causes images to become rotated
Report from @drwasho indicates that the `POST /ob/images` endpoint either persists the rotated image or persists the original image yet returns a rotated image. Whichever the case, the response should be consistent with the original image.

placer14:
Fixes #1630


drwasho:
Functionally tested this and it's working very nicely.

PR #1646, opened 2019-06-24
[#1645] Prevent notifications produced on MODERATOR_ADD and REMOVE

### LINKED ISSUE #1645: Squelch MODERATOR_ADD and MODERATOR_REMOVE notifications

placer14:
Fixes #1645


placer14:
Quick testing suggests the missing moderator notifications don't hurt the client experience. Would need to produce a bunch of notifications to see what volume notifications might do when mod requests are being omitted...

PR #1653, opened 2019-07-15
[WIP] Number precision changes 

cpacia:


drwasho:
Number precision changes will be separated from the ETH integration and merged first in #1666

PR #1654, opened 2019-07-15
Order payment network message

amangale:
This PR extracts the ORDER_PAYMENT network message code from the existing ethereum branch to be merged independently.


core/spend.go
```diff
@@ -37,6 +37,7 @@ type SpendResponse struct {
 	Timestamp          time.Time `json:"timestamp"`
 	Txid               string    `json:"txid"`
 	UnconfirmedBalance int64     `json:"unconfirmedBalance"`
+	PeerID             string    `json:"-"`
```
cpacia: I don't see this used anywhere in the PR. What's the purpose of this?

net/service/handlers.go
```diff
@@ -1486,6 +1558,102 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
...
+		if !(contract.VendorListings[0].VendorID.PeerID ==
+			service.node.IpfsNode.Identity.Pretty()) {
+			log.Debugf("mismatched peerID. wrong node is processing : orderID: %s, contractPeerID: %s",
+				paymentDetails.OrderID, contract.VendorListings[0].VendorID.PeerID)
+			return nil, errors.New("mismatched peer id")
+		}
+	}
+
+	toAddress, _ := wal.DecodeAddress(txn.ToAddress)
+	//fromAddress, _ := wal.DecodeAddress(txn.FromAddress)
```
cpacia: Make sure you removed the commented out lines

net/service/handlers.go
```diff
@@ -1486,6 +1558,102 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
...
+	output := wallet.TransactionOutput{
+		Address: toAddress,
+		Value:   txn.Value,
+		Index:   1,
```
cpacia: Make sure the index is correct here. I don't think it's always `1`. If it's incorrect we will error when spending from this transaction.

net/service/handlers.go
```diff
@@ -1486,6 +1558,102 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
...
+		OrderID: paymentDetails.OrderID,
+	}
+
+	input := wallet.TransactionInput{}
+
+	if paymentDetails.WithInput {
```
cpacia: Ideally this would be the correct number of inputs with the correct input for each although I don't believe we check the inputs when processing an incoming payment (we only check the outputs).

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -472,7 +472,7 @@ func (wallet *EthereumWallet) Spend(amount int64, addr btcutil.Address, feeLevel
 			// but valid txn like some contract condition causing revert
 			if rcpt.Status > 0 {
 				// all good to update order state
-				go wallet.callListeners(wallet.createTxnCallback(hash.Hex(), referenceID, amount, time.Now()))
+				go wallet.AssociateTransactionWithOrder(wallet.createTxnCallback(hash.Hex(), referenceID, amount, time.Now()))
```
placer14: I appreciate the `wallet.createTxnCallback` helper. I see in every invocation that the timestamp is always `time.Now()`. If you don't anticipate manually setting this to another value, you might internalize this detail instead of asking for it as a param.

net/service/handlers.go
```diff
@@ -1486,6 +1558,102 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
...
+			LinkedAddress: toAddress,
+			Value:         txn.Value,
+			OrderID:       paymentDetails.OrderID,
+		}
+	}
+
+	cb := wallet.TransactionCallback{
+		Txid:      txn.Txid,
+		Outputs:   []wallet.TransactionOutput{output},
+		Inputs:    []wallet.TransactionInput{input},
```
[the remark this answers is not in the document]
cpacia: Inputs would represents `spends` leaving the multisig. Since this tx is funds going into the multisig it should not hurt anything if the inputs are omitted.

net/service/handlers.go
```diff
@@ -1486,6 +1558,102 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
...
+		Height:    1,
```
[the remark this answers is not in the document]
cpacia: yeah height should be 0.

vendor/github.com/OpenBazaar/multiwallet/bitcoin/wallet.go
```diff
@@ -213,6 +214,31 @@ func (w *BitcoinWallet) Transactions() ([]wi.Txn, error) {
...
+		if err != nil {
+			return txn, err
+		}
+		outs := []wi.TransactionOutput{}
+		for i, out := range tx.TxOut {
+			_, addrs, _, err := txscript.ExtractPkScriptAddrs(out.PkScript, w.params)
+			if err != nil {
+				return txn, err
+			}
+			if len(addrs) == 0 {
```
cpacia: This code in each of the wallet implementations has a small intentional of putting us in an unrecoverable state. For example, suppose sender of this transaction included a change output that was _not_ of a standard address type. The change output is for their wallet, not ours, so technically this would still be a valid transaction on the network. However, our code would error here. Either where `err != nil` or where `len(addrs) == 0`.

So I think it's probably best to just set the address to "" in this case rather than error. Maybe log a warning as well.

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
amangale: As of now, I have set it to be time.Now(). However, this can change later if need be to actual block time.

core/spend.go
amangale: The peerID is used to create the ORDER_PAYMENT message when the spend returns.

net/service/handlers.go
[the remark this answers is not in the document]
cpacia: +1 forgot about this one. Make sure you change this @amangale

placer14:
I also don't see any godep updates at all. @cpacia if we're going to completely ignore that file, we should probably just remove it? I don't godep has any more use to us at this point. But that leaves the concern that we aren't managing the vendors folder properly which is greatly concerning. I don't know how we are ensuring that these commits are being included in the dependent libs.

net/service/handlers.go
```diff
@@ -1486,6 +1558,102 @@ func (service *OpenBazaarService) handleStore(pid peer.ID, pmes *pb.Message, opt
...
+	}
+
+	if contract.VendorOrderConfirmation != nil &&
+		contract.BuyerOrder.Payment.Method != pb.Order_Payment_MODERATED {
+
+		// the seller has confirmed the order, so a simple check of
+		// the addresses and we are good to proceed
+		if !u.AreAddressesEqual(contract.VendorOrderConfirmation.PaymentAddress, txn.ToAddress) {
+			log.Debugf("mismatched payment address details: orderID: %s, expectedAddr: %s, actualAddr: %s",
+				paymentDetails.OrderID, contract.VendorOrderConfirmation.PaymentAddress, txn.ToAddress)
```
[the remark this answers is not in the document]
amangale: Well if you are asking for a use case/scenario, then I have none. It is a check I put in to double check before proceeding. In all probability of this happening is quite small since orderID is used for creating the ORDER_PAYMENT message in the first place and the code would not reach here if the order was not found.
 The only real recovery is for the buyer to realize that the payment has been made to another address which does not match the address specified in the contract.

cpacia:
Final TODO @amangale can you send a PR to to the upstream repos with the changes that are being included in vendor? Thanks

net/service/handlers.go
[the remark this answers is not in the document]
placer14: @amangale will raise with @drwasho what follow up work needs to be done when the order lands in this state... currently we don't recover from this case and would be nice to have a plan before resolving this PR.

placer14:
Great work, @amangale, keeping up with the critiques. 🍻 Thanks!

net/service/handlers.go
placer14: Decided to refactor this as part of a refactor/redesign.

PR #1655, opened 2019-07-16
[EXPERIMENTAL] Restart method for mobile

hoffmabc:
This does two things:

1) Makes the gateway listener global so we can make sure to kill it too from the mobile stop
2) Creates a restart method that handles the stop/start scenario with one command for mobile


drwasho:
This PR can be closed in favor of #1656

PR #1656, opened 2019-07-16
Allow mobile node to be restarted

cpacia:
This PR adds a restart method to the mobile node.

### DIFF
--- a/api/gateway.go
+++ b/api/gateway.go
@@ -1,13 +1,12 @@
 package api
 
 import (
-	"net"
-	"net/http"
-
 	"github.com/OpenBazaar/openbazaar-go/core"
 	"github.com/OpenBazaar/openbazaar-go/schema"
 	"github.com/ipfs/go-ipfs/core/corehttp"
 	"github.com/op/go-logging"
+	"net"
+	"net/http"
 )
 
 var log = logging.MustGetLogger("api")
--- a/mobile/cmd/main.go
+++ b/mobile/cmd/main.go
@@ -6,6 +6,7 @@ import (
 	"github.com/jessevdk/go-flags"
 	"os"
 	"sync"
+	"time"
 )
 
 type Options struct {
@@ -45,6 +46,10 @@ func main() {
 	if err := n.Start(); err != nil {
 		fmt.Println(err.Error())
 	}
+
+	time.Sleep(time.Second*30)
+	fmt.Println("restarting...")
+	go n.Restart()
 
 	wg.Add(1)
 	wg.Wait()
--- a/mobile/cmd/main.go
+++ b/mobile/cmd/main.go
@@ -47,8 +47,8 @@ func main() {
 		fmt.Println(err.Error())
 	}
 
-	time.Sleep(time.Second*30)
-	fmt.Println("restarting...")
+	time.Sleep(time.Second*10)
+	fmt.Println("restarting...", time.Now())
 	go n.Restart()
 
 	wg.Add(1)
--- a/mobile/cmd/main.go
+++ b/mobile/cmd/main.go
@@ -2,11 +2,10 @@ package main
 
 import (
 	"fmt"
-	"os"
-	"sync"
-
 	"github.com/OpenBazaar/openbazaar-go/mobile"
 	"github.com/jessevdk/go-flags"
+	"os"
+	"sync"
 )
 
 type Options struct {
@@ -43,9 +42,10 @@ func main() {
 	if err != nil {
 		fmt.Println(err.Error())
 	}
-	wg.Add(1)
 	if err := n.Start(); err != nil {
 		fmt.Println(err.Error())
 	}
+
+	wg.Add(1)
 	wg.Wait()
 }
--- a/mobile/cmd/main.go
+++ b/mobile/cmd/main.go
@@ -47,7 +47,7 @@ func main() {
 		fmt.Println(err.Error())
 	}
 
-	time.Sleep(time.Second*10)
+	time.Sleep(time.Second * 10)
 	fmt.Println("restarting...", time.Now())
 	go n.Restart()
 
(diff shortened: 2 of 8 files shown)


api/gateway.go
```diff
@@ -1,13 +1,12 @@
...
 import (
-	"net"
-	"net/http"
-
 	"github.com/OpenBazaar/openbazaar-go/core"
 	"github.com/OpenBazaar/openbazaar-go/schema"
 	"github.com/ipfs/go-ipfs/core/corehttp"
 	"github.com/op/go-logging"
+	"net"
+	"net/http"
```
placer14: Can you remove this no-op diff?

mobile/cmd/main.go
```diff
@@ -2,11 +2,11 @@ package main
...
 import (
 	"fmt"
-	"os"
-	"sync"
-
 	"github.com/OpenBazaar/openbazaar-go/mobile"
 	"github.com/jessevdk/go-flags"
+	"os"
+	"sync"
+	"time"
```
placer14: Let's keep the packages organized consistently with newlines between each group:
```
import (
  // stdlib_packages

  // external_packages
 
  // gx_packages
)
```

mobile/node.go
```diff
@@ -295,7 +297,9 @@ func NewNodeWithConfig(config *NodeConfig, password string, mnemonic string) (*N
 	ncfg := ipfs.PrepareIPFSConfig(r, ignoredURI, config.Testnet, config.Testnet)
 	ncfg.Routing = constructMobileRouting
 
-	return &Node{OpenBazaarNode: core.Node, config: *config, ipfsConfig: ncfg, apiConfig: apiConfig}, nil
+	core.Node.PublishLock.Lock()
+
+	return &Node{OpenBazaarNode: core.Node, config: *config, ipfsConfig: ncfg, apiConfig: apiConfig, mtx: sync.Mutex{}}, nil
```
placer14: Can we avoid using the global node and only use this Node instance? I'm concerned we're using a global object state which does not match this instance state? (Why do we have a global `core.Node` anyway?)

mobile/node.go
```diff
@@ -338,10 +342,18 @@ func (n *Node) startIPFSNode(repoPath string, config *ipfscore.BuildCfg) (*ipfsc
...
+
+	return n.start()
+}
+
+func (n *Node) start() error {
 	nd, ctx, err := n.startIPFSNode(n.config.RepoPath, n.ipfsConfig)
 	if err != nil {
 		return err
 	}
+	obNode := n.OpenBazaarNode
```
placer14: This change doesn't appear to do anything for us? Can we revert?

mobile/node.go
```diff
@@ -64,6 +65,9 @@ type Node struct {
 	cancel         context.CancelFunc
 	ipfsConfig     *ipfscore.BuildCfg
 	apiConfig      *apiSchema.APIConfig
+	gateway        *api.Gateway
+	started        bool
+	mtx            sync.Mutex
```
placer14: Can you make the fieldname more descriptive? `startupMutex` or similar?

mobile/node.go
```diff
@@ -404,78 +416,121 @@ func (n *Node) Start() error {
...
 	return nil
 }
 
 // Stop stop openbazaard
-func (n *Node) Stop() error {
+func (n *Node) Stop() {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+
+	n.stop()
```
placer14: `return n.stop()`

mobile/node.go
```diff
@@ -404,78 +416,121 @@ func (n *Node) Start() error {
...
+}
+
+func (n *Node) stop() error {
+	//n.cancel()
```
placer14: Remove commented code.

mobile/node.go
```diff
@@ -404,78 +416,121 @@ func (n *Node) Start() error {
...
+	n.started = false
 	return nil
 }
 
+func (n *Node) Restart() error {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+
+	if n.started {
+		n.stop()
```
placer14: Handle error returned from `n.stop()`.

mobile/node.go
```diff
@@ -529,5 +584,5 @@ func newHTTPGateway(node *core.OpenBazaarNode, ctx commands.Context, authCookie
 	}
 
 	// Create and return an API gateway
-	return api.NewGateway(node, authCookie, manet.NetListener(gwLis), config, mainLoggingBackend, opts...)
+	return api.NewGateway(node.OpenBazaarNode, authCookie, manet.NetListener(gwLis), config, mainLoggingBackend, opts...)
```
placer14: I don't think we should be using the global node here. When we kill it or try to use any of the resources, it'll fail as soon as we start using a new node.

mobile/cmd/main.go
cpacia: Doesn't seem to be anyway around this. Gofmt always moves it into this position.

mobile/node.go
cpacia: It was a hack to get access to the lock from the main package which detects the sigint shutdown.

cpacia: We can fix it here but will need to be fixed in the main package before getting rid of it for the desktop.

mobile/node.go
cpacia: What do you mean by global node? This is an instance of it here, not a global.

api/gateway.go
cpacia: done

mobile/cmd/main.go
placer14: It will if you include the line breaks as shown above.

mobile/node.go
placer14: You're passing in this, which is global. (https://github.com/OpenBazaar/openbazaar-go/blob/mobile-stop/core/core.go#L105) I don't think we should be managing the running OB instance as a global variable.

placer14: Sorry, you're right. I was confusing it for what you were doing in the mobile package (which I see you've fixed.) Disregard all of this.

PR #1667, opened 2019-07-30
Re-send missed ORDER messages on processing failure

### LINKED ISSUE #1069: The buyers order state doesn't update after the vendor fulfills
This is a port of https://github.com/OpenBazaar/openbazaar-desktop/issues/1505. The main issue is that the buyer can't complete the order after the vendor fulfills because the order state remains `AWAITING_FULFILLMENT` even though the order does get a `vendorOrderFulfillment` data object.

The JSON of the order is included in the referenced client bug.

placer14:
This is some code to explore how the resending of missed ORDER messages might work. This PR will serve as a means for discussion on approach. If everyone feels good about it and I don't see further bugs in my tests, I'll cherry-pick these into a cleaner history and prepare a formal PR to merge.

Proposed to fix #1069


core/disputes.go
```diff
@@ -308,6 +308,9 @@ func (n *OpenBazaarNode) ProcessDisputeOpen(rc *pb.RicardianContract, peerID str
 		// Load out version of the contract from the db
 		myContract, state, _, records, _, _, err := n.Datastore.Sales().GetByOrderId(orderID)
 		if err != nil {
+			if err := n.SendProcessingError(DisputerID, orderID, pb.Message_DISPUTE_OPEN, nil); err != nil {
+				log.Errorf("failed sending ORDER_PROCESSING_FAILURE to peer (%s): %s", DisputerID, err)
+			}
```
amangale: If a order is not present in the node db, does it mean there is a processing error.
A typical scenario where the seller was offline and there are 2 offline messages awaiting processing - ORDER and ORDER_PAYMENT and ORDER_PAYMENT was processed first, then it not really any order processing error. 
It would be better if the processing error was flagged and checked after all the offline messages were processed before send a processing error to peer.

core/disputes.go
```diff
@@ -365,6 +368,9 @@ func (n *OpenBazaarNode) ProcessDisputeOpen(rc *pb.RicardianContract, peerID str
 			return err
 		}
 		if state == pb.OrderState_AWAITING_PAYMENT || state == pb.OrderState_AWAITING_FULFILLMENT || state == pb.OrderState_PARTIALLY_FULFILLED || state == pb.OrderState_PENDING {
+			if err := n.SendProcessingError(DisputerID, orderID, pb.Message_DISPUTE_OPEN, myContract); err != nil {
+				log.Errorf("failed sending ORDER_PROCESSING_FAILURE to peer (%s): %s", DisputerID, err)
+			}
```
amangale: If there was a way to request a specific message from the buyer or seller then it can be used here.
When a node receives DISPUTE_OPEN and it finds itself in the above state, then it should simply request the appropriate message from the peer. ProcessingError message should be sent in case an order reaches an unrecoverable or unstable state on a node. The situation here can be remedied by simply requesting a message.

core/disputes.go
placer14: In the offline case, all available offline messages are downloaded in bulk and processed in a deterministic order (to ensure that this failure is, in fact, due to the message being unavailable).

core/disputes.go
placer14: That is sort of what the ProcessingError is meant to do. But instead of saying, only send XYZ message, send ALL of the messages that you have that I don't upto and including XYZ.

api/jsonapi.go
```diff
@@ -4240,47 +4240,11 @@ func (i *jsonAPIHandler) POSTResendOrderMessage(w http.ResponseWriter, r *http.R
...
-		GetByOrderIDType(args.OrderID, pb.Message_MessageType(msgType))
-	if err != nil || msg == nil || msg.Msg.GetPayload() == nil {
-		ErrorResponse(w, http.StatusBadRequest, "order message not found")
-		return
-	}
-
-	p, err := peer.IDB58Decode(peerID)
-	if err != nil {
-		ErrorResponse(w, http.StatusBadRequest, "invalid peer id")
+	if err := i.node.ResendCachedOrderMessage(args.OrderID, args.MessageType); err != nil {
```
amangale: `args.MessageType` is name string. Something like:
```
	var msgType int32
	var ok bool

	if msgType, ok = pb.Message_MessageType_value[args.MessageType]; !ok {
		ErrorResponse(w, http.StatusBadRequest, "invalid order message type")
		return
	}
```
 is required before the call to `ResendCachedOrderMessage` is made.

api/jsonapi.go
```diff
@@ -4240,47 +4240,17 @@ func (i *jsonAPIHandler) POSTResendOrderMessage(w http.ResponseWriter, r *http.R
...
-	if msgType, ok = pb.Message_MessageType_value[args.MessageType]; !ok {
-		ErrorResponse(w, http.StatusBadRequest, "invalid order message type")
-		return
-	}
-
-	msg, peerID, err := i.node.Datastore.Messages().
-		GetByOrderIDType(args.OrderID, pb.Message_MessageType(msgType))
-	if err != nil || msg == nil || msg.Msg.GetPayload() == nil {
-		ErrorResponse(w, http.StatusBadRequest, "order message not found")
+	msgInt, ok := pb.Message_MessageType_value[args.MessageType]
```
cpacia: Should we ucase this so the client doesn't have to worry about case?

core/disputes.go
```diff
@@ -308,6 +308,9 @@ func (n *OpenBazaarNode) ProcessDisputeOpen(rc *pb.RicardianContract, peerID str
 		// Load out version of the contract from the db
```
cpacia: maybe fix type while you're in this part of the code

cpacia:
LGTM

Issue #1483, opened 2019-03-05
Extract GETOrder behavior into unit testable method


placer14:
The behavior within GETOrder is complex and untestable while embedded within the API. Need to extract behavior into (perhaps) `core/order.go` as `func (*OpenBazaarNode) GetOrder(orderID string) (*repo.Order, error)` and unit test the output state of `*repo.Order` for various valid states of Order. (With particular care to validate this bug does not regress: https://github.com/OpenBazaar/openbazaar-go/issues/1481)

PR #1682, opened 2019-08-20
Extract GETOrder behavior into unit testable method #1483

ozamiatin:
Extracted method core.GetOrder()
Added unit test for it

### DIFF
--- a/core/order_test.go
+++ b/core/order_test.go
@@ -384,7 +384,7 @@ func TestOpenBazaarNode_GetOrder(t *testing.T) {
 	}
 
 	if orderResponse.State != state {
-		t.Fatal(fmt.Errorf("expected attribute description to be %s, but was %s",
+		t.Fatal(fmt.Errorf("expected order state to be %s, but was %s",
 			pb.OrderState_name[int32(state)],
 			pb.OrderState_name[int32(orderResponse.State)]))
 	}
--- a/core/order_test.go
+++ b/core/order_test.go
@@ -1,14 +1,14 @@
 package core_test
 
 import (
-	"github.com/OpenBazaar/openbazaar-go/test/factory"
-	"github.com/go-errors/errors"
-	"testing"
-
 	"github.com/OpenBazaar/openbazaar-go/core"
 	"github.com/OpenBazaar/openbazaar-go/pb"
 	"github.com/OpenBazaar/openbazaar-go/test"
+	"github.com/OpenBazaar/openbazaar-go/test/factory"
 	"github.com/golang/protobuf/proto"
+
+	"fmt"
+	"testing"
 )
 
 func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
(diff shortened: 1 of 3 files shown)


core/order_test.go
```diff
@@ -356,3 +358,27 @@ func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
...
+
+func TestOpenBazaarNode_GetOrder(t *testing.T) {
+	node, err := test.NewNode()
+	if err != nil {
+		t.Error(err)
+	}
+
+	contract := factory.NewContract()
+	orderID, err := node.CalcOrderID(contract.BuyerOrder)
+	err = node.Datastore.Purchases().Put(orderID, *contract, pb.OrderState_AWAITING_PAYMENT, false)
```
placer14: Handle these errors.

core/order_test.go
```diff
@@ -356,3 +358,27 @@ func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
 		t.Error("Calculated wrong order total")
 	}
 }
+
+func TestOpenBazaarNode_GetOrder(t *testing.T) {
+	node, err := test.NewNode()
+	if err != nil {
+		t.Error(err)
```
placer14: When a test dependency fails, we should abort the rest of the test run (as we already know it can't pass, if the dep fails to be produced). Would prefer `t.Fatal(err)` instead which would ensure the test is exited/returned after the error state is reported.

core/order_test.go
```diff
@@ -356,3 +358,27 @@ func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
...
+	}
+
+	if orderResponse.Funded {
+		t.Error(errors.New("Should not be funded"))
+	}
+
+	if orderResponse.State != pb.OrderState_AWAITING_PAYMENT {
+		t.Error(errors.New("Expected state OrderState_AWAITING_PAYMENT"))
```
placer14: It might be helpful to indicate what `orderResponse.State` actually was. Something like `expected attribute description to be (X), but was (Y)` would be ideal.

core/order_test.go
```diff
@@ -1,6 +1,8 @@
 package core_test
 
 import (
+	"github.com/OpenBazaar/openbazaar-go/test/factory"
+	"github.com/go-errors/errors"
```
placer14: Organize imports like so:
```
import (
  // all 3rd party library imports

  // all stdlib imports
)
```
(with the newline separating the groups)

core/order_test.go
```diff
@@ -356,3 +358,27 @@ func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
...
+		t.Error(err)
+	}
+
+	contract := factory.NewContract()
+	orderID, err := node.CalcOrderID(contract.BuyerOrder)
+	err = node.Datastore.Purchases().Put(orderID, *contract, pb.OrderState_AWAITING_PAYMENT, false)
+
+	orderResponse, err := node.GetOrder(orderID)
+	if err != nil {
+		t.Error(err)
```
placer14: `t.Fatal(err)`

core/order_test.go
```diff
@@ -356,3 +358,27 @@ func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
...
 	}
 }
+
+func TestOpenBazaarNode_GetOrder(t *testing.T) {
+	node, err := test.NewNode()
+	if err != nil {
+		t.Error(err)
+	}
+
+	contract := factory.NewContract()
```
placer14: We should manually set the state of the `contract` object instead of depending on the factory to never change (as it might). Please manually define the `Funded` and `State` fields which are checked below.

placer14:
PS: Don't worry about the coveralls CI failure. It always fails for external contributions.

core/order_test.go
ozamiatin: done

core/order_test.go
ozamiatin: done

core/order_test.go
ozamiatin: done

core/order_test.go
ozamiatin: done

core/order_test.go
ozamiatin: done

core/order_test.go
ozamiatin: done

ozamiatin:
Thank you very much for the review @placer14

core/order_test.go
```diff
@@ -356,3 +358,34 @@ func TestOpenBazaarNode_CalculateOrderTotal(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	orderResponse, err := node.GetOrder(orderID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if orderResponse.State != state {
+		t.Fatal(fmt.Errorf("expected attribute description to be %s, but was %s",
```
placer14: Sorry, that was a template. "attribute description" here is suppose to be "order state".

ozamiatin: done

placer14:
Thanks @ozamiatin. If you'd like to continue, here are some good next steps: https://github.com/OpenBazaar/openbazaar-go/issues/1683

ozamiatin:
Great, thanks a lot! Will look at #1683

PR #1688, opened 2019-08-26
Update offline message sending and receiving. 

cpacia:
This PR makes a change to send offline messages to the push nodes in parallel to publishing to the DHT. 

It also updates the message retriever to download messages from the push nodes in parallel to the DHT as well. 

Open questions:
- This change to the retriever that processes the messages in parallel potentially create duplicate message processing. Afaict it only results in a log saying duplicate message when putting to the offline message table, and the rest of our handlers *should* be able to handle duplicates. But is there a better way to do this that doesn't entail rewriting the file?

- Is there any way we can think of to test this class given all the objects it's instantiated with?

### DIFF
--- a/net/retriever/retriever.go
+++ b/net/retriever/retriever.go
@@ -96,7 +96,7 @@ func NewMessageRetriever(cfg MRConfig) *MessageRetriever {
 		WaitGroup:  new(sync.WaitGroup),
 	}
 
-	mr.Add(1)
+	mr.Add(2)
 	return &mr
 }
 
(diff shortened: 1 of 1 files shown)


drwasho:
Fetching listings via IPFS appears to take longer with this change, at least here in Australia.

hoffmabc:
@cpacia we need to strip the receiving functionality of this PR into a separate one and then we can merge this as a change to sending ONLY.

hoffmabc:
Let's rebase this to ethereum-master before moving forward.

PR #1695, opened 2019-09-04
Consider node state when returning NormalizedCurrencyCode

### LINKED ISSUE #1694: Currencies added to a profile are mainnet even when creating the profile on testnet
When you POST (create) a profile, the server appends a currency list based on the currently supported wallet currencies.

Previously, ob-go had correctly been putting testnet codes into the list when the app is run on testnet and mainnet otherwise.

The current `ethereum-master` branch is putting in mn codes regardless if we're running the app on mn or tn.

(I think the normalizeCurrency function and subsequent lookup just aren't `testnet/mainnet` aware)

placer14:
Fixes #1694

### DIFF
--- a/core/utils.go
+++ b/core/utils.go
@@ -158,11 +158,14 @@ func (n *OpenBazaarNode) BuildTransactionRecords(contract *pb.RicardianContract,
 }
 
 // NormalizeCurrencyCode standardizes the format for the given currency code
-func NormalizeCurrencyCode(currencyCode string) string {
+func (n *OpenBazaarNode) NormalizeCurrencyCode(currencyCode string) string {
 	var c, err = repo.LoadCurrencyDefinitions().Lookup(currencyCode)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


cpacia:
LGTM

core/utils.go
```diff
@@ -158,12 +158,15 @@ func (n *OpenBazaarNode) BuildTransactionRecords(contract *pb.RicardianContract,
...
 
 // NormalizeCurrencyCode standardizes the format for the given currency code
-func NormalizeCurrencyCode(currencyCode string) string {
+func (n *OpenBazaarNode) NormalizeCurrencyCode(currencyCode string) string {
 	var c, err = repo.LoadCurrencyDefinitions().Lookup(currencyCode)
 	if err != nil {
 		log.Errorf("invalid currency code (%s): %s", currencyCode, err.Error())
 		return ""
 	}
+	if n.TestnetEnable == true {
```
cpacia: govet `if n.TestnetEnable`

placer14: Yikes. Thx.

PR #1699, opened 2019-09-06
Update wallet request/responses to be simpler

placer14:
- `POST /wallet/spend` request now allows CurrencyCode or a full Currency definition 
- `GET /wallet/estimatefee` response is flatter (removed top-level key)
- `POST /wallet/spend` response is less verbose (removed definition from `amount`, `confirmed`, `unconfirmed` keys and included only once)
- `POST /wallet/spend` accepts `currencyCode` instead of definition
- Websockets `OrderNotification`, `IncomingTransaction`, and `walletUpdate` now use CurrencyValue
- Bug fix when relying on `wallet-interface.ExchangeRate().CoinsPerUnit()` divisibility being incorrectly calculated
- `GET /wallet/balance` is flatter (removed CurrencyDefinition from each balance and included only once)


tyler-smith:
Lgtm

PR #1702, opened 2019-09-11
Improve CurrencyValue handling of divisibility

placer14:
- Fix how we apply the CurrencyValue to ModeratorFee object in pb.Profile

...which prompted the below improvements:
- Add helper functions to convert to/from pb.CurrencyValue
- CurrencyValue.AdjustDivisibility() updates divisibility and amount
- Allow CurrencyValue.Equal() to consider different divisibilities
- CurrencyValue.Normalize() will AdjustDivisibility() to match local def


core/profile.go
```diff
@@ -102,6 +116,16 @@ func (n *OpenBazaarNode) UpdateProfile(profile *pb.Profile) error {
 		return err
 	}
 	profile.LastModified = ts
+	if err := ValidateProfile(profile); err != nil {
+		return err
+	}
+
+	m := jsonpb.Marshaler{
+		EnumsAsInts:  false,
+		EmitDefaults: true,
```
cpacia: Do we want default fields shown in the profile?

cpacia: Looks like we currently do.

PR #1675, opened 2019-08-08
[WIP] Repo.Listing changes

amangale:
This PR builds upon https://github.com/OpenBazaar/openbazaar-go/pull/1666. 
This introduces repo.Listing which will be used to handle different versions of existing/legacy contracts in the nodes. Repo.Listing will used for interaction between pb.Listing and the core.

PR #1715, opened 2019-09-23
Support legacy (v4) listing index and other refactorings

placer14:
This PR comes AFTER #1675 is merged (as it builds on top of it).

Contains following changes:
- support legacy (v4) listing index when consumed from other peers
- separate Testnet and Mainnet currencies in CurrencyDictionary
- bug fix which restores lost significant figures when doing math w custom divisibility
- make CurrencyDictionary more mutation-safe


repo/currency_definition.go
```diff
@@ -336,8 +374,8 @@ func (c *CurrencyDefinition) Valid() error {
 
 // Equal indicates if the receiver and other have the same code
 // and divisibility
-func (c *CurrencyDefinition) Equal(other *CurrencyDefinition) bool {
-	if c == nil || other == nil {
+func (c CurrencyDefinition) Equal(other CurrencyDefinition) bool {
+	if c == NilCurrencyDefinition || other == NilCurrencyDefinition {
```
cpacia: Not that this is an expected usage, but should the case where both c and other are NilCurrencyDefinition return true?

repo/currency_definition.go
```diff
@@ -353,24 +391,24 @@ func (c *CurrencyDefinition) Equal(other *CurrencyDefinition) bool {
...
 
-func (c CurrencyDictionary) All() map[string]*CurrencyDefinition {
-	if c.mainnetDefinitions == nil {
-		return nil
-	}
-	return c.mainnetDefinitions
+// AsMap returns a mutable map of all known the definitions. Note: This map is not threadsafe and
+// care should be taken to not mutate this.
+func (c CurrencyDictionary) AsMap() map[string]CurrencyDefinition {
+	return c.definitions
```
cpacia: Should this return a copy of the map so as to avoid accidental mutations?

Issue #1701, opened 2019-09-10
ETH - wallet message is missing


jjeffryes:
The client listens for a socket message with a .wallet object containing a tx and coin type to show new transactions.

As of 4ba53f7d in the ethereum-master branch, that message is no longer being sent when new outgoing or incoming transactions happen.

This was working previously, so a recent change might be responsible.

rmisio:
@placer14 Just to clarify, this is a `wallet` socket which is distinct from the `walletUpdate` socket. I'm fairly sure it was in a few days ago and wondering if somehow it got borked when you updated the `walletUpdate` socket?

rmisio:
Just to be clear, the `wallet` socket comes when a transaction is first seen and when it initially confirms (goes from 0 to 1 confirmations).

rmisio:
Just tried this on `ethereum-master` and it's still an issue - no `wallet` socket comes.

drwasho:
Updated this in https://github.com/OpenBazaar/openbazaar-go/issues/1420 but we don't have a websocket to report new ETH/ERC-20 transactions. The only socket message we'll get is the `orderspend` network message for now.

rmisio:
> Updated this in #1420 but we don't have a websocket to report new ETH/ERC-20 transactions. The only socket message we'll get is the orderspend network message for now.

@drwasho Just to be clear, this issue is reported that the `wallet` socket is missing for every coin, not just ETH.

I was in up to a few weeks ago and somehow along the way got lost.

Issue #1696, opened 2019-09-04
Please make the `wallet-update` socket schema match the schema returned via `wallet/balance`


rmisio:
`wallet/balance` returns

```
{
    "TBCH": {
        "confirmed": {
            "amount": "0",
            "currency": {
                "code": "BCH",
                "currencyType": "crypto",
                "divisibility": 8,
                "name": "Bitcoin Cash",
                "testnetCode": "TBCH"
            }
        },
        "height": 1325948,
        "unconfirmed": {
            "amount": "0",
            "currency": {
                "code": "BCH",
                "currencyType": "crypto",
                "divisibility": 8,
                "name": "Bitcoin Cash",
                "testnetCode": "TBCH"
            }
        }
    },
    ...
}
```

Currently the `walletUpdate` socket returns:

![image](https://user-images.githubusercontent.com/794517/64284846-b3d83800-cf17-11e9-93dd-f934f8eb78f0.png)

The expectation is that it returns the same data so clients can use common code to parse the result.

rmisio:
Honestly, the schema of the api call could be updated to match the socket, since clients could get the divisibility from the wallet cur def. It would make for a leaner payload, but force client to have to call a second API (which I imagine most would do only once and cache).
 
Up to you which way you want to go, but the critical part is consistency between the two returns.

PR #1728, opened 2019-09-28
Final change the protobufs (supporting v4 and v5 schemas)

placer14:
Many changes are in this PR. They stem from the protobufs being updated and propagate out to the rest of the app and unit tests (which build and pass). QA tests are currently untouched and still need attention. This also doesn't support any of the untested examples (legacy support scenarios, primarily).

This is a good milestone, though.


drwasho:
My feedback:

- [ ] 1. We have a bug where the API fails when trying to spend ETH/TETH, but the transactions still go through https://github.com/OpenBazaar/openbazaar-go/issues/1730. Does it make sense to solve it in this PR, or do we want to merge this and fix the bug as a separate PR?
 
- [ ] 2. Updating the websocket message to be consistent with the REST API is something that I think this PR should address: https://github.com/OpenBazaar/openbazaar-go/issues/1696

- [ ] 3. Related to (2) is the bug reported here: https://github.com/OpenBazaar/openbazaar-go/issues/1701

- [ ] 4. In the interests of time we should merge this PR and follow-up with updated QA tests as soon as possible. To expedite that, @placer14 it would be really helpful if you could list the QA tests that are failing and require refactoring.

pb/protos/contracts.proto
```diff
@@ -222,13 +215,15 @@ message Order {
...
-    string paymentAddress                     = 3;
-    //uint64 requestedAmount             = 4 [json_name = ""];
-    repeated RatingSignature ratingSignatures = 5;
-    CurrencyValue requestedAmountValue             = 6 [json_name = "requestedAmount"];
+    string paymentAddress                      = 3;
+    uint64 requestedAmount                     = 4 [deprecated = true];
+
+    repeated RatingSignature ratingSignatures  = 5;
+    string bigRequestedAmount                  = 6;
+    CurrencyDefinition requestedAmountCurrency = 7;
```
cpacia: Does look like we're using this anywhere. Not sure it needs to be included.

core/listings.go
```diff
@@ -310,13 +314,14 @@ func (n *OpenBazaarNode) extractListingData(listing *pb.SignedListing) (repo.Lis
 		ContractType:       listing.Listing.Metadata.ContractType.String(),
 		Description:        listing.Listing.Item.Description[:descriptionLength],
 		Thumbnail:          repo.ListingThumbnail{listing.Listing.Item.Images[0].Tiny, listing.Listing.Item.Images[0].Small, listing.Listing.Item.Images[0].Medium},
-		Price:              repo.CurrencyValue{Currency: defn, Amount: amt},
+		Price:              priceValue,
 		Modifier:           listing.Listing.Metadata.PriceModifier,
```
cpacia: using deprecated metadata.PriceModifier rather than item.PriceModifier

cpacia:
It appears `func (l *Listing) GetPriceModifier() (float32, error) ` is returning from the deprecated metadata.PriceModifier rather than new item.PriceModifier

cpacia:
I have opened an new issue for the final changes that we are going to need. https://github.com/OpenBazaar/openbazaar-go/issues/1738

So far this PR LGTM other than the few comments I made.

drwasho:
- [ ] There is inconsistency when fetching a listing:

`GET /ob/listing/:slug`

image

`GET /ob/listing/ipfs/:hash`

image

drwasho:
- [ ] Estimating a purchase (`POST /ob/estimatetotal`) && making a purchase order (`POST /ob/purchase`) is failing with the following payload and response:

Payload:

```JSON
{
    "shipTo": "Elwood Blues",
    "address": "1060 W Addison",
    "city": "Chicago",
    "state": "Illinois",
    "countryCode": "UNITED_STATES",
    "postalCode": "60613",
    "addressNotes": "",
    "items": [
        {
            "listingHash": "QmTeGATiiTEbNwkLosSRivSApstXLof45jWRjSAVoB1QUv",
            "bigQuantity": "1",
            "options": [
                {
                    "name": "Sizes",
                    "value": "Small"
                },
                {
                    "name": "Color",
                    "value": "Red"
                }
            ],
            "shipping": {
                "name": "Worldwide",
                "service": "Standard"
            },
            "memo": "thanks!",
            "coupons": []
        }
    ],
    "moderator": "QmcdkKM2fWCKzTZdpjEY6abzRbDhRRJvNmqwNKVyiDtGky",
    "paymentCoin": "TLTC"
}
```

Reponse:

```JSON
{
    "success": false,
    "reason": "unknown currency"
}
```

PR #1729, opened 2019-09-29
PeerID is not included in SpendResponse

hoffmabc:
For some reason, the spendresponse struct never included the vendor’s peer id and caused the payment message to never be sent.

### DIFF
--- a/core/spend.go
+++ b/core/spend.go
@@ -124,6 +124,7 @@ func (n *OpenBazaarNode) Spend(args *SpendRequest) (*SpendResponse, error) {
 		Timestamp:          txn.Timestamp,
 		Memo:               memo,
 		OrderID:            args.OrderID,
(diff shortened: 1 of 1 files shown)


placer14:
Good catch here. It appears it was already fixed as part of changes delivered to `ethereum-master`. 

Impact analysis: This looks like it definitely would have prevented the vendor from ever getting a ORDER_PAYMENT message from the buyer. This is a new message type which was created primarily for ETH which has no way of detecting payments for a specific order. This seems mostly benign unless there were changes in how the order state progresses when this new message type was added. It's probably okay for this PeerID to be missing for any non-eth order as the failure doesn't impact the regular checkout flow. I'm saying this, of course, assuming that nothing else has been changed to become dependent on the PeerID being present or the OrderMessage getting delivered successfully as part of the conditions for completing a checkout.

Closing this PR as already resolved.

Issue #1431, opened 2019-01-30
Unable to manually sweep ETH wallet all funds


placer14:
Exploring the wallet's limits and I can't seem to drain the remaining 0.000032 ETH (32e12 wei) from the wallet. A call to `/ob/spend` trying to send 1 wei is being blocked with responses of `"insufficient funds for gas * price + value"`.

Reproducing steps:
I filled my wallet with arbitrary amount... then attempted to manually sweep the wallet via the Spend call:

- I started with the full amount of wei in my wallet,
- then estimated the fees
- subtracted that estimated fee amount from the full amount
- added a little more than available and then walked back downward until the actual spend completed.

I know this approach is imperfect, but I would've expected a LOT fewer wei to be left over.

drwasho:
Couple of questions:

1. Is it always 0.000032 ETH?
2. What was the gas price (in Gwei) and total gas that you estimated for the fee?

placer14:
1. In 3 trials, it was always 0.000032 remaining when the spend finally succeeds.
2. I used Normal Priority (not sure what the gas setting is there). The fee was consistently estimated as `53000000000000 wei` regardless of the amount sent. Could this be another bug? Intended behavior?

drwasho:
It sounds like a gas calculation bug, @amangale can take a look.

drwasho:
At the time this issue was created, we didn't have the `spendAll` flag for `POST /wallet/spend`.

- [ ] As as result, calling this API with `spendAll: true` needs to be tested by @amangale

Issue #1418, opened 2019-01-22
releasefunds and orderconfirmation have insufficient gas for ETH orders


jjeffryes:
Attempting to accept an offline order with `orderconfirmation` or accept a dispute resolution with `releasefunds` returns an error of `insufficient funds for gas * price + value` when using ETH.

This appears to be true for orders between $0.01 and $1.00 of ETH. Larger amounts couldn't be tested due to the number as string bug (#1417).

cpacia:
Yeah people will need to have a positive eth balance to accept eth orders due to how eth works.

cpacia:
Maybe we need a custom error code for insuffient gas

PR #1733, opened 2019-09-30
Fix 1418 and 1431

amangale:
This PR addresses the issues #1418 and #1431.

There is a balance check in place before a call to multisign is made.

Also, the spendAll flag is now checked before transfer.


cpacia:
LGTM, just make sure these changes also get into the eth wallet repo.

PR #1755, opened 2019-10-02
Fix eth dispute resolution

amangale:
This PR makes the changes required for progressing DECIDED disputes to RESOLVED.

### DIFF
--- a/net/service/handlers.go
+++ b/net/service/handlers.go
@@ -1817,11 +1817,23 @@ func (service *OpenBazaarService) handleOrderPayment(peer peer.ID, pmes *pb.Mess
 		}
 	}
 
-	toAddress, _ := wal.DecodeAddress(txn.ToAddress)
+	toAddress, err := wal.DecodeAddress(contract.BuyerOrder.Payment.RedeemScript)
+	if err != nil {
+		log.Error(err)
+	}
+
+	tvalue, ok := new(big.Int).SetString(txn.Value, 10)
+	if ok && tvalue.Int64() == 0 {
+		toAddress, err = wal.DecodeAddress(contract.BuyerOrder.Payment.RedeemScript)
+		if err != nil {
+			log.Error(err)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


net/service/handlers.go
```diff
@@ -1817,11 +1817,23 @@ func (service *OpenBazaarService) handleOrderPayment(peer peer.ID, pmes *pb.Mess
...
 	}
 
-	toAddress, _ := wal.DecodeAddress(txn.ToAddress)
+	toAddress, err := wal.DecodeAddress(contract.BuyerOrder.Payment.RedeemScript)
+	if err != nil {
+		log.Error(err)
+	}
+
+	tvalue, ok := new(big.Int).SetString(txn.Value, 10)
+	if ok && tvalue.Int64() == 0 {
```
placer14: `tvalue.Int64` is undefined if not an Int64 which is possible for ETH tx.

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -388,7 +389,25 @@ func (wallet *EthereumWallet) NewAddress(purpose wi.KeyPurpose) btcutil.Address
 
 // DecodeAddress - Parse the address string and return an address interface
 func (wallet *EthereumWallet) DecodeAddress(addr string) (btcutil.Address, error) {
-	ethAddr := common.HexToAddress(addr)
+	var ethAddr common.Address
+	if len(addr) > 64 {
```
placer14: It would be nice to document the reason for this check somehow. Either with a variable or a comment.

placer14: Would also be nice to see a test case protecting both versions of Decoding inputs.

placer14: I extracted the ethScript conversion to document this a bit. Still missing a unit test though.

PR #1774, opened 2019-10-10
Add blocktime to currency definition

### LINKED ISSUE #1722: Blocktime should be stored in repo.CurrencyDefinition
In the current implementation, the block time is assumed to be 10 minutes across the supported cryptocurrencies. By storing this inside the repo.CurrencyDefinition struct as a new time.Duration field, the escrow timeout specific to a crypto can be handled properly.

An additional check to `repo.CurrencyDefinition.Valid()` should also be included that the blocktime Duration is greater than zero.

amangale:
This PR resolves #1722.

### DIFF
--- a/api/jsonapi_data_test.go
+++ b/api/jsonapi_data_test.go
@@ -333,7 +333,8 @@ const walletBalanceJSONResponse = `{
     "TBCH": {
         "confirmed": "0",
 				"currency": {
-						"code": "TBCH",
+                        "BlockTime": 600000000000,
+                        "code": "TBCH",
 						"currencyType": "crypto",
 						"divisibility": 8,
 						"name": "Testnet Bitcoin Cash"
@@ -344,7 +345,8 @@ const walletBalanceJSONResponse = `{
     "TBTC": {
         "confirmed": "0",
 				"currency": {
-						"code": "TBTC",
+                        "BlockTime": 600000000000,
+                        "code": "TBTC",
 						"currencyType": "crypto",
 						"divisibility": 8,
 						"name": "Testnet Bitcoin"
@@ -355,7 +357,8 @@ const walletBalanceJSONResponse = `{
     "TLTC": {
         "confirmed": "0",
 				"currency": {
-						"code": "TLTC",
+                        "BlockTime": 600000000000,
+                        "code": "TLTC",
 						"currencyType": "crypto",
 						"divisibility": 8,
 						"name": "Testnet Litecoin"
@@ -366,7 +369,8 @@ const walletBalanceJSONResponse = `{
     "TZEC": {
         "confirmed": "0",
 				"currency": {
-						"code": "TZEC",
+                        "BlockTime": 600000000000,
+                        "code": "TZEC",
 						"currencyType": "crypto",
 						"divisibility": 8,
 						"name": "Testnet Zcash"
--- a/api/jsonapi_data_test.go
+++ b/api/jsonapi_data_test.go
@@ -333,7 +333,6 @@ const walletBalanceJSONResponse = `{
     "TBCH": {
         "confirmed": "0",
 				"currency": {
-                        "BlockTime": 600000000000,
                         "code": "TBCH",
 						"currencyType": "crypto",
 						"divisibility": 8,
@@ -345,7 +344,6 @@ const walletBalanceJSONResponse = `{
     "TBTC": {
         "confirmed": "0",
 				"currency": {
-                        "BlockTime": 600000000000,
                         "code": "TBTC",
 						"currencyType": "crypto",
 						"divisibility": 8,
@@ -357,7 +355,6 @@ const walletBalanceJSONResponse = `{
     "TLTC": {
         "confirmed": "0",
 				"currency": {
-                        "BlockTime": 600000000000,
                         "code": "TLTC",
 						"currencyType": "crypto",
 						"divisibility": 8,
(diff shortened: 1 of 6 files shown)


api/jsonapi_data_test.go
```diff
@@ -366,7 +369,8 @@ const walletBalanceJSONResponse = `{
     "TZEC": {
         "confirmed": "0",
 				"currency": {
-						"code": "TZEC",
+                        "BlockTime": 600000000000,
+                        "code": "TZEC",
```
placer14: I don't think we should put blocktime into the persisted schema. I think leaving it as a runtime datapoint is sufficient. Let's ensure this data doesn't leak into our various outputs.

repo/currency_definition.go
```diff
@@ -53,7 +57,7 @@ var (
 	ErrDictionaryIndexMismatchedCode   = errors.New("dictionary index mismatched with definition currency code")
 	ErrDictionaryCurrencyCodeCollision = errors.New("currency code is used by more than one currency")
 
-	NilCurrencyDefinition = CurrencyDefinition{Name: "", Code: NilCurrencyCode, Divisibility: 0, CurrencyType: ""}
+	NilCurrencyDefinition = CurrencyDefinition{Name: "", Code: NilCurrencyCode, Divisibility: 0, CurrencyType: "", BlockTime: DefaultBlockTime}
```
placer14: NilBlockTime should not be a useable default. I would use the zero value of `time.Time`.

repo/currency_definition.go
```diff
@@ -415,3 +419,16 @@ func (c CurrencyDictionary) AsMap() map[string]CurrencyDefinition {
 	}
 	return defCopy
 }
+
+// ConfirmationsPerHour will calculate the no of confirmations in 1 hr
+// this is valid only for a crypto
+func ConfirmationsPerHour(currency string) uint64 {
+	defn, err := AllCurrencies().Lookup(currency)
+	if err != nil || defn.CurrencyType != "crypto" {
+		return 0
```
placer14: I'm not sure I like `confirmationsForTimeout := contract.VendorListings[0].Metadata.EscrowTimeoutHours * 			uint32(repo.ConfirmationsPerHour(order.Payment.AmountCurrency.Code))` being `0` when the `order.Payment.AmountCurrency.Code` is empty or invalid. This seems like a bug that will hurt us unexpectedly later.

Instead, why don't we do the lookup inside `verifyEscrowFundsAreDisputeable` where we can properly handle the error case of having an invalid currency code. Then, we can make a method that looks like `(CurrencyDefinition) ConfirmationsPerHour() uint64` which will always have a valid response. `NilCurrencyDefinition` will still return `0` but you know the dictionary lookup result will be a valid currency if there's no error.

repo/currency_definition.go
```diff
@@ -62,18 +66,18 @@ var (
...
 	testnetCryptoDefinitions = map[string]CurrencyDefinition{
-		"TBTC": {Name: "Testnet Bitcoin", Code: CurrencyCode("TBTC"), CurrencyType: Crypto, Divisibility: 8},
-		"TBCH": {Name: "Testnet Bitcoin Cash", Code: CurrencyCode("TBCH"), CurrencyType: Crypto, Divisibility: 8},
-		"TLTC": {Name: "Testnet Litecoin", Code: CurrencyCode("TLTC"), CurrencyType: Crypto, Divisibility: 8},
-		"TZEC": {Name: "Testnet Zcash", Code: CurrencyCode("TZEC"), CurrencyType: Crypto, Divisibility: 8},
-		"TETH": {Name: "Testnet Ethereum", Code: CurrencyCode("TETH"), CurrencyType: Crypto, Divisibility: 18},
+		"TBTC": {Name: "Testnet Bitcoin", Code: CurrencyCode("TBTC"), CurrencyType: Crypto, Divisibility: 8, BlockTime: DefaultBlockTime},
+		"TBCH": {Name: "Testnet Bitcoin Cash", Code: CurrencyCode("TBCH"), CurrencyType: Crypto, Divisibility: 8, BlockTime: DefaultBlockTime},
+		"TLTC": {Name: "Testnet Litecoin", Code: CurrencyCode("TLTC"), CurrencyType: Crypto, Divisibility: 8, BlockTime: DefaultBlockTime},
+		"TZEC": {Name: "Testnet Zcash", Code: CurrencyCode("TZEC"), CurrencyType: Crypto, Divisibility: 8, BlockTime: DefaultBlockTime},
```
placer14: I don't like the idea of having a `DefaultBlockTime`. It suggests that one block time is preferred over another. Instead, we should just put `10 * time.Minute` for BTC/BCH/ZEC. And please correct LTC and TLTC blocktime to be 2.5 minutes.

core/disputes.go
```diff
@@ -176,6 +172,8 @@ func (n *OpenBazaarNode) verifyEscrowFundsAreDisputeable(contract *pb.RicardianC
 			log.Errorf("Failed GetConfirmations(%s): %s", hash.String(), err.Error())
 			return false
 		}
+		confirmationsForTimeout := contract.VendorListings[0].Metadata.EscrowTimeoutHours *
+			uint32(repo.ConfirmationsPerHour(order.Payment.AmountCurrency.Code))
```
placer14: And since you're casting to `uint32` here and not using `ConfirmationsPerHour` anywhere else, why don't you just change the return type of the function?

Issue #1785, opened 2019-10-15
Create compatibility QA tests for purchasing between v4 and v5 nodes


drwasho:
With the Ethereum update, we've taken a lot of measure to ensure that v4 (legacy) and v5 (upgraded 'eth-compatible') nodes can transact. We need to ensure that these interactions will proceed without error; note that this only affects UTXO purchases as v5 nodes cannot purchase listings with ETH from v4 nodes.

The best way to do this is to create some dedicated compatibility QA tests where: 

**The buyer is a v4 node and the seller is a v5 node**

- [ ] 1. cancel direct offline | blah |
- [ ] 2. complete direct online
- [ ] 3. complete disputed
- [ ] 4. complete moderated online
- [ ] 5. complete moderated with timeout
- [ ] 6. dispute close buyer
- [ ] 7. dispute close split
- [ ] 8. dispute close vendor
- [ ] 9. escrow release after timeout
- [ ] 10. fulfill direct online
- [ ] 11. market price modifier
- [ ] 12. out of inventory
- [ ] 13. purchase crypto listing
- [ ] 14. purchase digital
- [ ] 15. purchase direct offline
- [ ] 16. purchase direct online
- [ ] 17. purchase moderated offline
- [ ] 18. purchase moderated online
- [ ] 19. purchase offline error
- [ ] 20. refund direct
- [ ] 21. refund moderated
- [ ] 22. reject direct offline
- [ ] 23. reject moderated offline

**The buyer is a v5 node and the seller is a v4 node**

- [ ] 24. cancel direct offline | blah |
- [ ] 25. complete direct online
- [ ] 26 complete disputed
- [ ] 27. complete moderated online
- [ ] 28. complete moderated with timeout
- [ ] 29. dispute close buyer
- [ ] 30. dispute close split
- [ ] 31. dispute close vendor
- [ ] 32. escrow release after timeout
- [ ] 33. fulfill direct online
- [ ] 34. market price modifier
- [ ] 35. out of inventory
- [ ] 36. purchase crypto listing
- [ ] 37. purchase digital
- [ ] 38. purchase direct offline
- [ ] 39. purchase direct online
- [ ] 40. purchase moderated offline
- [ ] 41. purchase moderated online
- [ ] 42. purchase offline error
- [ ] 43. refund direct
- [ ] 44. refund moderated
- [ ] 45. reject direct offline
- [ ] 46. reject moderated offline

PR #1792, opened 2019-10-16
(#1785) Update test_framework.py to support version compatibility tests

cpacia:
The current way of running tests is:
```python
python3 test_name.py -b /path/to/obgo-binary -d /path/to/bitcoind
```
This way still works, but a new option is available:
```python
python3 test_name.py -i /path/to/obgo-binary-v1 /path/to/obgo-binary-v2 -d /path/to/bitcoind
```
Where `i` is a list of string paths for each ob-go binary that is used. The binaries can be different versions. The number of binaries in the list must be equal to the number of nodes created by the test and the index of the binary string must map to the node (buyer, vendor, moderator) that you want it to map to. 

This PR is technically incomplete as it only allows the use of multiple binaries, however all the test data is currently formatted using the current version and not prior versions. Getting the old version shouldn't be difficult as it's just copying and pasting the test data from the old version. However, each test will need to be updated to decide which version of the data file it needs to load and post to the server. 

I've added another flag `-v` that can be used to specify the versions which map to each node. For example:
```python
`-v v1 v2 v2`
```
Means that node 0 is a v1 node and nodes 2 and 3 are v2 nodes. Each test would need to do something like:

```python
alice = self.nodes[1]

file_path = "testdata/listing.json"
if self.versions is not None and self.versions[1] is "v1":
    file_path = "testdata/listing_old.json"
with open(file_path) as listing_file:
    listing_json = json.load(listing_file, object_pairs_hook=OrderedDict)
```


hoffmabc:
@cpacia are there any tests being run against different versions in this PR so far?

hoffmabc:
The Travis issue is resolved in my transport PR so we don't need to deal with the linting issue here.

cpacia:
v4 tests:

- [ ] cancel_direct_offline
- [ ] chat
- [ ] chat_offline
- [ ] complete_direct_online
- [ ] complete_disputed
- [ ] complete_moderated_online
- [ ] complete_moderated_with_timeout
- [ ] dispute_close_buyer
- [ ] dispute_close_vendor
- [ ] dispute_close_split
- [ ] escrow_release_after_timeout
- [ ] fulfill_direct_online
- [ ] listings
- [ ] manage_crypto_listings
- [ ] market_price_modifier.py
- [ ] out_of_inventory
- [ ] patch_profile
- [ ] purchase_crypto_listing
- [ ] purchase_digital
- [ ] purchase_direct_online
- [ ] purchase_direct_offline
- [ ] purchase_moderated_offline
- [ ] purchase_moderated_online
- [ ] purchase_offline_error
- [ ] receive_coins
- [ ] refund_direct
- [ ] refund_moderated
- [ ] reject_direct_offline
- [ ] reject_moderated_offline
- [ ] smtp_notifications
- [ ] upload_listing

Issue #1797, opened 2019-10-17
Eliminating the fragility when receiving order messages


drwasho:
There was a test scenario with Ethereum that revealed a fundamental problem in handling ETH orders and our messaging system in general:

1. Buyer sends a purchase order
2. Buyer funds the escrow address
3. Seller fulfils the order
4. Buyer sends the `ORDER_COMPLETION` message
5. Seller is stuck in the `FULFILLED` state since they have no Ethereum (gas) to call the `execute` function, which releases funds from the escrow address.
6. To make matters worse, even if the seller tops-up their wallet with Ethereum, there is no way for them to process the `ORDER_COMPLETION` message again.

The reason for (6) is that any online or offline messages received, while initially stored in the database at first, is removed to memory for processing. Irrespective of whether the message is successfully processed or not, the message is cleared from memory… never to be recovered.

In the scenario above, the `ORDER_COMPLETION` message failed to be processed successfully because the seller didn’t have the gas to release the funds, which is required to transition the order state from `FULFILLED` to `COMPLETED`. The `ORDER_COMPLETION` message gets dropped by the seller’s node, and there is no logic to save the message and retry at a later stage.

If the buyer triggers the `POST /ob/resendmessage` API call for the `ORDER_COMPLETION` message, and the seller has the funds to call the `execute` function in the escrow smart contract, then the funds will be released and the order state will correctly transition to `COMPLETED`. The problem with this process is that the seller is basically stuck unless the buyer happens to resend the *correct* order message to the seller again, and they have no way of knowing that they need to.

### Improving incoming message durability

What we need to do is replicate durability of the online/offline message inbox that we have built for the outbox. Specifically:

1. Every single online and offline message received by a node should be permanently stored in the database
2. We should also be tracking whether or not the message was successfully processed
3. We should have some logic to retry processing the message when:
    - The node starts up
    - The user manually triggers an API to process any failed messages for a specific orderId
    - For ETH orders only:
      - When funds are detected in the `eth-wallet`, the server will automatically check if there are any unprocessed `ORDER_COMPLETION`, `DISPUTE_CLOSE`, `ORDER_REJECT`, or `REFUND` messages
      - If so, the server double-checks INFURA to make sure funds haven’t moved
      - If so, the server triggers the execute function and updates the order state (critical to prevent replays here
[...]

PR #1817, opened 2019-10-23
(#1797) Inbound msg persistence

amangale:
Store inbound messages too as per issue #1797.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -20,7 +20,6 @@ import (
 	"github.com/OpenBazaar/openbazaar-go/net"
 	rep "github.com/OpenBazaar/openbazaar-go/net/repointer"
 	ret "github.com/OpenBazaar/openbazaar-go/net/retriever"
-	"github.com/OpenBazaar/openbazaar-go/pb"
 	"github.com/OpenBazaar/openbazaar-go/repo"
 	sto "github.com/OpenBazaar/openbazaar-go/storage"
 	"github.com/btcsuite/btcutil/hdkeychain"
@@ -122,6 +121,10 @@ type OpenBazaarNode struct {
 	seedLock    sync.Mutex
 
 	InitalPublishComplete bool
+
+	// InboundMsgScanner is a worker that scans the messages
+	// table and tries to retry a failed order message
+	InboundMsgScanner *inboundMessageScanner
 }
 
 // TestNetworkEnabled indicates whether the node is operating with test parameters
@@ -310,36 +313,4 @@ func (n *OpenBazaarNode) EncryptMessage(peerID peer.ID, peerKey *libp2p.PubKey, 
 // IPFSIdentityString - IPFS identifier
 func (n *OpenBazaarNode) IPFSIdentityString() string {
 	return n.IpfsNode.Identity.Pretty()
-}
-
-// CheckErroredOrderMessages check if there are any order messages which can be retried
-func (n *OpenBazaarNode) CheckErroredOrderMessages(interval time.Duration) {
-	if interval == 0 {
-		return
-	}
-	ticker := time.NewTicker(interval)
-	go func() {
-		for range ticker.C {
-			msgs, err := n.Datastore.Messages().GetAllErrored()
-			if err != nil {
-				log.Error(err)
-			} else {
-				for _, m := range msgs {
-					if m.MsgErr.Error() == ErrInsufficientFunds.Error() {
-
-						// Get handler for this msg type
-						handler := n.Service.HandlerForMsgType(pb.Message_MessageType(m.MessageType))
-						if handler != nil {
-							// Dispatch handler
-							//rpmes, err := handler(m.PeerID, m.Message, nil)
-							//if err != nil {
-							//	log.Debugf("%s handle message error from %s: %s", pmes.MessageType.String(), mPeer.Pretty(), err)
-							//}
-						}
-
-					}
-				}
-			}
-		}
-	}()
 }
--- a/core/core.go
+++ b/core/core.go
@@ -20,6 +20,7 @@ import (
 	"github.com/OpenBazaar/openbazaar-go/net"
 	rep "github.com/OpenBazaar/openbazaar-go/net/repointer"
 	ret "github.com/OpenBazaar/openbazaar-go/net/retriever"
+	"github.com/OpenBazaar/openbazaar-go/pb"
 	"github.com/OpenBazaar/openbazaar-go/repo"
 	sto "github.com/OpenBazaar/openbazaar-go/storage"
 	"github.com/btcsuite/btcutil/hdkeychain"
@@ -309,4 +310,36 @@ func (n *OpenBazaarNode) EncryptMessage(peerID peer.ID, peerKey *libp2p.PubKey, 
 // IPFSIdentityString - IPFS identifier
 func (n *OpenBazaarNode) IPFSIdentityString() string {
 	return n.IpfsNode.Identity.Pretty()
+}
+
+// CheckErroredOrderMessages check if there are any order messages which can be retried
+func (n *OpenBazaarNode) CheckErroredOrderMessages(interval time.Duration) {
+	if interval == 0 {
+		return
+	}
+	ticker := time.NewTicker(interval)
+	go func() {
+		for range ticker.C {
+			msgs, err := n.Datastore.Messages().GetAllErrored()
+			if err != nil {
+				log.Error(err)
+			} else {
+				for _, m := range msgs {
+					if m.MsgErr.Error() == ErrInsufficientFunds.Error() {
+
+						// Get handler for this msg type
+						handler := n.Service.HandlerForMsgType(pb.Message_MessageType(m.MessageType))
+						if handler != nil {
+							// Dispatch handler
+							//rpmes, err := handler(m.PeerID, m.Message, nil)
+							//if err != nil {
+							//	log.Debugf("%s handle message error from %s: %s", pmes.MessageType.String(), mPeer.Pretty(), err)
+							//}
+						}
+
+					}
+				}
+			}
+		}
+	}()
 }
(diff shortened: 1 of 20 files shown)


vendor/github.com/OpenBazaar/wallet-interface/wallet.go
```diff
@@ -51,6 +51,11 @@ type Wallet interface {
 	walletCanBumpFee
 }
 
+var (
+	// ErrInsufficientFunds is returned when the wallet is unable to send the amount specified due to the balance being too low
+	ErrInsufficientFunds = errors.New("ERROR_INSUFFICIENT_FUNDS")
```
placer14: We should be very thoughtful before prescribing an error sentinel as part of the interface. This prescription requires all wallets which implement the interface to also be aware of and properly return this specific error. I don't see what's producing this error which makes it hard to justify where it should go, but seeing that it's consumed within `core`, that seems like the best place for it's declaration.

core/core.go
```diff
@@ -310,3 +311,35 @@ func (n *OpenBazaarNode) EncryptMessage(peerID peer.ID, peerKey *libp2p.PubKey,
...
+							//if err != nil {
+							//	log.Debugf("%s handle message error from %s: %s", pmes.MessageType.String(), mPeer.Pretty(), err)
+							//}
+						}
+
+					}
+				}
+			}
+		}
+	}()
```
placer14: Please build this as a separate service which runs parallel to the node and manages it's own dependencies with a proper Start and Stop. See `RecordAgingNotifier` for an example of this pattern. Ensure that Stop handles the graceful shutdown of this goroutine and calls `ticker.Stop()` release that resource.

repo/db/messages.go
```diff
@@ -21,15 +22,15 @@ func NewMessageStore(db *sql.DB, lock *sync.Mutex) repo.MessageStore {
 }
 
 // Put will insert a record into the messages
-func (o *MessagesDB) Put(messageID, orderID string, mType pb.Message_MessageType, peerID string, msg repo.Message) error {
+func (o *MessagesDB) Put(messageID, orderID string, mType pb.Message_MessageType, peerID string, msg repo.Message, rErr error, receivedAt int) error {
```
placer14: `receivedAt` should be int64 so we don't unintentionally allow 32-bit values on 32-bit arch machines.

vendor/github.com/OpenBazaar/wallet-interface/wallet.go
amangale: The rationale here is that during multisig, the wallet would try to submit a txn and in case of eth, there is a distinct possibility that the existing wallet balance may be insufficient to fund the gas required for the txn. In such a case, `ErrInsufficientFunds` will be returned by Multisign. If there is a cleaner solution for wallet implementation and core to share, please let me know.

core/net.go
```diff
@@ -376,18 +378,22 @@ func (n *OpenBazaarNode) SendCancel(peerID, orderID string) error {
...
 			return err
 		}
 		kp = &k
+		pub = order.VendorListings[0].VendorID.Pubkeys.Identity
 	}
 	err = n.Datastore.Messages().Put(
 		fmt.Sprintf("%s-%d", orderID, int(pb.Message_ORDER_CANCEL)),
-		orderID, pb.Message_ORDER_CANCEL, peerID, repo.Message{Msg: m})
+		orderID, pb.Message_ORDER_CANCEL, peerID, repo.Message{Msg: m},
+		"", 0, pub)
```
placer14: What's the significance of putting a `0` timestamp versus `time.Now()`?

repo/datastore.go
```diff
@@ -439,8 +439,11 @@ type MessageStore interface {
 	Queryable
 
 	// Save a new message
-	Put(messageID, orderID string, mType pb.Message_MessageType, peerID string, msg Message) error
+	Put(messageID, orderID string, mType pb.Message_MessageType, peerID string, msg Message, err string, receivedAt int64, pubkey []byte) error
```
placer14: At what point is `Put` getting overloaded for all sorts of purposes? Some cases we provide empty keys, some cases we provide 0 timestamp. Plus I don't see any validation happening inside `Put` to prevent us from setting the DB state improperly. We should always approach our abstractions with allowing the least amount of behavior necessary and protecting the inputs for the cases we do allow with extreme prejudice.

repo/db/messages.go
```diff
@@ -61,31 +64,97 @@ func (o *MessagesDB) Put(messageID, orderID string, mType pb.Message_MessageType
 }
 
 // GetByOrderIDType returns the message for the specified order and message type
-func (o *MessagesDB) GetByOrderIDType(orderID string, mType pb.Message_MessageType) (*repo.Message, string, error) {
+func (o *MessagesDB) GetByOrderIDType(orderID string, mType pb.Message_MessageType) (*repo.Message, string, string, error) {
```
placer14: Only the `repo.Message` response and peerID seem to be used. What does the error string get used for? Should we collapse the response into `repo.Message` and rename it to `repo.MessageRecord`?

repo/migrations/Migration032.go
```diff
@@ -0,0 +1,143 @@
...
+			return fmt.Errorf("rollback failed: (%s) due to (%s)", rErr.Error(), err.Error())
+		}
+		return err
+	}
+	if err = tx.Commit(); err != nil {
+		return err
+	}
+
+	// Bump schema version
+	err = ioutil.WriteFile(repoVersionFilePath, []byte(migrationCreateMessagesAM09UpVer), os.ModePerm)
```
placer14: `writeRepoVer()` here

repo/migrations/Migration032.go
```diff
@@ -0,0 +1,143 @@
...
+			return fmt.Errorf("rollback failed: (%s) due to (%s)", rErr.Error(), err.Error())
+		}
+		return err
+	}
+	if err = tx.Commit(); err != nil {
+		return err
+	}
+
+	// Bump schema version
+	err = ioutil.WriteFile(repoVersionFilePath, []byte(migrationCreateMessagesAM09DownVer), os.ModePerm)
```
placer14: `writeRepoVer()`

repo/migrations/Migration032_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	var (
+		databasePath = appSchema.DatabasePath()
+		schemaPath   = appSchema.DataPathJoin("repover")
+
+		insertSQL = "insert into messages(messageID, err, received_at) values(?,?,?)"
+	)
+
+	// create schema version file
+	if err = ioutil.WriteFile(schemaPath, []byte("31"), os.ModePerm); err != nil {
```
placer14: Migration test should start at `32`, not `31`.

vendor/github.com/OpenBazaar/wallet-interface/wallet.go
placer14: I don't see where `Multisign` is actually returning `ErrInsufficientFunds`. I also don't know why a signing function would be concerned with wallet balance. In my mind, a transaction should be valid going into `Multisign` and should be considering the `ErrInsufficientFunds` scenario well before that point. This seems like a leaky abstraction that should be left to the wallet consumer to handle. Maybe we should discuss this particulars of this before I reject it completely, but I'm not seeing any benefit from this approach.

drwasho: > I also don't know why a signing function would be concerned with wallet balance. 

This applies to ETH as the Multisign function will call the `execute` function in the smart contract, and that function call requires gas to call, and if you wallet doesn't have enough gas it will fail.

But is your point is that we're checking the balance too late and we should fail early for this particular scenario?

core/net.go
amangale: The field received_at is set to 0 is for outbound messages and time.Now() is for inbound messages.

repo/datastore.go
amangale: The empty fields are the ones which make no sense for inbound or outbound messages. For example, received_at makes no sense for outbound messages and pubkey does not make sense for inbound messages.

core/inbound_message_scanner_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	datastore := db.NewSQLiteDatastore(database, new(sync.Mutex), wi.Bitcoin)
+	worker := &inboundMessageScanner{
+		datastore: datastore,
+		logger:    logging.MustGetLogger("testInboundMsgScanner"),
+	}
+
+	//worker.PerformTask()
+	fmt.Println(worker)
```
cpacia: Just remove the debug prints before merging

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -1381,7 +1381,7 @@ func (wallet *EthereumWallet) Multisign(ins []wi.TransactionInput, outs []wi.Tra
 
 	if requiredBalance.Cmp(currentBalance) > 0 {
 		// the wallet does not have the required balance
-		return nil, errors.New("insufficient balance to release funds or confirm order")
+		return nil, wi.ErrInsufficientFunds
```
placer14: This is fine for now. We don't need to reconcile the open question form now.

vendor/github.com/OpenBazaar/wallet-interface/wallet.go
placer14: I'm not sure if there's an earlier opportunity to check/fail for this case. My argument is whether it belongs as part of a signing function which feels like it should have separate responsibilities from balance checking. 

I don't suppose I have enough information to make a good design decision here so I'll push this off for the time being.

placer14:
It doesn't look like the scanner is being started. Is that intended @amangale?

repo/datastore.go
placer14: We can resolve this afterward. Tracking: https://github.com/OpenBazaar/openbazaar-go/issues/1855

PR #1824, opened 2019-10-24
[WIP] Send offline messages through web relay if configured 

hoffmabc:
This does not address the config file migration so for now to test this you need to add this to the bottom of your config file:

```
"WebRelays": [
        "wss://webchat.ob1.io:8080"
    ]
```

Offline messages will now also be sent through the configured web relay using websockets with this PR.


drwasho:
Excellent! @hoffmabc will you be making addition commits for this PR to add the offline message listener via the relay?

drwasho:
@hoffmabc bump

placer14:
@hoffmabc if we're pursuing this, I'd like to see the following improvements after doing a quick scan through:

0) Update the PR description indicating what impact these changes will have. Maybe a use case illustrating their need. And perhaps instructions for setting this up for independent testing/verification. (Something that can later be used for updating our official docs.)
1) Error handling inside the relay manager.
2) Unit tests around the relay manager.

placer14:
Thinking about this further, I like that this was produced as a separate manager from a pattern perspective. However, I think that our message sending is getting spread across a few different parts of our app and this implementation might be improved if we take a second to refactor our message sending behavior to support sending a message via multiple transports (a message mux of sorts).

I don't think we're in a hurry to deliver this, perhaps we can reconsider our approach before digging into unit tests and error handling as mentioned before? /cc @drwasho @hoffmabc

hoffmabc:
I'm fine with whatever approach you guys want to take. I'll defer to @drwasho on this one.

hoffmabc:
> Thinking about this further, I like that this was produced as a separate manager from a pattern perspective. However, I think that our message sending is getting spread across a few different parts of our app and this implementation might be improved if we take a second to refactor our message sending behavior to support sending a message via multiple transports (a message mux of sorts).
> 
> I don't think we're in a hurry to deliver this, perhaps we can reconsider our approach before digging into unit tests and error handling as mentioned before? /cc @drwasho @hoffmabc

Perhaps this refactor to mux model could be a larger, broader refactor later on? I think that pursuing that in this PR would probably expand the scope and prevent rollout for quite a while.

core/net.go
```diff
@@ -575,6 +636,7 @@ func (n *OpenBazaarNode) SendChat(peerID string, chatMessage *pb.Chat) error {
 		return err
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), n.OfflineMessageFailoverTimeout)
+	n.SendRelayedMessage(p, nil, &m) // send relayed message immediately
```
rmisio: `SendRelayedMessage` is called in `SendMessage` which is called below. Why call it twice?

core/net.go
```diff
@@ -795,6 +857,9 @@ func (n *OpenBazaarNode) SendOrderPayment(peerID string, paymentMessage *pb.Orde
 	if err != nil {
 		return err
 	}
+
+	n.SendRelayedMessage(p, nil, &m) // send relayed message immediately
```
rmisio: Same question as above:

`SendRelayedMessage` is called in `SendMessage` which is called below. Why call it twice?`

net/web_relay_manager.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Establish connections
+	var conns []*websocket.Conn
+	for _, relay := range wrm.webrelays {
+
+		// Connect and subscribe to websocket server
+		conn, err := wrm.connectToServer(relay, wrm.peerID)
+		if err != nil {
+			log.Error("Could not connect to: %s", relay)
+		}
```
rmisio: @hoffmabc `wrm.connections = append(conns, conn)` should be in an `else` bock of `if err != nil {` otherwise in the case of a connection error, your are adding a `nil` connection to the list.

cpacia: good catch

core/net.go
cpacia: I'm not seeing where else this is called.

core/net.go
cpacia: Same here, not seeing it.

PR #1825, opened 2019-10-25
(#1820) Support unknown currencies in market price lookup

### LINKED ISSUE #1820: Unable to purchase a crypto currency listing
It fails with `lookup currency (ZRX): unknown currency`.

Here is the purchase data:

```
{
  "address": "",
  "addressNotes": "",
  "alternateContactInfo": "",
  "city": "",
  "countryCode": "",
  "items": [
    {
      "listingHash": "QmVuGBmhTjKkdfj9QRuSP12nzqgpY2hhKSoK4VaufnqExh",
      "options": [],
      "shipping": {
        "name": ""
      },
      "memo": "",
      "coupons": [],
      "bigQuantity": "90000000",
      "paymentAddress": "howdy"
    }
  ],
  "moderator": "",
  "paymentCoin": "BTC",
  "postalCode": "",
  "shipTo": "",
  "state": ""
}
```

placer14:
Fixes #1820

### DIFF
--- a/core/order.go
+++ b/core/order.go
@@ -1305,7 +1305,9 @@ func (n *OpenBazaarNode) getMarketPriceInSatoshis(pricingCurrency, currencyCode 
 		pricingDef, pErr  = n.LookupCurrency(pricingCurrency)
 	)
 	if cErr != nil {
-		return big.NewInt(0), fmt.Errorf("lookup currency (%s): %s", currencyCode, cErr)
+		// if the currency is unknown, it is likely a cryptocurrency and will
+		// assume the default divisibility
+		currencyDef = repo.NewUnknownCryptoDefinition(currencyCode, repo.DefaultCryptoDivisibility)
 	}
 	if pErr != nil {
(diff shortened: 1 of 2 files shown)


core/order.go
```diff
@@ -1305,7 +1305,9 @@ func (n *OpenBazaarNode) getMarketPriceInSatoshis(pricingCurrency, currencyCode
 		pricingDef, pErr  = n.LookupCurrency(pricingCurrency)
```
rmisio: What is the point of the currency lookups in this function? The only way they're used is to compare whether the codes of each cur def are equal, but can't that just be done by comparing the pricingCurrency and currencyCode arguments?

placer14: Mostly to take advantage of `currencyDef.Equal(pricingDef)` and supporting the eventuality that we will be passing around CurrencyValues (with variable divisibilities) throughout instead of the raw lookup code. `Equal` not only accounts for the code being normalized and matching, but also the divisibilities of each currency matching.

PR #1838, opened 2019-10-30
(ob-desktop #1825) Use ethereum transaction hash as is

amangale:
Earlier, there were places where the `0x` prefix was being trimmed before being sent to ob-go. This is not removed. All the txn ids/hashes/addresses will not be trimmed of their `0x` prefix.

### DIFF
--- a/net/service/handlers.go
+++ b/net/service/handlers.go
@@ -1132,7 +1132,7 @@ func (service *OpenBazaarService) handleOrderCompletion(p peer.ID, pmes *pb.Mess
 		outValue := big.NewInt(0)
 		for _, r := range records {
 			if !r.Spent && r.Value.Cmp(big.NewInt(0)) > 0 {
-				outpointHash, err := hex.DecodeString(ut.NormalizeAddress(r.Txid))
+				outpointHash, err := hex.DecodeString(r.Txid)
 				if err != nil {
 					return nil, err
 				}
--- a/vendor/github.com/OpenBazaar/go-ethwallet/util/util.go
+++ b/vendor/github.com/OpenBazaar/go-ethwallet/util/util.go
@@ -1,7 +1,6 @@
 package util
 
 import (
-	"log"
 	"math/big"
 	"reflect"
 	"regexp"
@@ -13,8 +12,11 @@ import (
 	"github.com/btcsuite/btcutil/hdkeychain"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/op/go-logging"
 	"github.com/shopspring/decimal"
 )
+
+var log = logging.MustGetLogger("ethwallet-util")
 
 // ExtractChaincode used to get the chaincode out of extended key
 func ExtractChaincode(key *hdkeychain.ExtendedKey) []byte {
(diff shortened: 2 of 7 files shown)


amangale:
> I expected these functions (https://github.com/OpenBazaar/openbazaar-go/blob/ethereum-master/util/utils.go#L66-L74) to be changed as part of this PR.
> 
> I also expected to see a pull request with the same changes in https://github.com/OpenBazaar/go-ethwallet/pulls. How are we keeping the upstream repos in sync?
> 
> Let me know what your plans are with the above items.

I dont think we need NormalizeAddress now but I will keep them till all the eth qa tests pass. As for the eth wallet repo, as of now, there is only manual pushing of changes from the ob-go to eth wallet. Clearly, there will be many more changes and once the qa tests stabilize for eth, the wallet repo will be synced manually.

net/service/handlers.go
```diff
@@ -1132,7 +1132,7 @@ func (service *OpenBazaarService) handleOrderCompletion(p peer.ID, pmes *pb.Mess
 		outValue := big.NewInt(0)
 		for _, r := range records {
 			if !r.Spent && r.Value.Cmp(big.NewInt(0)) > 0 {
-				outpointHash, err := hex.DecodeString(ut.NormalizeAddress(r.Txid))
+				outpointHash, err := hex.DecodeString(r.Txid)
```
cpacia: Does this ever have 0x in the txid? If so I think this will return an error no?

amangale: If there is a 0x prefix, there will be an error. Since the Txid itself is a chainhash.Hash (as returned by the wallet), there is no such possibility.

vendor/github.com/OpenBazaar/go-ethwallet/util/util.go
```diff
@@ -119,3 +122,29 @@ func SigRSV(isig interface{}) ([32]byte, [32]byte, uint8) {
...
+	}
+	return "0x" + str
+}
+
+// CreateChainHash is a wrapper to the chainhash.new hash function
+// this allows for a cleaner way to check if we are not in any way
+// letting the 0x prefix hinder the chainhash generation
+func CreateChainHash(str string) (*chainhash.Hash, error) {
+	hash, err := chainhash.NewHashFromStr(str)
+	if err.Error() == "max hash string length is 64 bytes" {
```
placer14: This is too brittle. Instead, let's do `if err == chainhash.ErrHashStrSize`

vendor/github.com/OpenBazaar/go-ethwallet/util/util.go
```diff
@@ -119,3 +122,29 @@ func SigRSV(isig interface{}) ([32]byte, [32]byte, uint8) {
...
+		hash, err = chainhash.NewHashFromStr(strings.TrimPrefix(str, "0x"))
+		if err != nil {
+			hash, err = chainhash.NewHashFromStr("")
```
placer14: Why empty string chainhash?

vendor/github.com/OpenBazaar/go-ethwallet/util/util.go
```diff
@@ -119,3 +122,29 @@ func SigRSV(isig interface{}) ([32]byte, [32]byte, uint8) {
...
+			if err != nil {
+				log.Println("err : ", err.Error())
```
placer14: Can we please use go-logging instead of printing to Stdout? (See https://github.com/OpenBazaar/openbazaar-go/blob/master/repo/init.go#L22)

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -563,7 +563,10 @@ func (wallet *EthereumWallet) ChainTip() (uint32, chainhash.Hash) {
 	if err != nil {
 		return 0, *h
 	}
-	h, _ = chainhash.NewHashFromStr(hash.Hex()[2:])
+	h, err = util.CreateChainHash(hash.Hex())
+	if err != nil {
+		log.Error(err)
+	}
 	return num, *h
```
placer14: It would seem that the empty string chainhash is returned to handle this return properly. I think it would be better for the ChainTip method to handle creation of the empty string chainhash instead of forcing that behavior on all of the methods which call CreateChainHash. Further, this goes against go convention where a non-nil error being returned is also providing a valid response as well. Typically if a non-nil error comes back, we want to ignore the result. It should be the same here, IMO.

Issue #1859, opened 2019-11-14
ETH payments failing when account balance is 'low' relative to the price of the item


drwasho:
### Overview

- This bug has been around for a few weeks now, but we've only recently replicated and identified what might be going wrong
- There are 2 issues related to this bug, which are covered below

#### Issue 01: Insufficient funds for attempted payments in ETH

- My balance was $0.86 and the price of the item is $0.5, with an estimated fee of $0.01
- The seller was _offline_
- The `POST /ob/purchase` call went through within a few seconds as expected
- However, the `POST /ob/orderspend` call didn't return a response until 187 seconds when it timed-out
- The following was observed in STDOUT:

```
[Haven] 09:00:44.157 [ERROR] [ethwallet/Spend] error call add txn: insufficient funds for gas * price + value
```

The 400 error response to `POST /ob/orderspend`:

![image](https://user-images.githubusercontent.com/3887694/68831234-810da580-06f9-11ea-846d-2c150d351e3a.png)

After some discussion with @amangale we think what's happening here is:

- We are estimating whether an ETH payment can be afforded on the basis of the _expected_ gas fee and not the gas limit (i.e. the maximum amount of gas we are prepared to pay to make this function call, `addTransaction()`, to the smart contract)
  - Every transfer or transaction (transactions => function calls to smart contracts in this context) requires gas to execute
  - In Ethereum, gas is an abstraction of the amount of computation effort to execute the transaction
  - But, the total gas required for a transfer/transaction varies on the gas price
  - When making a transaction, you set your gas limit higher than your expected gas fee in case of increases in demand that would prevent the execution of the transaction
- `obgo` should only attempt to make an ethereum transaction when the balance > item_cost + fee + gas_limit
  - This is what we need to investigate

#### Issue 02: Unsuccessful payment with ETH is returning a 200 OK for `POST /ob/orderspend`

- The order state is correctly set to `AWAITING_PAYMENT`, but the API should not return a 200 OK
  - Rather, it should return an 'insufficient funds'
- In this scenario, the ETH wallet balance hasn't reduced and no transaction has been made, but the server erroneously thinks that there has 

![image](https://user-images.githubusercontent.com/3887694/68832151-ebbfe080-06fb-11ea-8476-eb4755adce1b.png)

Issue #1860, opened 2019-11-14
GET /wallet/fees returns zero values for ETH


drwasho:
### Overview

Presently, `GET /wallet/fees` returns this for ETH:

<img src="https://user-images.githubusercontent.com/3887694/68832275-3fcac500-06fc-11ea-85ec-f7b9248388e2.png" width="300" />

- What you typically find in most ETH gas estimations is the gas price in Gwei (as per below)

<img src="https://user-images.githubusercontent.com/3887694/68832343-6a1c8280-06fc-11ea-8b5e-5d6982f3a46e.png" width="300" />

- To be consistent with the other wallets, we should return the value in wei
- This value would represent wei/gas, which is functionally equivalent to satoshis/byte for UTXO currencies (i.e. BTC, LTC, ZEC, BCH)

<img src="https://user-images.githubusercontent.com/3887694/68832457-b5cf2c00-06fc-11ea-9c31-5752f813eaac.png" width="300" />

---

**Informational side note**

- While the average UTXO transaction is 250 bytes or something, the average gas limit is 21,000
- The size in bytes for a UTXO multisig transaction is a bit higher than a normal transaction; similarly, the gas limit of the escrow smart contract in ETH is much higher (i.e. 4,000,000)
- The gas limit doesn't mean that's how much you'll spend in gas though
  - The gas cost for `addTransaction()`, to add funds to the escrow smart contract, is `346,041`
  - The gas cost for `execute()`, to release funds from escrow, is `141,217`

PR #1863, opened 2019-11-17
(#1859, #1860) Fix eth fees lookup 

amangale:
This PR fixes issues #1859 and #1860 for the ethereum wallet.


drwasho:
There are 2 important API calls that will affected by the problems raised in #1859 and how we're approaching to resolve them:

#### 1. `GET /wallet/estimatefee/ETH?amount=&feeLevel=normal`

- We use this API to obtain the estimated fee when making this purchase; this should not include the gas limit but only the final fee extracted from the transaction
- @amangale are we measuring this correctly for online vs offline & moderated purchases that have different gas costs due to the use of the smart contract in the latter?
- If one of the UTXO wallets doesn't have enough funds for the transaction fee, this API will return a `400 error` with the reason: `ERROR_INSUFFICIENT_FUNDS`
- **Critically**, for Ethereum this error needs to be returned for the same reason as above && if there is insufficient funds for the gas limit (which again will larger for offline and moderated purchases)
- If there is sufficient funds for the transaction, the value returned for this call **should not** include the gas limit but rather the expected transaction fee

In future we should give a breakdown of the fee vs gas limit and have more information in the error if there is insufficent funds for the gas limit; for now this is the minimum we should target.

#### 2. `GET /ob/estimatetotal`

- We use this API to obtain the final total price
- This API always returns a value even if there is insufficient funds in wallet to purchase the item
- For ETH, the value returned **should not** include the gas limit but rather the expected transaction fee

drwasho:
To test whether #1859 is successful:

- On mainnet ETH, top-up your wallet so that the balance is only a few cents above the value of the listing to purchase and the expected transaction fee (not including the gas limit)
- A good listing to purchase is a $0.5 item from `ETH seller test` => `QmRqKrN52gRZKtRqVKGVD62b3VjE2vE3Wpb2MYjo9yjhVz` or https://openbazaar.com/QmXP1aBmZWx2Goe2JAZDQXz3mpMQGgCH2G9pcHGPd9LMpA/store/eth-physical-order-testing-w-options/
- If you use another listing, the seller **must be offline** in the case of a direct purchase
- Attempt to make the purchase, and we should see the `GET /ob/orderspend` call 400 error almost immediately with an `ERROR_INSUFFICIENT_FUNDS` reason

cpacia:
Just make sure to remove all the debug logs you have in there ashwin

PR #1882, opened 2019-11-29
(#1881) Check for non nil moderator fixed fee amount 

### LINKED ISSUE #1881: ETH-Precision PATCH to Profile Fails on Moderator Info
When saving moderator info via PATCH to a node that previously had a fixed fee, a `invalid profile: percentage moderator fee should not include a fixed fee or should use (FIXED_PLUS_PERCENTAGE) feeType` error is returned even though no fixed fee data was sent.

This is happening on ethereum-master 257a3e8

Below is the payload of the call.

```
{
  "moderator": true,
  "moderatorInfo": {
    "fee": {
      "feeType": "PERCENTAGE",
      "percentage": 50
    },
    "description": "This is just a test account, do not use.",
    "termsAndConditions": "No disputes will be moderated. This is only a test account.",
    "languages": [
      "en-US"
    ]
  }
}
```

hoffmabc:
Fixes #1881 

Also attempts to read/write from the old profile schema in case the node is reading a v4 schema node's profile. The v4 schema data will not be available if the FixedFee amounts exceed the available precision. (This behavior contributes to #1904 but may not complete it.)

### DIFF
--- a/core/profile.go
+++ b/core/profile.go
@@ -215,8 +215,12 @@ func (n *OpenBazaarNode) PatchProfile(patch map[string]interface{}) error {
 		return fmt.Errorf("building profile for validation: %s", err.Error())
 	}
 
-	if err := repoProfile.Valid(); err != nil {
-		return fmt.Errorf("invalid profile: %s", err.Error())
+	if repoProfile.IsModerationEnabled() {
+		validatedFees, err := repoProfile.ToValidModeratorFee()
+		if err != nil {
+			return err
+		}
+		p.ModeratorInfo.Fee = validatedFees
 	}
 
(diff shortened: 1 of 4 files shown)


placer14:
There is something that went wrong when GH triggered a build on travis. The version its testing is complaining of a missing variable that is definitely checked in. (See https://travis-ci.org/OpenBazaar/openbazaar-go/builds/623302859#L236 and https://github.com/OpenBazaar/openbazaar-go/blob/1818caf8f1477b4a126b9cdef34f01130aa6b375/repo/profile.go#L18)

I'm going to try to close and recreate this PR to see if the problem can be resolved this way.

Issue #1879, opened 2019-11-27
ETH-Precision PATCH profile returns nothing


jjeffryes:
When doing a PATCH on profile to save moderator info, nothing is returned. 

The data below was sent, and the status of the call was 200, but nothing was returned which causes an error in the client. We expect at least {} to be returned on calls.

This is happening on ethereum-master 257a3e84 

```
{
  "moderator": false,
  "moderatorInfo": {
    "fee": {
      "feeType": "PERCENTAGE",
      "percentage": 5
    },
    "description": "test",
    "termsAndConditions": "test",
    "languages": [
      "en-US"
    ]
  }
}
```

hoffmabc:
@jjeffryes I think this is out of date since the call has changed. 

{ 
   "moderator": false,
   "moderatorInfo":{ 
      "fee":{ 
         "feeType":"PERCENTAGE",
         "fixedFee":{ 
            "bigAmount": "0",
            "amountCurrency": { 
               "code":"USD",
               "divisibility": 2
            }
         },
         "percentage": 5
      },
      "description": "test",
      "termsAndConditions":"asdf",
      "languages":[ 
         "en-US"
      ]
   }
}

jjeffryes:
@hoffmabc I'm not sure what you mean, I have the latest code for the desktop and server, it is not sending a fixed fee object, and is getting nothing as a response.

jjeffryes:
This is fixed on ethereum master

PR #1884, opened 2019-11-29
Don’t require both fixed and percentage fee for moderator info

hoffmabc:
Don’t require clients to explicitly include fixed fee and percentage settings.

### DIFF
--- a/repo/profile.go
+++ b/repo/profile.go
@@ -64,7 +64,23 @@ func ProfileFromProtobuf(p *pb.Profile) (*Profile, error) {
 			return nil, ErrMissingModeratorFee
 		}
 
-		if fees.FixedFee != nil {
+		if fees.FeeType == pb.Moderator_Fee_FIXED_PLUS_PERCENTAGE {
+			amtCurrency, err := AllCurrencies().Lookup(fees.FixedFee.AmountCurrency.Code)
+			if err != nil {
+				return nil, fmt.Errorf("lookup currency (%s): %s", fees.FixedFee.AmountCurrency, err.Error())
+			}
+			amtCurrency.Divisibility = uint(fees.FixedFee.AmountCurrency.Divisibility)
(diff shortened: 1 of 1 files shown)


repo/profile.go
```diff
@@ -58,24 +58,50 @@ type Profile struct {
...
-
-		modInfo = &ModeratorInfo{
-			Fee: &ModeratorFee{
-				FeeType: fees.FeeType.String(),
-				FixedFee: &ModeratorFixedFee{
-					Amount:         fees.FixedFee.BigAmount,
-					AmountCurrency: amtCurrency,
+
+		if fees.FeeType == pb.Moderator_Fee_FIXED_PLUS_PERCENTAGE {
+			amtCurrency, err := AllCurrencies().Lookup(fees.FixedFee.AmountCurrency.Code)
```
cpacia: I think you need a nil check on `AmountCurrency` here else it might panic.

repo/profile.go
```diff
@@ -58,24 +58,50 @@ type Profile struct {
...
+						Amount:         fees.FixedFee.BigAmount,
+						AmountCurrency: amtCurrency,
+					},
+					Percentage: fees.Percentage,
 				},
-				Percentage: fees.Percentage,
-			},
+			}
+		} else if fees.FeeType == pb.Moderator_Fee_FIXED {
+			amtCurrency, err := AllCurrencies().Lookup(fees.FixedFee.AmountCurrency.Code)
```
cpacia: Same here: `AmountCurrency` nil check

placer14:
The changes to normalize the fixed fee protobuf were fixed as part of https://github.com/OpenBazaar/openbazaar-go/pull/1912. I also see the branch name appears to point to #1879 but does not appear to fix the problem directly or test that it's resolved. Since this PR's work is a dupe of PR 1912, I'm going to close this and reopen a new one which tests that 1879 is resolved.

Issue #1862, opened 2019-11-14
Direct purchase hangs


rmisio:
Not sure if others could reproduce this, but this has happened to me dozens of times. And it's pretty mysterious because sometimes it happens and other times with seemingly the same data, it doesn't happen.

Anyhow, make a direct order. The vendor seems to process it properly, aside from maybe `Currency not tracked`:

```
13:19:25.592 [DEBUG] [service/handleOrder] handling normal order from: QmUhBVZDCcY9RzhnYXbQxUFETCG8ECvHcRfSGHiR5pecSJ
13:19:25.594 [DEBUG] [service/handleOrder] incoming order linked to LTC wallet
13:19:25.594 [DEBUG] [service/handleOrder] received direct online order from QmUhBVZDCcY9RzhnYXbQxUFETCG8ECvHcRfSGHiR5pecSJ
13:19:26.511 [ERROR] [core/CalculateOrderTotal] Currency not tracked
13:19:26.514 [DEBUG] [service/handleOrder] storing sales order Qmdwv643wnThj5mrGy6tmAkC9q5Pi4MVi85JYukzfpjwbp into the database and awaiting payment
13:19:26.516 [DEBUG] [service/handleOrder] sending order confirmation message to QmUhBVZDCcY9RzhnYXbQxUFETCG8ECvHcRfSGHiR5pecSJ
13:19:26.516 [DEBUG] [service/handleOrder] received addr-req ORDER message from QmUhBVZDCcY9RzhnYXbQxUFETCG8ECvHcRfSGHiR5pecSJ
13:19:26.516 [DEBUG] [service/handleNewMessage] sending response message to: QmUhBVZDCcY9RzhnYXbQxUFETCG8ECvHcRfSGHiR5pecSJ
```

The buyer appears to receive the response:

```
13:19:25.589 [ERROR] [core/CalculateOrderTotal] Currency not tracked
13:19:25.591 [DEBUG] [service/SendRequest] Sending ORDER request to QmeRfQcEiefLYgEFRsNqn1WjjrLjrJVAddt85htU1Up32y
13:19:25.591 [DEBUG] [service/messageSenderForPeer] Found existing message sender for: QmeRfQcEiefLYgEFRsNqn1WjjrLjrJVAddt85htU1Up32y
13:19:26.517 [DEBUG] [service/handleNewMessage] Received new message from: QmeRfQcEiefLYgEFRsNqn1WjjrLjrJVAddt85htU1Up32y
[...]

Issue #1861, opened 2019-11-14
ob/purchase call hangs when offline order send times out


rmisio:
1.) Attempt to make an offline purchase
2.) Simulate a slow connection or otherwise cause the order request to time out:
`12:50:28.029 [ERROR] [core/SendOrder] failed to send order request: timed out writing message`

The `ob/purchase` API call never comes back with an error. It just hangs until you force kill the server.

hoffmabc:
@rmisio I have gone through and hard-coded timed out orders and timed out ticker endpoints to replicate this issue and even on a 0.1mbps bad connection it was still going through eventually.

I'm not sure how else to replicate this. Perhaps there is something else causing the hang here.

hoffmabc:
If currency is not tracked, that just means it falls back to the default fee value.

rmisio:
ok, then I don't know why this is happening for me and fairly regularly. Is it possible it would still go through after more than 10 minutes (I'm pretty sure I've waited at least that long)?

It looks at least one other person is seeing this:

https://github.com/OpenBazaar/openbazaar-desktop/issues/1779

rmisio:
@hoffmabc Yeah, I don't know if this is a node having the "currency not tracked" issue. I do know the logs clearly show this:

`12:50:28.029 [ERROR] [core/SendOrder] failed to send order request: timed out writing message`

Which looks like a failure that prevents the purchase. But... this failure is never bubbled back to the UI via a 500 error on the API. Instead, the call just hangs.

Can you find in the code where that log message would happen, simulate an error and if you're able to replicate the order hanging, fix it so the call comes back with an error? Otherwise, it's a horrible UX where the user has a ghost never-ending HTTP call clogging up their request queue, until they restart.

hoffmabc:
All that error means is that an online purchase did not work.

rmisio:
@hoffmabc You may be conflating this bug with another bug. This bug is specifically about an offline order failing.

> All that error means is that an online purchase did not work and yes this is what I did. I forced it to error on that.

Ok, and did the purchase API call return an error to the client... or does it just remain in pending state?

PR #1898, opened 2019-12-05
[WIP] Revert custom transports to default transports

hoffmabc:
This is a WIP experiment to mitigate against the multiple timeout issues we're seeing when many requests are being made at once by any specific http.Client. We have anecdotal evidence that removal of the default `net.Dial` used with `http.Client{Transport}` has created some connectivity issues with many requests in-flight at once.

Will consider this for merge after running internal tests with this changeset.

### DIFF
--- a/core/images.go
+++ b/core/images.go
@@ -168,11 +168,15 @@ func (n *OpenBazaarNode) FetchImage(peerID string, imageType string, size string
 // GetBase64Image - fetch the image and return it as base64 encoded string
 func (n *OpenBazaarNode) GetBase64Image(url string) (base64ImageData, filename string, err error) {
 	dial := net.Dial
+	var client *http.Client
 	if n.TorDialer != nil {
 		dial = n.TorDialer.Dial
+		tbTransport := &http.Transport{Dial: dial}
+		client = &http.Client{Transport: tbTransport, Timeout: time.Second * 30}
+	} else {
+		client = &http.Client{Timeout: time.Second * 30}
 	}
-	tbTransport := &http.Transport{Dial: dial}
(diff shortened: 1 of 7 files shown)


core/images.go
```diff
@@ -168,11 +168,15 @@ func (n *OpenBazaarNode) FetchImage(peerID string, imageType string, size string
 // GetBase64Image - fetch the image and return it as base64 encoded string
 func (n *OpenBazaarNode) GetBase64Image(url string) (base64ImageData, filename string, err error) {
 	dial := net.Dial
+	var client *http.Client
 	if n.TorDialer != nil {
 		dial = n.TorDialer.Dial
+		tbTransport := &http.Transport{Dial: dial}
```
placer14: http.Transport.Dial is deprecated and prefers to use DialContext instead. This is fine for testing though as it only affects the Tor case which we aren't testing with at the moment.

hoffmabc: is the tordialer using .dial?

placer14: It's possible and might also be way those resources aren't getting released as expected.

hoffmabc:
Deals with #1861 #1862

core/images.go
cpacia: Yeah proxy does not offer a DialContext method

placer14:
I checked and the difference between `Dial` and `DialContext` is:
```
func (d *Dialer) Dial(network, address string) (Conn, error) {
	return d.DialContext(context.WithTimeout(30 * time.Second), network, address)
}
```
Do we want to add this so we don't have to use the deprecated `http.Transport.Dial` interface anymore?

Issue #1711, opened 2019-09-23
Cleanup repo.Listing abstraction


placer14:
Lots of repeated code throughout `repo.Listing` with an unclear pattern of handling the protobuf and conversion to/from JSON ensuring that mutated changes aren't lost. Refactoring will commit the following principles:

- SignedListing is readonly
- Listing can (Listing).Sign() (SignedListing, error) or (Listing).Valid() error
- mutation will be managed by pb.Listing object which will be expressed via (Listing).MarshalJSON() ([]byte, error) and UnmarshalJSON(Signed)Listing([]byte) ((Signed)Listing, error)
- Values into/out of Listing use getters and setters to ensure the proto state is kept consistent

Unhandled guarantees may leak pointers and as such, the abstraction is still not threadsafe...but this was the case with protobufs in the first place. As long as we're mindful of this limitation, we should not introduce accidental mutation bugs of this nature.

PR #1906, opened 2019-12-06
(#1711) Refactor repo.(Signed)Listing

placer14:
This applies the recommendations outlined in #1711 with some followup work still due. Most notable changes include: 
- move core/listings.go behavior back into repo.Listing where possible
- prefer to use functions which are associated with *OpenBazaarNode rather than reach into their struct's state to use node state directly (such as `GetNodeID` and `Sign`)
- getter/setter pattern around repo.Listing to ensure that the underlying protobuf remains in-sync with the repo.Listing that wraps it. (Note: I'm inconsistently using `Get<Object>()` and `<Object>()` as the name of the getter and will follow up to make them all `<Object>()` without the Get prefix.)
- started creating dependent objects (like `repo.ListingVariant` which we'll use to access SKUs, Inventory, etc) which are not consistently used yet, but wanted to present that pattern. This will help us organize all the parts where we need to remember which index a SKU had and other implementation details (validation, relationships between options and prices, and (most importantly) pricing schema changes.

Wanted to share this for review earlier than later, but here are a few tasks I still need to complete before wrapping this PR:
- [x] merge in latest `ethereum-master` changes since the start of this refactor
- [ ] analyze how the refactoring (and the resulting consistant application of validations across all endpoints which mutate listings) will affect the API (if at all)
- [x] update the `Get<Object>()` calls to be `<Object>()`
- [x] ~double-check that the QAs pass with this refactoring~ Will resolve QA test failures separate from this PR.


test/factory/peer_info.go
```diff
@@ -36,35 +29,32 @@ func NewPubkeysIdentityKeyBytes() []byte {
 func NewPubkeysProtobuf() *pb.ID_Pubkeys {
 	return &pb.ID_Pubkeys{
 		Identity: NewPubkeysIdentityKeyBytes(),
-		//Bitcoin:  []byte("AwD4y8eIx7F0bnwNmssZGi+XFqydypxuFRtA4TPyWiqJ"),
+		Bitcoin:  []byte("AwD4y8eIx7F0bnwNmssZGi+XFqydypxuFRtA4TPyWiqJ"),
```
cpacia: Pretty sure this wont validate. Is the goal to make it validate?

placer14: Yeah, trying to get some raw factory data which works through the normal validation.

cpacia: To make it validate you will need to do base64.stdencoding.Decode()

core/listings.go
```diff
@@ -140,86 +98,104 @@ func (n *OpenBazaarNode) UpdateListing(r []byte, publish bool) error {
...
-		if err != nil {
+		if err := l.SetCryptocurrencyListingDefaults(); err != nil {
 			return err
 		}
 	}
 
-	err = n.SetListingInventory(listing)
-	if err != nil {
+	if err := n.validateListingIsSellable(l); err != nil {
+		return fmt.Errorf("validate sellable listing (%s): %s", l.GetSlug(), err.Error())
```
cpacia: This line is what's causing the API test to fail. Basically this used to return a type `repo.ErrPriceModifierOutOfRange` but this changes it to return a generic error. 

We can either modify the API test to just check the string representation of the expected error or we need to create a `ErrSellableListingError`

test/factory/peer_info.go
cpacia: Looking in to this, `NewPubkeysProtobuf` is used by `NewPeerIDProtobuf` which is not used in the code at all. Should we remove both?

placer14: Thanks for comments here. Will resolve this in a separate PR.

Issue #1904, opened 2019-12-06
On GET /ob/profile/<peer-id> convert profile to the latest schema


rmisio:
I'm not sure if this is the only thing requiring conversion, but mod fees need to be updated (e.g. an integer fixed fee `amount` of 1 should be a `bigAmount` of `"1"`. Also, `currencyCode` should now be a full currency definition in `amountCurrency`).

hoffmabc:
Profile needs to be normalized for the new version (i.e. moderation info).

hoffmabc:
- Profile schema still includes old attributes so we can just include them for v4 nodes so that they don't bomb against reading v5 profiles
- This fix requires a migration of v5 users profiles to v5 format AND normalization for reading profiles.

placer14:
FYI, https://github.com/OpenBazaar/openbazaar-go/pull/1882 contains the normalization behavior for reading profiles.

PR #1912, opened 2019-12-10
(#1881) Check for non nil moderator fixed fee amount

### LINKED ISSUE #1881: ETH-Precision PATCH to Profile Fails on Moderator Info

placer14:
Fixes #1881 

Also attempts to read/write from the old profile schema in case the node is reading a v4 schema node's profile. The v4 schema data will not be available if the FixedFee amounts exceed the available precision. (This behavior contributes to #1904 but may not complete it.)

### DIFF
--- a/core/profile.go
+++ b/core/profile.go
@@ -215,8 +215,12 @@ func (n *OpenBazaarNode) PatchProfile(patch map[string]interface{}) error {
 		return fmt.Errorf("building profile for validation: %s", err.Error())
 	}
 
-	if err := repoProfile.Valid(); err != nil {
-		return fmt.Errorf("invalid profile: %s", err.Error())
+	if repoProfile.IsModerationEnabled() {
+		validatedFees, err := repoProfile.ToValidModeratorFee()
+		if err != nil {
+			return err
+		}
+		p.ModeratorInfo.Fee = validatedFees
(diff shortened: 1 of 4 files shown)


placer14:
No, the error was legit and I had some bad state. Closing in order to prefer original PR (https://github.com/OpenBazaar/openbazaar-go/pull/1882). :|

placer14:
Nevermind, the original PR is now locked form being reopened.

placer14:
Merging via approval provided at https://github.com/OpenBazaar/openbazaar-go/pull/1882#pullrequestreview-330074386

Issue #1783, opened 2019-10-14
Improve ETH Spend to sync return more quickly or to become entirely async


placer14:
As a stopgap, we have allowed the ETH Spend method to take as long as 3 minutes waiting for the transaction to be mined by the network as there is no other way to ensure whether the transaction is successful or not until then. We would either improve this to emit a websocket message result for ETH transactions which would allow the Spend request to return the client response more quickly while still tracking the result of the Spend.

This improvement should be discussed with desktop and mobile client integration teams.

rmisio:
> We would either improve this to emit a websocket message result for ETH transactions which would allow the Spend request to return the client response more quickly while still tracking the result of the Spend.

If we do that, I would make that configurable via an api parameter. Some clients may prefer to wait than have a disjointed response.

(see `async` on `ob/fetchprofiles`)

Secondly, for the socket option, the client should be allowed to pass in a socket id that the socket will return and that would be used as the mapping between the API call and the socket response. We found this pattern worked well as opposed to the server generating and sending the socket id in the API response, since we found that there were sometimes race conditions where the socket arrived before the API returned.

(see `asyncID` on `ob/fetchprofiles`)

placer14:
Thanks for the `fetchprofile` examples, agree that we should follow prior patterns if they've worked well for us.

drwasho:
Fixed in PRs 1915, 1951, and 1985.

PR #1915, opened 2019-12-11
(#1783) Make ETH transactions asynchronous

amangale:
This PR fixes https://github.com/OpenBazaar/openbazaar-go/issues/1783.
All the eth transactions now return immediately and the status is updated asynchronously.


drwasho:
Need to do some additional testing before merging, which is why we added the `[WIP]`

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -809,16 +706,22 @@ func (wallet *EthereumWallet) CheckTxnRcpt(hash *common.Hash, data []byte) (*com
 			if err != nil {
 				return nil, err
 			}
-			wallet.db.Txns().Delete(chash)
-			toAddr := common.HexToAddress(util.EnsureCorrectPrefix(pTxn.To))
+			err = wallet.db.Txns().Delete(chash)
```
placer14: It looks like the arg for `Delete` is generated out of `util.CreateChainHash((*hash).Hex())` but the arg which `Delete` is matching on is generated out of `ut.NormalizeAddress(hash.Hex())`. Are these going to produce the same value so that the Putted data will actually get Deleted?

placer14: We also completely ignore the `err` set here, so if the Delete fails (say, because the arg doesn't exist) we'd never know.

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -788,7 +686,6 @@ func (wallet *EthereumWallet) AssociateTransactionWithOrder(txnCB wi.Transaction
 
 // CheckTxnRcpt check the txn rcpt status
 func (wallet *EthereumWallet) CheckTxnRcpt(hash *common.Hash, data []byte) (*common.Hash, error) {
```
placer14: Can we remove the `data` arg entirely and instead use `hash` to `wallet.db.Txns().Get(hash)` so we aren't accidentally operating on a different version of `data` from one execution path to another?

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
```diff
@@ -809,16 +706,22 @@ func (wallet *EthereumWallet) CheckTxnRcpt(hash *common.Hash, data []byte) (*com
...
+				toAddr := common.HexToAddress(util.EnsureCorrectPrefix(pTxn.To))
+				go wallet.AssociateTransactionWithOrder(
+					wallet.createTxnCallback(util.EnsureCorrectPrefix(hash.Hex()), pTxn.OrderID, EthAddress{&toAddr},
+						*n, time.Now(), pTxn.WithInput))
+			} else {
+				addrScrHash := common.HexToAddress(pTxn.To)
+				go wallet.AssociateTransactionWithOrder(wallet.createTxnCallback(util.EnsureCorrectPrefix(hash.Hex()),
+					pTxn.OrderID, EthAddress{address: &addrScrHash}, *n,
+					time.Now(), true))
+			}
```
placer14: Don't repeat the `AssociateTransactionWithOrder` in both branches. Instead, make the change that `if pTxn.Amount != "0" {` is interested in cleared by adjusting a `var hasInput bool` value can calling the whole method after the branch.

placer14: I'm generally concerned that `AssociateTransactionWithOrder` can run indefinitely here and there'd be no way for us to clean this goroutine up if we wanted to stop the parent threads.

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
amangale: The key for delete matches.

placer14: Okay. Let's still handle the error, though.

vendor/github.com/OpenBazaar/go-ethwallet/wallet/wallet.go
amangale: This func is now internal and called from one place inside the wallet only.

vendor/github.com/nanmu42/etherscan-api/account.go
```diff
@@ -97,6 +97,12 @@ func (c *Client) ERC20Transfers(contractAddress, address *string, startBlock *in
 	compose(param, "startblock", startBlock)
 	compose(param, "endblock", endBlock)
 
+	if desc {
+		param["sort"] = "desc"
+	} else {
+		param["sort"] = "asc"
+	}
+
```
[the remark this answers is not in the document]
placer14: Looks like there was already a patch applied. Getting this updated in the vendors from source.

placer14:
Updated description to indicate that this PR does _not_ resolve https://github.com/OpenBazaar/openbazaar-go/issues/1783 as originally indicated.

PR #1925, opened 2019-12-13
(#1861, 1862) Update vendored transports for Master

hoffmabc:
Fix to mitigate against #1861 and #1862.


hoffmabc:
@placer14 I looked into wrapping the proxy.dialer but we have an issue because we have vendored an older golang.org/x/net/proxy lib and it does not include the ContextDialer code needed.

`https://github.com/golang/net/blob/master/proxy/dial.go#L13`

PR #1926, opened 2019-12-16
Update zcash branchID

cpacia:


hoffmabc:
The Travis issue is resolved in my transport PR so we don't need to deal with the linting issue here.

PR #1929, opened 2019-12-17
Re-connect to Watched Addresses on Websocket Reconnection

hoffmabc:
Check this out @placer14 for approach.

### DIFF
--- a/vendor/github.com/OpenBazaar/multiwallet/client/blockbook/client.go
+++ b/vendor/github.com/OpenBazaar/multiwallet/client/blockbook/client.go
@@ -100,7 +100,7 @@ type BlockBookClient struct {
 	apiUrl            *url.URL
 	blockNotifyChan   chan model.Block
 	closeChan         chan<- error
-	listenLock        sync.Mutex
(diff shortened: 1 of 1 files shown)


vendor/github.com/OpenBazaar/multiwallet/client/blockbook/client.go
```diff
@@ -556,6 +559,9 @@ func (i *BlockBookClient) setupListeners() error {
 		return nil
 	}
 
+	// Add stored watch addresses to listenQueue if there are any
+	i.listenQueue = append(i.listenQueue, i.listenAddrs...)
```
placer14: Access to `listenAddrs` needs to be protected by a mutex so this read isn't in a race condition with the write in `ListenAddresses`. If you change the `listenLock` to a `RWMutex` and use them appropriately in each place, that should be sufficient.

hoffmabc: Ok commit has been submitted to fix this.

PR #1934, opened 2019-12-17
Do not mark wallet address as watched + fix earlier txn errors

### LINKED ISSUE #1918: Multiwallet stops processing coin network updates for direct payment addresses
## Overview

We are exploring a bug where the multiwallet appears to be incorrectly treating addresses which are managed by the internal wallet as watch-only....a state which ultimately prevents future network updates (txs creating spendable UTXOs, inbound payments, etc) from being processed on these addresses and updated on the user's balance. This also prevents orders which expect these updates from progressing.

## Description

The bug manifests via the following scenario:

1. An order is placed by a Buyer and an ORDER message is transmitted to the Vendor. Vendor receives this message and marks the payment addr as an address to watch for new txs from blockbook. This case occurs where the Buyer and Vendor are both online which is necessary for the Buyer to query the Vendor on the appropriate payment address to use. (Notably: in the offline case, a 1-of-2 address is generated by the Buyer which falls outside of this bug's scenario entirely.)
2. One of two things happen at this point which ultimately allow the order to be captured by the ResyncManager:
    - `2a.` The websocket has died and restarted as a result of the new payment addr being added to the watch list. All previously subscribed addrs have been dropped and are not being reported on. (Details at: https://github.com/OpenBazaar/multiwallet/issues/106)
    - `2b.` The ResyncManager (which periodically runs every 15 minutes) scoops up all unfunded orders and ensures they are currently being watched. If (2a) hasn't occurred but (2b) occurs before the payment naturally propagates through the network and ultimately to the client, the Resync Manager will treat the order from (1) as unfunded.
3. When the ResyncManager processes the unfunded order, it incorrectly handles the payment address as a watch-only address and persists this addr in the DB as being a watchOnly addr forever more. (This issue is created, at least, by https://github.com/OpenBazaar/multiwallet/issues/107 but there could be other causes for this problem.)

After (3), the order moves into a state that the ob-go does not know how to recover from. Further, in this state, all future utxos which result from funds moving from this address are not persisted to the DB...essentially becoming a dead address despite the funds being properly moved on the network.

### Context

Watch-only addresses are typically reserved for moderated orders which use a 2-of-3 escrow address and are spent via escrow release rather than the normal spend flow of the wallet. Since escrow releases have a separate flow for tracking how those utxos get credits to the user's balance, these incorrectly-labeled watch-only addresses will never see either the normal/direct resolution of direct utxo payment, nor will it see the escrow-specific resolution.

## Acceptance criteria for fix

- [ ] 1. Protection of watched addresses from falling into the above unrecoverable state needs to be addressed. This is partially addressed by the unit test and validation recommended in https://github.com/OpenBazaar/multiwallet/issues/107
- [x] 2. Cleaning of WatchedScripts database (via migration or some idempotent runtime process) to ensure any bad data doesn't acceidentally leak into other parts of the app. (See https://github.com/OpenBazaar/openbazaar-go/issues/1918#issuecomment-565132346)
- [x] 3. Resubscribe addresses on websocket reconnection (https://github.com/OpenBazaar/multiwallet/issues/106)
- [x] 4. Consideration of potential redesign of our abstractions or usage of existing abstractions to help future devs not accidentally create the scenario again.

PS: Huge thanks to @rodkeys who helped significantly in the investigation into this problem. This was a tricky one and his help was instrumental in finding the needle in the proverbial haystack.

rodkeys:
These changes satisfy concerns in #1918 and rescues/corrects old order transactions

### DIFF
--- a/vendor/github.com/OpenBazaar/multiwallet/service/wallet_service.go
+++ b/vendor/github.com/OpenBazaar/multiwallet/service/wallet_service.go
@@ -522,7 +522,7 @@ func (ws *WalletService) saveSingleTxToDB(u model.Transaction, chainHeight int32
 	cb.Value = value
 	cb.WatchOnly = (hits == 0)
 	saved, err := ws.db.Txns().Get(*txHash)
-	if err != nil {
+	if err != nil || saved.WatchOnly != cb.WatchOnly {
 		ts := time.Now()
 		if u.Confirmations > 0 {
 			ts = time.Unix(u.BlockTime, 0)
@@ -619,7 +619,9 @@ func (ws *WalletService) getStoredAddresses() map[string]storedAddress {
 			}
 			addr = ltcAddr
 		}
-		addrs[addr.String()] = storedAddress{addr, true}
+		if _, ok := addrs[addr.String()]; !ok {
+			addrs[addr.String()] = storedAddress{addr, true}
+		}
 	}
 
 	return addrs


vendor/github.com/OpenBazaar/multiwallet/service/wallet_service.go
```diff
@@ -522,7 +522,7 @@ func (ws *WalletService) saveSingleTxToDB(u model.Transaction, chainHeight int32
 	cb.Value = value
 	cb.WatchOnly = (hits == 0)
 	saved, err := ws.db.Txns().Get(*txHash)
-	if err != nil {
+	if err != nil || saved.WatchOnly != cb.WatchOnly {
```
placer14: If `err != nil` then `saved` should not be expected to be valid. Did you mean `err != nil && ...`?

rodkeys: Answered on slack, reposting here for the records.

"I don't believe so, "ws.db.Txns().Get(*txHash)" returns an err on being unable to get the tx/saved variable for any reason.

In the first scenario where the tx doesn't exist, it would hit the first check in the or statement and continue. 

In the second scenario, where the tx exists and the WatchOnly parameter is wrong, the err would be nil but "saved" would exist. So it should check the second case in the or statement and then continue."

placer14: Yes, you're right. My mistake. 🤝

placer14: This seems to fix this one place, but will allow our bad data to infect at least two other places:

1. vendor/github.com/OpenBazaar/multiwallet/util/balance.go within `CalcBalance` which omits `WatchOnly` txs so the balance will be misrepresented. The impact is incorrect balances from `/wallet/balance`.
2. repo/db/txns.go within `GetAll` which flags whether certain records should or shouldn't be returned for downstream usage. The impact is missing transactions from `/wallet/transactions`.

We either need to handle the two above cases or we need to reconsider the data migration approach.

placer14: Okay, we can disregard. I misunderstood where this check was being added and arrived at the wrong conclusion. I do note that this will update the transaction record with a freshly calculated state.

szollo:

PR #1940, opened 2019-12-19
[#1681] Protect POSTReleaseEscrow sweep when there are no inputs

### LINKED ISSUE #1681: Order in FULFILLED state with no funds pending cannot completed by vendor
In the scenario that a vendor has an order that is in a FULFILLED state but the inputs to the payment address have already been spent (unsure of how this state was achieved), the vendor gets a 400 Bad Request from the wallet API due to have no inputs for sweeping.

We want to bypass the Sweep attempt if all inputs are spent to allow the order to progress to PROCESSING_ERROR with text associated indicating that there are no inputs to process to FINALIZE_PAYMENT.

hoffmabc:
This issue fixes the scenario where the BUYER released the funds but the SELLER did not receive the completion message (but likely did receive the coins).

### DIFF
--- a/core/completion.go
+++ b/core/completion.go
@@ -349,27 +349,29 @@ func (n *OpenBazaarNode) ReleaseFundsAfterTimeout(contract *pb.RicardianContract
 		}
 	}
 
-	if len(txInputs) != 0 {
-		chaincode, err := hex.DecodeString(contract.BuyerOrder.Payment.Chaincode)
-		if err != nil {
-			return err
-		}
-		mECKey, err := n.MasterPrivateKey.ECPrivKey()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


szollo:

hoffmabc:
Is this worth putting into 0.13.8 @placer14 ?

placer14:
@hoffmabc This is worth considering. Is there a reproduceable test scenario we can do before/after this patch to ensure we're making a useful change?

hoffmabc:
It's been reported in the wild, but I'm not sure of a reproducible scenario. I will see if I can replicate it on my side.

hoffmabc:
Ready for review @placer14

PR #1947, opened 2019-12-20
Add version to profile protobuf

### LINKED ISSUE #1923: Add version to profile protobuf
Add version field to profile protobuf to indicate which schema should be used. This is a flag that any consuming process can switch on to determine which set of data fields are canonical. For example, between schema v4 and v5 we've changed the `ModeratorInfo.Fee.FixedFee` schema.

cpacia:
Fixes #1923


szollo:

placer14:
This PR needs to point to `master` instead of `ethereum-master`. Need to force-push over the existing merge branch with just the protobuf fixed cherry-picked without the other `eth-master` changes.

PR #1949, opened 2019-12-23
(#1889) Refactor Pricing Conversion with Big.Int/Float

### LINKED ISSUE #1889: ETH Precision- ETH to LTC conversion is incorrect on purchase
This is happening on Dispute payout precision OpenBazaar/openbazaar-desktop#1838 and ethereum-master latest as of 11-27-19.

I made a purchase of an item at a price of 0.0000002 ETH in LTC (the item price is 0.0000000001, the quantity was 2000).

The receipt on the purchase page showed a total of 0.00000064 LTC, which seems right.

But the purchase call had an amount of 314000, or 0.00314 LTC when it was submitted to the server.

This is the listing `ob://QmVF14jczeRFii8ptcvXScFVkHWCg9WnThU2dPoqqDjLep/store/test-eth-precision`

This is the data sent:

```
{
  "address": "",
  "addressNotes": "",
  "alternateContactInfo": "",
  "city": "",
  "countryCode": "",
  "items": [
    {
      "listingHash": "QmVC7NMWYQ9NwQgS9ZKmns99xhqsQH4z9kF6HM1ARoKZ1G",
      "options": [],
      "shipping": {
        "name": ""
      },
      "memo": "",
      "coupons": [],
      "bigQuantity": "2000"
    }
  ],
  "moderator": "",
  "paymentCoin": "LTC",
  "postalCode": "",
  "shipTo": "",
  "state": ""
}
```

This is the response:
```
{
    "amount": {
        "amount": "314000",
        "currency": {
            "code": "LTC",
            "currencyType": "crypto",
            "divisibility": 8,
            "name": "Litecoin"
        }
    },
    "orderId": "QmWT4gdyQSvEH4aDaodtBgkNbUUafpT8g5v1d9iLYPVgXP",
    "paymentAddress": "ltc1q7frr4ywwc8a4alntwqlznrug57httajf63mwxqxv8h3mu8rw799s4kcghe",
    "vendorOnline": false
}
```

hoffmabc:
Closes #1889

### DIFF
--- a/core/order.go
+++ b/core/order.go
@@ -1200,10 +1200,10 @@ func (n *OpenBazaarNode) calculateShippingTotalForListings(contract *pb.Ricardia
 	return shippingTotal, nil
 }
 
-func (n *OpenBazaarNode) getPriceInSatoshi(paymentCoin, currencyCode string, amount *big.Int) (*big.Int, error) {
+func (n *OpenBazaarNode) getPriceInSatoshi(paymentCoin, originCurrencyCode string, amount *big.Int) (*big.Int, error) {
 	var reserveCurrency = n.reserveCurrency()
 	var (
-		originCurrencyDef, oErr  = n.LookupCurrency(currencyCode)
+		originCurrencyDef, oErr  = n.LookupCurrency(originCurrencyCode)
 		paymentCurrencyDef, pErr = n.LookupCurrency(paymentCoin)
 	)
 	if oErr != nil {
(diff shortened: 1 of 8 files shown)


hoffmabc:
@placer14 I took a look at your other branch concerning this code and I really started to think that having a straight CurrencyConverter object might be smarter than having the ExchangeRater object. Take a look at what I've done, I'd like to hear some feedback on this. I also removed the GetMarketPriceInSatoshis which was basically redundant code except for handling unknown crypto types. I refactored it into the other function.

core/order.go
```diff
@@ -1319,47 +1302,6 @@ func (n *OpenBazaarNode) reserveCurrency() string {
...
-	if err != nil {
-		return nil, err
-	}
-	newCV, _, err := cv.ConvertTo(pricingDef, 1/rate)
-	if err != nil {
-		return nil, err
-	}
-	return newCV.Amount, nil
-}
-
```
placer14: Love this! 👏

net/retriever/retriever.go
```diff
@@ -67,9 +68,11 @@ type offlineMessage struct {
 }
 
 func NewMessageRetriever(cfg MRConfig) *MessageRetriever {
+	dial := gonet.Dial
 	var client *http.Client
 	if cfg.Dialer != nil {
-		tbTransport := &http.Transport{Dial: cfg.Dialer.Dial}
+		dial = cfg.Dialer.Dial
+		tbTransport := &http.Transport{Dial: dial}
```
placer14: These transport changes shouldn't be repeated. Maybe rebase just the converter changes against lastest ethereum-master to ensure changes aren't leaking in.

core/order.go
```diff
@@ -1249,67 +1249,50 @@ func (n *OpenBazaarNode) calculateShippingTotalForListings(contract *pb.Ricardia
...
-		return big.NewInt(0), fmt.Errorf("converting from reserve: %s", err.Error())
+		return nil, err
 	}
-	return resultValue.Amount, nil
+
+	// Convert to satoshis
+	divisibility := new(big.Int).SetUint64(uint64(paymentCurrencyDef.Divisibility))
+	divisibilityFactor := new(big.Int).Exp(new(big.Int).SetInt64(10), divisibility, nil)
+	finalPrice.Mul(finalPrice, new(big.Float).SetInt(divisibilityFactor))
+	finalPriceInt, _ := finalPrice.Int(new(big.Int))
```
placer14: Looks like you do all this as part of `GetFinalPrice`. Let's make sure we don't apply it twice.

repo/currency_converter.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// CurrencyConverter
+type CurrencyConverter struct {
+	ReserveWallet wallet.Wallet
+}
+
+func NewCurrencyConverter(reserveWallet wallet.Wallet) CurrencyConverter {
+	return CurrencyConverter{
+		reserveWallet,
```
placer14: We should always label field names in explicit structs so the ordering is non-specific and the constructions are well-described instead of having to lookup the struct def.

hoffmabc: Okay @placer14 but I believe there's only one attribute of the struct to begin with...

hoffmabc: image

net/retriever/retriever.go
hoffmabc: ✅

core/order.go
hoffmabc: 👍

repo/currency_converter.go
placer14: Right. So we should have:
```
CurrencyConverter{
  ReserveWallet: reserveWallet,
}
```

This does two things for us in the future:
1. We know exactly what this is used as within CurrencyConverter without looking at the definition (we aren't always looking at code inside an IDE)
2. We can expand this struct without any additional noise in future changes (extra fields don't force us to change this line again)
3. You can be less verbose with the function argument name

Minor wins which all add up.

placer14:
Resolved with https://github.com/OpenBazaar/openbazaar-go/pull/1990. Closing.

PR #1974, opened 2020-01-31
Support legacy profiles

### LINKED ISSUE #1904: On GET /ob/profile/<peer-id> convert profile to the latest schema

placer14:
Closes #1904


cpacia:
LGTM

PR #1975, opened 2020-02-04
Fix failing signature validation

cpacia:
The primary change here was making sure MarshalProtobuf is used for listing serialization rather than MarshalJSON.

Resolves signing error remaining from https://github.com/OpenBazaar/openbazaar-go/pull/1906


cpacia:
Should this target ashwin's branch instead of ethereum-master?

drwasho:
No as the problems are in `ethereum-master`, which were merged into his branch from `ethereum-master`.

PR #1985, opened 2020-02-17
Fix transaction ordering and enable Tor flag for the qa script

amangale:
This PR fixes the reverse chronological ordering of eth txns. It also adds a tor flag to the qa script.

### DIFF
--- a/qa/test_framework/test_framework.py
+++ b/qa/test_framework/test_framework.py
@@ -112,7 +112,10 @@ class OpenBazaarTestFramework(object):
                     return
 
     def start_node(self, node):
-        args = [self.binary, "start", "-v", "-d", node["data_dir"], *self.options]
+        if self.useTor:
+            args = [self.binary, "start", "tor", "-v", "-d", node["data_dir"], *self.options]
+        else:
+            args = [self.binary, "start", "-v", "-d", node["data_dir"], *self.options]
         process = subprocess.Popen(args, stdout=PIPE)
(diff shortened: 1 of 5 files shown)


qa/test_framework/test_framework.py
```diff
@@ -189,12 +192,18 @@ def main(self, options=["--disablewallet", "--testnet", "--disableexchangerates"
...
         parser.add_argument('-c', '--cointype', help="cointype to test", default="BTC")
+        parser.add_argument('-T', '--Tor', help="use tor", default="no")
         args = parser.parse_args(sys.argv[1:])
         self.binary = args.binary
         self.temp_dir = args.tempdir
         self.bitcoind = args.bitcoind
         self.cointype = args.cointype
+        self.useTor = args.Tor
         self.options = options
+        if args.Tor == "yes":
```
cpacia: Not that big of a deal but this could be a boolean rather than a string from the command line.

cpacia:
The rest looks OK to me.

qa/test_framework/test_framework.py
```diff
@@ -112,7 +112,10 @@ def wait_for_init_success(process):
                     return
 
     def start_node(self, node):
-        args = [self.binary, "start", "-v", "-d", node["data_dir"], *self.options]
+        if self.useTor:
+            args = [self.binary, "start", "tor", "-v", "-d", node["data_dir"], *self.options]
```
placer14: The flag should be `--tor`.

PR #1995, opened 2020-02-21
(#1983) Calculate order total w normalized listings

### LINKED ISSUE #1983: core.CalculateOrderTotal needs normalized listing
Some values that `core.(*OpenBazaarNode).CalculateOrderTotal` are consuming may not be present if the listing is not using schema v5. Need to apply listing normalization to this behavior to ensure correct values are used.

placer14:
When processing the order total for contracts with legacy listings, they did not access the data properly in all cases. These changes ensure all access is done through a normalized `repo.Listing`.

Fixes #1983


cpacia:
Seems OK to me

PR #2013, opened 2020-03-04
Values from DB that get reflected into scientific notation are kept as strings

hoffmabc:
Problem: When pulling extremely large numbers out of SQLite as a string type but the Go scanner will convert the string using reflection into a scientific notation string. 

Proposed Solution: Check if reflected string has an 'e' in it then we identify it as a scientific notation and convert it back to a large number in a string.

### DIFF
--- a/repo/db/sales.go
+++ b/repo/db/sales.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


hoffmabc:
@cpacia im' not sure what the potential pitfalls would be here for this approach. maybe you have some thoughts.

PR #2019, opened 2020-03-05
Fix eth wallet websocket update height

### LINKED ISSUE #1996: Incorrect ETH transaction height coming from the socket
When an ETH transaction is created, a `wallet` socket message is received with the transaction data. The message has `height: 1` which is incorrect.

We determine the confirmations of a transaction by subtracting the confirmations from the block height in the `walletUpdate` socket message. This is causing the confirmations to show as 1 less than the current block height.

I've seen transactions suddenly change to the correct number of confirmations, so some process must be correcting the height eventually.

cpacia:
closes #1996


hoffmabc:
LGTM we need to submit to the eth wallet repo too right @cpacia ?

PR #2021, opened 2020-03-06
(#1949) Fix patching profile with moderator fee

### LINKED ISSUE #1946: Server does not clear out unneeded moderatorInfo fields based on the fee type
Here's my profile:

https://pastebin.com/b5jvEQJW

I send a PATCH with the following payload:

```
{
  "moderator": true,
  "moderatorInfo": {
    "fee": {
      "feeType": "PERCENTAGE",
      "percentage": 10
    },
    "description": "Describe the scribe.",
    "termsAndConditions": "The terms are germs.",
    "languages": [
      "en-US"
    ]
  }
}
```

It errors with:

invalid profile: percentage moderator fee should not include a fixed fee or should use (FIXED_PLUS_PERCENTAGE) feeType

Expectation:
The server will infer which fields moderatorInfo fields in the profile are needed based off of the provided feeType and clear out / reset unneeded ones.

cpacia:
closes #1946

### DIFF
--- a/core/profile.go
+++ b/core/profile.go
@@ -195,6 +195,12 @@ func (n *OpenBazaarNode) PatchProfile(patch map[string]interface{}) error {
 
 	repoProfile.NormalizeDataForAllSchemas()
 
+	if repoProfile.GetProtobuf().ModeratorInfo != nil && repoProfile.GetProtobuf().ModeratorInfo.Fee != nil {
+		if repoProfile.GetProtobuf().ModeratorInfo.Fee.FeeType == pb.Moderator_Fee_PERCENTAGE {
+			repoProfile.GetProtobuf().ModeratorInfo.Fee.FixedFee = nil
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


core/profile.go
```diff
@@ -195,6 +195,12 @@ func (n *OpenBazaarNode) PatchProfile(patch map[string]interface{}) error {
 
 	repoProfile.NormalizeDataForAllSchemas()
 
+	if repoProfile.GetProtobuf().ModeratorInfo != nil && repoProfile.GetProtobuf().ModeratorInfo.Fee != nil {
+		if repoProfile.GetProtobuf().ModeratorInfo.Fee.FeeType == pb.Moderator_Fee_PERCENTAGE {
```
hoffmabc: Why not just make a single if condition rather than a nested conditional?

cpacia: Yeah I guess that could work too. I was just thinking about the nil checks as being separate.

PR #2031, opened 2020-03-11
Changed ReleaseFunds func to check if transaction goes through before saving 

rodkeys:
image

### DIFF
--- a/core/disputes.go
+++ b/core/disputes.go
@@ -1210,6 +1210,12 @@ func (n *OpenBazaarNode) ReleaseFunds(contract *pb.RicardianContract, records []
 
 	peerID := order.BuyerID.PeerID
 
+	// Build, sign, and broadcast transaction
+	txnID, err := wal.Multisign(inputs, outputs, mySigs, moderatorSigs, redeemScriptBytes, *big.NewInt(0), true)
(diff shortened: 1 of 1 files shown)


drwasho:
Rod and I tested this the following way:

1. The buyer makes a purchase to an **offline seller**
2. The buyer then immediately disputes the order
3. The buyer then goes **offline**
4. The moderator decides the outcome of the dispute
5. The seller comes **online** and accepts the decision from the moderator, and then goes **offline**
6. The buyer comes **online**

Now, for one reason or another, DISPUTE_CLOSE messages from the seller aren't being detected, so the buyer sees that they can accept the dispute. What was happening was that they would attempt to do this and it would fail with `unknown script type` as the response.

This response is still there and is expected because the funds have already moved out of the multisig escrow (if anything the error should be more descriptive here). **THE PROBLEM** this PR fixes is that the order state would update _before_ it would try and spend the funds. If it did this, it would leave the order in a bad state where it didn't have all the data it needed and it even if it did receive the DISPUTE_CLOSE message, it couldn't progress to the DECIDED state in the correct way.

As a side note, we're looking into why offline messages are being so poorly retrieved from the push nodes.

PR #2046, opened 2020-04-10
Fix testnet to mainnet

hoffmabc:
Need to update migration to use mainnet instead of testnet.


hoffmabc:
This is ready @cpacia

PR #2049, opened 2020-04-13
Fix purchase crypto test

hoffmabc:
We were assuming all listings were for cryptos with base unit of 8 digits. This uses the actual precision from the listing itself.

### DIFF
--- a/qa/testdata/v5/order_crypto.json
+++ b/qa/testdata/v5/order_crypto.json
@@ -2,7 +2,7 @@
     "moderator": "",
     "items": [{
         "listingHash": "",
-        "bigQuantity": "100000000",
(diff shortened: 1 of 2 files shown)


qa/testdata/v5/order_crypto.json
```diff
@@ -2,7 +2,7 @@
     "moderator": "",
     "items": [{
         "listingHash": "",
-        "bigQuantity": "100000000",
+        "bigQuantity": "1000000000000000",
```
hoffmabc, line 5: This is to get above the dust level

hoffmabc: as a note @cpacia ^

PR #2053, opened 2020-04-17
Check for pre-bigInt notifications pricing

hoffmabc:
Currently, we don't unmarshal pre-big.int notifications properly for the client.


cpacia:
LGTM

PR #2059, opened 2020-04-21
Update to support Tor connections and update command

hoffmabc:


hoffmabc:
ok @drwasho and @cpacia let's test this. I think it's working well but you guys let me know if you have any concerns.

PR #2099, opened 2020-09-15
Require ethereum key at startup to use eth

cpacia:


hoffmabc:
Ok these look fine to me

hoffmabc:
Do we know what happens if a bad infura key is passed in @cpacia?

